PR #59, opened 2019-11-26
Custom JSON Coders

calebkleveter:
This PR lays the groundwork for allowing custom JSON encoders/decoders to be used when encoding or decoding Postgres data.


calebkleveter:
This should probably have tests, but I don't have time for that just now.

calebkleveter:
I completely agree. However, unless that is implemented for beta 2, this
should be merged in until the other fix is implemented so this at least
works.

PR #139, opened 2021-02-12
Remove unintended SASL print statement

fabianfett:
We currently print the internal SASL state when authenticating via `SCRAM-SHA256`. This messes up logs and is a potential security issue.

### Modifications

- Remove a print `statement`


gwynne (approved):
Maybe someday that `print()` will actually stay gone after being removed!

tanner0101:
These changes are now available in [1.4.3](https://github.com/vapor/postgres-nio/releases/tag/1.4.3)

PR #142, opened 2021-02-25
Ensure no interference with other services in tests

fabianfett:
Ensure no interference with other services in tests.

### Motivation

- `testConnectionFailure` tried to create a Connection against `1234`. If a service was running on `1234` we might end up with a flaky test

### Modifications

- `testConnectionFailure` now starts a local server on a random local port, stops it and tries to connect to its previously used port. Therefore this test can not run into any crazy side effects any more.
- `testConnectionFailure` was moved from `IntegrationTests` to `PSQLConnectionTests`

### Result

- fewer potentially flaky tests
- needs no release!


0xTim (approved):
LGTM

PR #135, opened 2021-01-29
State machine

fabianfett:
## Motivation

Connection state changes can be modeled with state machines. This PR introduces a `ConnectionStateMachine` that models the `PostgresConnection`'s state. This will improve the Unit-testability of this package.

## Changes

- This PR **does not** change any public API
- It replaces the underlying implementation of `PostgresConnection` with a new `PSQLConnection`
- All new classes and structs that support the new `PSQLConnection` live in the `/New` folder
- The `PSQLConnection` is mainly implemented using these `structs` and `classes`:
	- `PSQLFrontendMessage` a new enum modeling outgoing client messages to the server
	- `PSQLBackendMessage` a new enum modeling incoming server messages to the client
	- `PSQLFrontendMessage.Encoder` a new `MessageToByteEncoder`
	- `PSQLBackendMessage.Decoder` a new `ByteToMessageEncoder`
	- `ConnectionStateMachine` with the sub state machines:
		- `AuthenticationStateMachine`
		- `CloseStateMachine`
		- `ExtendedQueryStateMachine`
		- `PrepareStatementStateMachine`
	- `PSQLChannelHandler` is the main `ChannelDuplexHandler` for the `PSQLConnection`
	- `PSQLEventsHandler` is a new `ChannelHandler` that observes the channel events and closes the connection when appropriate.
	- `PSQLEncodable` is a new protocol to convert Swift types to Postgres types.
	- `PSQLDecodable` is a new protocol to convert Postgres types to Swift types.
	- `PSQLEncodable` and `PSQLDecodable` defer the encoding and decoding as much as possible, saving unnecessary allocations and conversions.
- Thanks to these changes extended queries and prepared statements will now take the same code paths. Thus we ensure that code improvements for one query type will also lead to improvements for the other.

## Result

We get a better testable `PostgresConnection`.

PR #144, opened 2021-02-26
Fixes an issue where empty values were treated as null values and vica versa

### LINKED ISSUE #143: "Could not convert to String: <null>" error with empty strings
### Describe the bug

After upgrading to postgres-nio 1.5.0, our server started crashing with a bunch of errors like this:

`[ ERROR ] invalid field: stripe_id type: String error: typeMismatch(Swift.String, Swift.DecodingError.Context(codingPath: [], debugDescription: "Could not convert to String: <null>", underlyingError: nil)) [request-id: 4A11F25B-AD9C-43EC-BC9B-996A434DBAF2]`

The indicated field is non-null on Postgres and non optional on Swift. I discovered that this is happening when the field's value is an empty string, so it would seem like empty strings are being treated as null values.

### To Reproduce

1) Create non-null text field in postgres.
2) Create Vapor model and connect with that field (with a non optional var)
3) Add a new row with an empty string.
4) Try to fetch that row in Vapor -> it should crash.

### Expected behavior

It should return the empty string instead of crashing.

Vapor 4.32.0
Fluent 4.0.0
Postgres-nio 1.5.0
Postrges 11.5

fabianfett:
Fix encoding and decoding optional values to and from the database due to a regression in #135. Resolves #143 (#144)

### DIFF
--- a/Sources/PostgresNIO/New/Data/Array+PSQLCodable.swift
+++ b/Sources/PostgresNIO/New/Data/Array+PSQLCodable.swift
@@ -91,7 +91,7 @@ extension Array: PSQLEncodable where Element: PSQLArrayElement {
         buffer.writeInteger(1, as: Int32.self)
 
         try self.forEach { element in
-            try element._encode(into: &buffer, context: context)
+            try element.encodeRaw(into: &buffer, context: context)
         }
     }
 }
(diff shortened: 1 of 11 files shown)


0xTim (approved):
LGTM

tanner0101:
These changes are now available in [1.5.1](https://github.com/vapor/postgres-nio/releases/tag/1.5.1)

tanner0101:
These changes are now available in [1.5.1](https://github.com/vapor/postgres-nio/releases/tag/1.5.1)

PR #149, opened 2021-03-23
Add proper support for timestamp arrays

0xTim:


fabianfett:
Do we want to add this support also for the potentially new `PSQLCodable` type? You would add this by adding

```swift
extension Date: PSQLArrayElement {
    static var psqlArrayType: PSQLDataType { .timestamptzArray }
    static var psqlArrayElementType: PSQLDataType { .timestamptz }
}
```

to `Array+PSQLCodable`. This should be it!

0xTim:
Yeah if that's simpler and better long term we should migrate to that

gwynne:
Superseded (at last) by #463

PR #151, opened 2021-04-21
Ignore errors when closing the connection

### LINKED ISSUE #150: Connections to Azure Postgres dbs raise `NIOSSLError.uncleanShutdown` on shutdown
### Description

This is a follow-up to a [discussion on Discord](https://discord.com/channels/431917998102675485/448584561845338139/833567195849424956) where I first asked about the error and was advised to raise an issue.

We've recently switched the Swift Package Index' db from a non-SSL docker Postgres instance to Azure hosted Postgres, which uses SSL. Seemingly every command (Vapor.Command.run) that we're running in batch jobs is reporting

```
[ ERROR ] Channel error caught. Closing connection. [psql_connection_id: E70F3BE3-BBE1-4CD7-AE49-83DFF7877C13, psql_error: PSQLError(base: PostgresNIO.PSQLError.Base.connectionError(underlying: NIOSSL.NIOSSLError.uncleanShutdown))]
```

on shutdown.

This doesn't seem to affect operation except that it's really hard to diagnose issues with all the error noise.

It's not happening in the request loops of the webserver, so it seems to be an issue with Vapor.Command.run exiting uncleanly.

It only seems to be happening with Azure dbs. Running against a local dockerized PG db with SSL (same version, 11.6) does not exhibit the problem.

I've raised an issue with Azure to see if it's a config/server issue.

### To Reproduce

One example command that triggers the error is [here](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/main/Sources/App/Commands/ReAnalyzeVersions.swift) but there's actually a stand-alone recipe to observe the error that I'm attaching as a playground.

It can also be re-created as follows:

Run [Arena](https://github.com/finestructure/Arena) to create a new playground with the required dependencies:

```
arena https://github.com/vapor/vapor https://github.com/vapor/fluent https://github.com/vapor/fluent-postgres-driver -o ssl-error-repro
```

and add:

```swift
import Vapor
import Fluent
import FluentPostgresDriver

final class Test: Model, Content {
    static let schema = "test"

    @ID(key: .id)
    var id: UUID?
}

func run() throws {
    let app = Application(.testing)
    defer { app.shutdown() }

    let tlsConfig: TLSConfiguration = .forClient()
    app.databases.use(.postgres(hostname: "pgnio-debug.postgres.database.azure.com",
                                port: 5432,
                                username: "test@pgnio-debug",
                                password: "<ask me for the password>",
                                database: "test",
                                tlsConfiguration: tlsConfig), as: .psql)

    let db = app.db

    let t = try Test.find(UUID("cafecafe-cafe-cafe-cafe-cafecafecafe"), on: db).wait()
    print("t: \(String(describing:  t))")
}

run()
```

When running, the console should print

```
t: Optional(Test(output: ["test_id": CAFECAFE-CAFE-CAFE-CAFE-CAFECAFECAFE]))
2021-04-19T12:05:25+0200 error codes.vapor.postgres : psql_connection_id=A8F4884F-DCB4-47CD-AB80-A48A3D320F00 psql_error=uncleanShutdown Channel error caught.
2021-04-19T12:05:25+0200 error codes.vapor.postgres : psql_connection_id=A8F4884F-DCB4-47CD-AB80-A48A3D320F00 psql_error=PSQLError(base: PostgresNIO.PSQLError.Base.connectionError(underlying: NIOSSL.NIOSSLError.uncleanShutdown)) Channel error caught. Closing connection.
```

And the expectation is that it _shouldn't_ 😆

fabianfett:
Resolves an issue where errors thrown by some database servers during closing a connection would cause an error in PostgresNIO. Resolves #150 

### Modifications

- We ignore errors when the connection is already closing
- Updated logging to only issue debug and trace logs.
- Don't print password in trace logging mode.

### DIFF
--- a/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
+++ b/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
@@ -313,8 +313,7 @@ struct ConnectionStateMachine {
              .waitingToStartAuthentication,
              .authenticated,
              .readyForQuery,
-             .error,
-             .closing:
+             .error:
             return self.closeConnectionAndCleanup(.server(errorMessage))
         case .authenticating(var authState):
             if authState.isComplete {
@@ -352,6 +351,13 @@ struct ConnectionStateMachine {
                 machine.state = .prepareStatement(preparedState, connectionContext)
                 return machine.modify(with: action)
             }
+        case .closing:
+            // If the state machine is in state `.closing`, the connection shutdown was initiated
+            // by the client. This means a `TERMINATE` message has already been sent and the
+            // connection close was passed on to the channel. Therefore we await a channelInactive
+            // as the next event.
+            // Since a connection close was already issued, we should keep cool and just wait.
+            return .wait
         case .initialized, .closed:
             preconditionFailure("We should not receive server errors if we are not connected")
         case .modifying:
@@ -367,8 +373,7 @@ struct ConnectionStateMachine {
              .sslHandlerAdded,
              .waitingToStartAuthentication,
              .authenticated,
-             .readyForQuery,
-             .closing:
+             .readyForQuery:
             return self.closeConnectionAndCleanup(error)
         case .authenticating(var authState):
             let action = authState.errorHappened(error)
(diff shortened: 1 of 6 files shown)


0xTim (approved):
LGTM

finestructure:
I've just run my test case (without `trace` logging) and it logs simply

```
elapsed: 1.7917180061340332
```

without any errors. Looks like it's all fixed, thanks a lot! :)

PR #152, opened 2021-04-29
Use sync pipeline operations to add NIOSSLClientHandler

fabianfett:
- Bump required NIO dependencies.
- Modify the `PSQLChannelHandler`'s `enableSSLCallback` to a synchronous API
- Use `channel.pipeline.syncOperations` to add the `NIOSSLClientHandler` after it was verified that SSL is supported on this connection
- Extend test `testEstablishSSLCallbackIsCalledIfSSLIsSupported` to ensure startup is send after SSL connection is established.

### DIFF
--- a/Sources/PostgresNIO/New/PSQLConnection.swift
+++ b/Sources/PostgresNIO/New/PSQLConnection.swift
@@ -215,17 +215,16 @@ final class PSQLConnection {
                     .channelInitializer { channel in
                         let decoder = ByteToMessageHandler(PSQLBackendMessage.Decoder())
                         
-                        var enableSSLCallback: ((Channel) -> EventLoopFuture<Void>)? = nil
+                        var configureSSLCallback: ((Channel) throws -> ())? = nil
                         if let tlsConfiguration = configuration.tlsConfiguration {
-                            enableSSLCallback = { channel in
-                                channel.eventLoop.submit {
-                                    let sslContext = try NIOSSLContext(configuration: tlsConfiguration)
-                                    return try NIOSSLClientHandler(
-                                        context: sslContext,
-                                        serverHostname: configuration.sslServerHostname)
-                                }.flatMap { sslHandler in
-                                    channel.pipeline.addHandler(sslHandler, position: .before(decoder))
-                                }
+                            configureSSLCallback = { channel in
+                                channel.eventLoop.assertInEventLoop()
+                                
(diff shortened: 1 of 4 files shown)


0xTim (approved):
LGTM. Will let @gwynne have the final say. Changed to patch update since there are no public API additions

fabianfett:
@0xTim Agree there are no api changes in Postgres, but we bump or dependencies, that we in part reexport (like NIO) in a minor way. Therefor this needs IMO a minor release.

0xTim:
Oh FFS why is PostgresNIO exporting everything! * sigh * you're right I'll change it back. I can't wait to get rid of all these exports!

fabianfett:
We'll it's not so much export everything. But in this case EventLoopFuture is enough. 😎

Sources/PostgresNIO/New/PSQLConnection.swift
```diff
@@ -215,17 +215,16 @@ final class PSQLConnection {
                     .channelInitializer { channel in
                         let decoder = ByteToMessageHandler(PSQLBackendMessage.Decoder())
                         
-                        var enableSSLCallback: ((Channel) -> EventLoopFuture<Void>)? = nil
+                        var enableSSLCallback: ((Channel) throws -> ())? = nil
```
glbrntt, line 218: Super nit but this might be clearer as `configureSSLCallback`

Sources/PostgresNIO/New/PSQLConnection.swift
```diff
@@ -215,17 +215,16 @@ final class PSQLConnection {
...
                             enableSSLCallback = { channel in
-                                channel.eventLoop.submit {
-                                    let sslContext = try NIOSSLContext(configuration: tlsConfiguration)
-                                    return try NIOSSLClientHandler(
-                                        context: sslContext,
-                                        serverHostname: configuration.sslServerHostname)
-                                }.flatMap { sslHandler in
-                                    channel.pipeline.addHandler(sslHandler, position: .before(decoder))
-                                }
+                                channel.eventLoop.preconditionInEventLoop()
```
glbrntt, line 221: FWIW the sync operations will do this check for you; keeping this (and potentially degrading it to an assertion) is a useful as an indicator to whoever looks at this code though.

glbrntt (approved):
LGTM!

fabianfett:
Renamed `enableSSLCallback` to `configureSSLCallback`.

PR #153, opened 2021-05-01
Boost performance by batching Rows

fabianfett:
In #135 a connection state machine was introduced. To keep this simple at first, the state machine operates on single `PostgresBackendMessage`s. This is very inefficient for queries that result in a lot of rows. This PR improves query performance by badging Postgres data row messages at the `Decoder` and operating on those batches from there on.

Open ends:
- Two open warnings


Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -91,12 +91,9 @@ struct ConnectionStateMachine {
...
         // actions if query has requested next row but we are waiting for backend
-        case forwardRow([PSQLData], to: EventLoopPromise<StateMachineStreamNextResult>)
-        case forwardCommandComplete(CircularBuffer<[PSQLData]>, commandTag: String, to: EventLoopPromise<StateMachineStreamNextResult>)
-        case forwardStreamError(PSQLError, to: EventLoopPromise<StateMachineStreamNextResult>, cleanupContext: CleanUpContext?)
-        // actions if query has not asked for next row but are pushing the final bytes to it
-        case forwardStreamErrorToCurrentQuery(PSQLError, read: Bool, cleanupContext: CleanUpContext?)
-        case forwardStreamCompletedToCurrentQuery(CircularBuffer<[PSQLData]>, commandTag: String, read: Bool)
+        case forwardRows(CircularBuffer<PSQLBackendMessage.DataRow>)
+        case forwardStreamComplete(CircularBuffer<PSQLBackendMessage.DataRow>, commandTag: String, read: Bool)
+        case forwardStreamError(PSQLError, read: Bool, cleanupContext: CleanUpContext?)
```
fabianfett, line 96: Reduction in `Actions` 🥳

Sources/PostgresNIO/New/PSQLChannelHandler.swift
```diff
@@ -18,6 +18,9 @@ final class PSQLChannelHandler: ChannelDuplexHandler {
             self.logger.trace("Connection state changed", metadata: [.connectionState: "\(self.state)"])
         }
     }
+    
+    #warning("This is an ugly hack, is there some way around it?")
+    private var handlerContext: ChannelHandlerContext?
```
fabianfett, line 23: @weissi How bad is it to store a `ChannelHandlerContext` for use later? With this property we can use it once we get an event that is not triggered by the ChannelPipeline.

Sources/PostgresNIO/New/PSQLChannelHandler.swift
```diff
@@ -491,6 +463,19 @@ final class PSQLChannelHandler: ChannelDuplexHandler {
     }
 }
 
+extension PSQLChannelHandler: PSQLRowsDataSource {
+    
+    func request() {
+        let action = self.state.requestQueryRows()
+        self.run(action, with: self.handlerContext!)
```
fabianfett, line 470: @weissi the stored [`ChannelHandlerContext`](https://github.com/vapor/postgres-nio/pull/153/files#r624473616) is used here. If we don't want to store the `ChannelHandlerContext` here, we could trigger the requestQueryRows by sending a `ChannelOutboundEvent`. What is the preferred way?

Sources/PostgresNIO/New/PSQLRows.swift
```diff
@@ -53,48 +70,93 @@ final class PSQLRows {
...
+                self.downstreamState = .consuming
+                self.upstreamState = .streaming(buffer: buffer, dataSource: dataSource)
                 let row = Row(data: data, lookupTable: self.lookupTable, columns: self.rowDescription, jsonDecoder: self.jsonDecoder)
                 return self.eventLoop.makeSucceededFuture(row)
             }
             
+            let promise = self.eventLoop.makePromise(of: Row?.self)
+            self.downstreamState = .waitingForNext(promise)
+            self.upstreamState = .streaming(buffer: buffer, dataSource: dataSource)
+            dataSource.request()
```
fabianfett, line 91: This is the invocation, where the PSQLRowStream asks for more rows. We could change this to:

```swift
context.channel.triggerUserOutboundEvent(PSQLRowStreamEvent.requestMoreRows, promise: nil)
```

Sources/PostgresNIO/New/PSQLChannelHandler.swift
weissi: It’s fine to store a context on a handler. The usual warnings that they’re not thread-safe and everybody who has your context has the same access to the pipeline as the handler has. But here you’re storing it on the handler which is totally fine

Sources/PostgresNIO/New/PSQLChannelHandler.swift
weissi: It’s fine. See below

fabianfett:
### Adding some performance numbers...

Running a query `SELECT * FROM "large_table"` that returns 1million rows with three number each:

main branch:
```
root@3bcc272414db:/vapor/postgres-perf# time .build/release/PostgresPerf
2021-05-02T13:12:26+0000 info psql : PostgresQueryMetadata(command: "SELECT", oid: nil, rows: Optional(1000000))
2021-05-02T13:12:26+0000 info psql : query took: 5.985878944396973 seconds

real	0m6.256s
user	0m5.955s
sys	0m0.135s
```

this branch:
```
root@3bcc272414db:/vapor/postgres-perf# time .build/release/PostgresPerf
2021-05-02T13:13:23+0000 info psql : PostgresQueryMetadata(command: "SELECT", oid: nil, rows: Optional(1000000))
2021-05-02T13:13:23+0000 info psql : query took: 3.5723689794540405 seconds

real	0m4.003s
user	0m1.110s
sys	0m0.435s
```

While the speed bump is nice. Even more important: The query doesn't use 100% cpu anymore.

jdmcd:
This is really enticing, any chance we can port this easily to MySQL as well? 👀

fabianfett:
@jdmcd Since this pr builds on top of #135, I'm afraid some groundwork would need to be done to have this land.

jdmcd:
Ah, I see

gwynne:
> This is really enticing, any chance we can port this easily to MySQL as well? 👀

@jdmcd FWIW this is something I've already got on my list to look into for MySQL 🙂

fabianfett:
Just applied another fix. Will push shortly. On Swift 5.5 nightly:

```
Read 1m rows into an array
main        : 5.354 sec
this branch : 1.790 sec

Consume 1m rows on el
main        : 5.176 sec
this branch : 1.475 sec
```

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -289,20 +283,20 @@ struct ExtendedQueryStateMachine {
...
         switch self.state {
         case .waitingForNextRow:
             preconditionFailure("Too greedy. `consumeNextRow()` only needs to be called once.")
             
-        case .bufferingRows(let columns, var buffer, let readOnEmpty):
+        case .bufferingRows(let columns, let buffer, let readOnEmpty):
             return self.avoidingStateMachineCoW { state -> Action in
-                guard let row = buffer.popFirst() else {
-                    state = .waitingForNextRow(columns, buffer, promise)
+                guard buffer.count > 0 else {
```
gwynne, line 293:
```suggestion
                guard !buffer.isEmpty else {
```

PR #154, opened 2021-05-03
Add platform requirements for iOS, watchOS and tvOS

fabianfett:
To use `postgres-nio` from iOS, watchOS and tvOS the platform requirements need to be set correctly.

Requirement raised in the forums: https://forums.swift.org/t/trouble-importing-postgresnio/47815/10


0xTim (approved):

0xTim:
We'll batch this up into a release with some of the other PRs

PR #155, opened 2021-05-04
[RFC] Create connection and authenticate with a single function

fabianfett:
### Current State

The current API forces users to do some state handling themselves during connection setup. Also `throwing` functions are mixed with functions that return `EventLoopFuture`s. From the README:

```swift
import PostgresNIO

let eventLoop: EventLoop = ...
let conn = try PostgresConnection.connect(
    to: .makeAddressResolvingHost("my.psql.server", port: 5432),
    on: eventLoop
).wait()
defer { try! conn.close().wait() }
```

There is a couple of issues with this:

1. `.makeAddressResolvingHost("my.psql.server", port: 5432)` is a throwing function
2. `PostgresConnection.connect` returns an `EventLoopFuture`. 
3. The connection is only setup and the user needs to authenticate from here on. In an error case the user needs to teardown the connection themselves.

If users want to create a connection themselves they are forced to write this code, which has a lot of implicit connection state:
```swift
eventLoop.submit { () throws -> SocketAddress in
    // callback only to transform the throwing SocketAddress into a Future for consistency
    try .makeAddressResolvingHost("my.psql.server", port: 5432)
}.flatMap {
    // connect to the server. Can setup TLS
    PostgresConnection.connect(to: $0, on: eventLoop)
}.flatMap { connection in
    // authenticate to database
    connection.authenticate(
        username: env("POSTGRES_USER") ?? "vapor_username",
        database: env("POSTGRES_DB") ?? "vapor_database",
        password: env("POSTGRES_PASSWORD") ?? "vapor_password"
    ).map {
        // map to connection to return the connection to the user
        return conn
    }.flatMapError { error in
        // close connection in case of auth error
        conn.close().flatMapThrowing {
            // rethrow error for the user to capture
            throw error
        }
    }
}
```

### Changes

With the StateMachine PR #135 a new underlying PSQLConnection has landed that supports connection creation and authentication with a single call. This PR brings this much easier API to the public interface:

#### Additions

```swift
class PostgresConnection {
    /// Create a new Postgres Connection to a Postgres Server.
    ///
    /// - Parameters:
    ///   - hostname: The server's hostname as domain or IP address.
    ///   - port: The server's port. Defaults to the Postgres default port 5432.
    ///   - username: The username to authenticate with to the remote server.
    ///   - password: An optional password to authenticate with to the remote server.
    ///   - database: The database name to connect to
    ///   - tlsConfiguration: An optional `TLSConfiguration` that configures the TLS connection to the server. If
    ///                       none is provided the connection to the server is created without TLS.
    ///   - logger: A logger to use for connection state changes and action. The logger is only with the `debug` and
    ///             `trace` log level.
    ///   - eventLoop: The SwiftNIO `EventLoop` the connection should be created on.
    /// - Returns: An `EventLoopFuture` which provides the `PSQLConnection` if the connection creation and
    ///            authentification was successful. Otherwise an `Error`.
    public static func connect(
        hostname: String,
        port: Int = 5432,
        username: String,
        password: String?,
        database: String?,
        tlsConfiguration: TLSConfiguration?,
        logger: Logger = Logger(label: "codes.vapor.postgres", factory: { _ in SwiftLogNoOpLogHandler() }),
        on eventLoop: EventLoop
    ) -> EventLoopFuture<PostgresConnection>
}
```

For users that want to connect to their Postgres server via unix sockets, we offer a new advanced API:
```swift
    /// Create a new Postgres Connection to a Postgres Server.
    ///
    /// - Parameters:
    ///   - address: The remote server's address as a `SocketAddress`.
    ///   - serverName: The remote server's name to use for domain name certificate verification, if a
    ///                 `TLSConfiguration` is supplied.
    ///   - username: The username to authenticate with to the remote server.
    ///   - password: An optional password to authenticate with to the remote server.
    ///   - database: The database name to connect to
    ///   - tlsConfiguration: An optional `TLSConfiguration` that configures the TLS connection to the server. If
    ///                       none is provided the connection to the server is created without TLS.
    ///   - logger: A logger to use for connection state changes and action. The logger is only with the `debug` and
    ///             `trace` log level.
    ///   - eventLoop: The SwiftNIO `EventLoop` the connection should be created on.
    /// - Returns: An `EventLoopFuture` which provides the `PSQLConnection` if the connection creation and
    ///            authentification was successful. Otherwise an `Error`.
    public static func connect(
        address: SocketAddress,
        serverName: String? = nil,
        username: String,
        password: String?,
        database: String?,
        tlsConfiguration: TLSConfiguration?,
        logger: Logger = Logger(label: "codes.vapor.postgres", factory: { _ in SwiftLogNoOpLogHandler() }),
        on eventLoop: EventLoop
    ) -> EventLoopFuture<PostgresConnection>
```

#### Deprecations

The current `connect` and `authenticate` methods are deprecated.

```swift

    @available(*, deprecated, message: "Use the new `PostgresConnection.connect(hostname:port:username:password:database:tlsConfiguration:logger:on eventLoop:)` that allows you to specify username, password and database directly when creating the connection.")
    public static func connect(
        to socketAddress: SocketAddress,
        tlsConfiguration: TLSConfiguration? = nil,
        serverHostname: String? = nil,
        logger: Logger = .init(label: "codes.vapor.postgres"),
        on eventLoop: EventLoop
    ) -> EventLoopFuture<PostgresConnection>

    @available(*, deprecated, message: "Use the new `PostgresConnection.connect(hostname:port:username:password:database:tlsConfiguration:logger:on eventLoop:)` that allows you to specify username, password and database directly when creating the connection.")
    public func authenticate(
        username: String,
        database: String? = nil,
        password: String? = nil,
        logger: Logger = .init(label: "codes.vapor.postgres")
    ) -> EventLoopFuture<Void>
}
```

### Alternatives

Instead of using eight parameters for the connect function, we could use the a Config struct like this from PSQLConnection: https://github.com/vapor/postgres-nio/blob/main/Sources/PostgresNIO/New/PSQLConnection.swift#L12-L77

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection+Authenticate.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection+Authenticate.swift
@@ -1,6 +1,7 @@
 import NIO
 
 extension PostgresConnection {
+    @available(*, deprecated, message: "Use the new `PostgresConnection.connect()` that allows you to specify username, password and database directly when creating the connection.")
     public func authenticate(
         username: String,
         database: String? = nil,
--- a/Sources/PostgresNIO/Connection/PostgresConnection+Connect.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection+Connect.swift
@@ -1,6 +1,8 @@
 import NIO
+import Logging
 
 extension PostgresConnection {
+    @available(*, deprecated, message: "Use the new `PostgresConnection.connect(hostname:port:username:password:database:tlsConfiguration:logger:on eventLoop:)` that allows you to specify username, password and database directly when creating the connection.")
     public static func connect(
         to socketAddress: SocketAddress,
         tlsConfiguration: TLSConfiguration? = nil,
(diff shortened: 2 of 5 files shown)


weissi:
Nice, this certainly simplifies things a lot.

The only concern I have with this is that it gives username/password a privileged place in the API but PostgreSQL supports [many auth mechanisms](https://www.postgresql.org/docs/10/auth-methods.html).

I think before committing to new APIs, we should think through how we could in the future evolve the APIs to support the other mechanisms. What doesn't work super well IMHO is to have a large number of `connect` methods, each only differing in the auth mechanism. (just for the record, I'm not saying that we should _implement_ all this now but we should have a clear API evolution plan on how we could add things in the future).

One particular auth mechanism to consider is [cert auth](https://www.postgresql.org/docs/10/auth-methods.html#AUTH-CERT) because it has implications on the TLS configuration too. So we can't fully separate TLS & auth for all the mechanisms. I'm sure @lukasa has opinions here, we've learnt how not to do a few things from AHC :).

Lukasa:
Yeah the mere presence of cert auth interacts weirdly with all of this.

I'm with @weissi: the plan should be to find a way to encode auth fairly generally, even if we don't actually _do_ it right away.

fabianfett:
Thanks, for your feedback @weissi.

We will always need a username and potentially a database. The auth mechanisms 

Let's look at all Authentication methods one by one. From Postgres docs [20.3 Authentication Methods](https://www.postgresql.org/docs/13/auth-methods.html)

> [Trust authentication](https://www.postgresql.org/docs/13/auth-trust.html), which simply trusts that users are who they say they are.

Even though we don't need to authenticate we need to supply a **username** during startup.

> [Password authentication](https://www.postgresql.org/docs/13/auth-password.html), which requires that users send a password.

We need a **username** and a **password**. The password may be send in one of three modes:

- scram-sha-256
- md5
- password

All are supported by the current implementation \o/. Thanks @gwynne for providing `scram-sha-256`.

> [GSSAPI authentication](https://www.postgresql.org/docs/13/gssapi-auth.html), which relies on a GSSAPI-compatible security library. Typically this is used to access an authentication server such as a Kerberos or Microsoft Active Directory server.

Currently not supported. Quick look... I have no idea what is going on there.

> [SSPI authentication](https://www.postgresql.org/docs/13/sspi-auth.html), which uses a Windows-specific protocol similar to GSSAPI.

Currently not supported. Quick look... I have no idea what is going on there.

> [Ident authentication](https://www.postgresql.org/docs/13/auth-ident.html), which relies on an “Identification Protocol” (RFC 1413) service on the client's machine. (On local Unix-socket connections, this is treated as peer authentication.)

Currently not supported. IIUC this doesn't need any user input.

> [Peer authentication](https://www.postgresql.org/docs/13/auth-peer.html), which relies on operating system facilities to identify the process at the other end of a local connection. This is not supported for remote connections.

This should work. No user input needed. Looks to the client like a trusted connection.

> [LDAP authentication](https://www.postgresql.org/docs/13/auth-methods.html), which relies on an LDAP authentication server.

Similar to the password dance. The actual authentication is just delegated to a LDAP server.

> [RADIUS authentication](https://www.postgresql.org/docs/13/auth-radius.html), which relies on a RADIUS authentication server.

Similar to the password dance. The actual authentication is just delegated to a RADIUS server.

> [Certificate authentication](https://www.postgresql.org/docs/13/auth-cert.html), which requires an SSL connection and authenticates users by checking the SSL certificate they send.

I assume this is from a protocol standpoint the same as trust authentication. Certificate cn name must match the username.

> [PAM authentication](https://www.postgresql.org/docs/13/auth-pam.html), which relies on a PAM (Pluggable Authentication Modules) library.

Similar to the password dance. The actual authentication is just delegated to a PAM service.

> [BSD authentication](https://www.postgresql.org/docs/13/auth-bsd.html), which relies on the BSD Authentication framework (currently available only on OpenBSD).

Similar to the password dance. The actual authentication is just delegated to a BSD Authentication service.

----

The [startup message](https://www.postgresql.org/docs/13/protocol-message-formats.html) requires a username and an optional database name (the fallback is the username).

On the one hand from a first glance nearly all auth methods need either nothing or only a password to work. Exception from what I can see right now is `GSSAPI`/`SSPI`.

On the other hand we have quite a bulk of additional Parameter Key Words in the libpq docu: https://www.postgresql.org/docs/13/libpq-connect.html#LIBPQ-PARAMKEYWORDS

fabianfett:
To allow for different authentication modes down the road and given that we don't know what authentication scheme is, that the server requires we will need to create some kind of AuthenticationHandler with callbacks for different auth methods.

```swift
struct AuthenticationHandler: ExpressibleByStringLiteral {
  var providePassword: ((EventLoop) -> EventLoopFuture<String>)?

  static func password(_ pw: String) -> Self {
    init(providePassword: {
      $0.makeSucceededFuture(pw)
    })
  }

  init(stringLiteral value: String) {
    self.providePassword: {
      $0.makeSucceededFuture(value)
    })
  }
}

class PostgresConnection {
    public static func connect(
        hostname: String,
        port: Int = 5432,
        username: String,
        database: String?,
        authentication: AuthenticationHandler?,
        tlsConfiguration: TLSConfiguration?,
        logger: Logger = Logger(label: "codes.vapor.postgres", factory: { _ in SwiftLogNoOpLogHandler() }),
        on eventLoop: EventLoop
    ) -> EventLoopFuture<PostgresConnection>
}
```

PR #157, opened 2021-05-10
Split out `IntegrationTests` into separate `testTarget`

fabianfett:
This PR splits out all the integration tests into a separate test target. This makes a number of things easier in day to day development: 

1. Unit tests can be run without setting up a test database.
2. We can observe the test coverage of our Unit tests only.
3. This allows us to decouple testing the language vs. testing the functionality against a database.

I think after this change we could change CI to:

Unit testing for:
```
      matrix:
        swiftver:
          - swift:5.2
          - swift:5.3
          - swift:5.4
          - swiftlang:swift/5.5-nightly
          - swiftlang:swift/main-nightly
        swiftos:
          - focal
```

Integration Testing tests for:
```
      matrix:
        dbimage:
          - postgres:13
          - postgres:12
          - postgres:11
        dbauth:
          - trust
          - md5
          - scram-sha-256
        swiftver:
          - swift:5.4
        swiftos:
          - focal
```


Tests/IntegrationTests/Utilities.swift
```diff
@@ -0,0 +1,75 @@
...
+
+extension XCTestCase {
+    
+    public static var shouldRunLongRunningTests: Bool {
+        // The env var must be set and have the value `"true"`, `"1"`, or `"yes"` (case-insensitive).
+        // For the sake of sheer annoying pedantry, values like `"2"` are treated as false.
+        guard let rawValue = ProcessInfo.processInfo.environment["POSTGRES_LONG_RUNNING_TESTS"] else { return false }
+        if let boolValue = Bool(rawValue) { return boolValue }
+        if let intValue = Int(rawValue) { return intValue == 1 }
+        return rawValue.lowercased() == "yes"
```
glbrntt, line 62: nit: might as well pull this out into a separate function given it's used below as well

Tests/IntegrationTests/Utilities.swift
```diff
@@ -0,0 +1,75 @@
...
+import Logging
+#if canImport(Darwin)
+import Darwin.C
+#else
+import Glibc
+#endif
+
+extension PostgresConnection {
+    static func address() throws -> SocketAddress {
+        try .makeAddressResolvingHost( env("POSTGRES_HOSTNAME") ?? "localhost", port: 5432)
```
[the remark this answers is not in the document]
fabianfett: Fixed.

Tests/IntegrationTests/Utilities.swift
fabianfett: This is previous code just copy and pasted here. If I try to refactor this we end up with a `warning`. Would love to keep as is for now. We should refactor the perf tests to allocation tests anyway.

glbrntt: Fine by me to leave this for now. Out of curiosity, why do we end up with a warning?

glbrntt (approved):

PR #167, opened 2021-07-24
Support AWS redshift queries

### LINKED ISSUE #166: Support for AWS Redshift database
**Is your feature request related to a problem? Please describe.**
When attempting to query an AWS Redshift database, I can connect successfully but any query results in the error:

In driver version 1.4.2:
`Uncaught error: NIOPostgres error: protocol error: Unexpected query metadata: SELECT (PostgresNIO/Connection/PostgresConnection+Connect.swift:51)`

In driver version 1.5.2:
`PostgresNIO/PostgresConnection+Database.swift:32: Fatal error: Unexpectedly found nil while unwrapping an Optional value` where it is trying to initialize a `PostgresQueryMetadata` using commandTag `SELECT` which it doesn't parse because it seems to be expecting a number here like `SELECT 3` which I think is the row count.

**Describe the solution you'd like**
It would be great if this driver would work with Redshift which is a postgres variant.

**Describe alternatives you've considered**
I have not been able to find any information on this or any other way to connect to Redshift.

grennis:
This changes the metadata parser to accept a response that is missing the rowcount, event though this does not follow the spec. In my testing this allows my to now successfully query Redshift databases.

Fixes #166

### DIFF
--- a/Sources/PostgresNIO/PostgresDatabase+Query.swift
+++ b/Sources/PostgresNIO/PostgresDatabase+Query.swift
@@ -74,7 +74,7 @@ public struct PostgresQueryMetadata {
             self.command = .init(parts[0])
             self.oid = Int(parts[1])
             self.rows = Int(parts[2])
-        case "DELETE", "UPDATE", "SELECT", "MOVE", "FETCH", "COPY":
+        case "SELECT":
             // <cmd> rows
             // Note, Redshift does not return the actual count as per the spec
             guard parts.count == 1 || parts.count == 2 else {
@@ -83,6 +83,14 @@ public struct PostgresQueryMetadata {
             self.command = .init(parts[0])
             self.oid = nil
             self.rows = parts.count == 2 ? Int(parts[1]) : 0
+        case "DELETE", "UPDATE", "MOVE", "FETCH", "COPY":
+            // <cmd> rows
+            guard parts.count == 2 else {
+                return nil
+            }
+            self.command = .init(parts[0])
+            self.oid = nil
+            self.rows = Int(parts[1])
         default:
             // <cmd>
             self.command = string
--- a/Sources/PostgresNIO/PostgresDatabase+Query.swift
+++ b/Sources/PostgresNIO/PostgresDatabase+Query.swift
@@ -76,12 +76,13 @@ public struct PostgresQueryMetadata {
             self.rows = Int(parts[2])
         case "DELETE", "UPDATE", "SELECT", "MOVE", "FETCH", "COPY":
             // <cmd> rows
-            guard parts.count == 2 else {
+            // Note, Redshift does not return the actual count as per the spec
+            guard parts.count == 1 || parts.count == 2 else {
                 return nil
             }
             self.command = .init(parts[0])
             self.oid = nil
-            self.rows = Int(parts[1])
+            self.rows = parts.count == 2 ? Int(parts[1]) : 0
         default:
             // <cmd>
             self.command = string
--- a/Sources/PostgresNIO/PostgresDatabase+Query.swift
+++ b/Sources/PostgresNIO/PostgresDatabase+Query.swift
@@ -75,8 +75,8 @@ public struct PostgresQueryMetadata {
             self.oid = Int(parts[1])
             self.rows = Int(parts[2])
         case "SELECT":
-            // <cmd> rows
-            // Note, Redshift does not return the actual count as per the spec
+            // AWS Redshift does not return the actual row count as defined in the postgres wire spec:
+            // https://www.postgresql.org/docs/13/protocol-message-formats.html in section `CommandComplete`
             guard parts.count == 1 || parts.count == 2 else {
                 return nil
             }
--- a/Sources/PostgresNIO/PostgresDatabase+Query.swift
+++ b/Sources/PostgresNIO/PostgresDatabase+Query.swift
@@ -74,16 +74,13 @@ public struct PostgresQueryMetadata {
             self.command = .init(parts[0])
             self.oid = Int(parts[1])
             self.rows = Int(parts[2])
-        case "SELECT":
-            // AWS Redshift does not return the actual row count as defined in the postgres wire spec:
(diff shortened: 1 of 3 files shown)


Sources/PostgresNIO/PostgresDatabase+Query.swift
```diff
@@ -76,12 +76,13 @@ public struct PostgresQueryMetadata {
...
             // <cmd> rows
-            guard parts.count == 2 else {
+            // Note, Redshift does not return the actual count as per the spec
+            guard parts.count == 1 || parts.count == 2 else {
                 return nil
             }
             self.command = .init(parts[0])
             self.oid = nil
-            self.rows = Int(parts[1])
+            self.rows = parts.count == 2 ? Int(parts[1]) : 0
```
fabianfett, line 85: Would you mind refactoring out, the `“SELECT”` into its own case? Right now we would also support `UPDATE` and friends without row number as well.
```swift
case “SELECT” where parts.count == 1:
    // AWS Redshift does not return the actual row count as defined in the postgres wire spec:
    // https://www.postgresql.org/docs/13/protocol-message-formats.html in section `CommandComplete`
    self.command = .init(“SELECT”)
    self.oid = nil
    self.rows = nil
case “DELETE”, “UPDATE”, “SELECT”, “MOVE”, “FETCH”, “COPY”:
    // all the previous code
```

Tests/PostgresNIOTests/New/PSQLDataTests.swift
```diff
@@ -15,4 +15,10 @@ class PSQLDataTests: XCTestCase {
         XCTAssertNoThrow(emptyResult = try data.decode(as: String?.self, context: .forTests()))
         XCTAssertNil(emptyResult)
     }
+    
+    func testMetadataParsing() {
+        XCTAssertEqual(100, PostgresQueryMetadata(string: "SELECT 100")?.rows)
+        XCTAssertEqual(0, PostgresQueryMetadata(string: "SELECT")?.rows)
+        XCTAssertNil(PostgresQueryMetadata(string: "SELECT 100 100"))
+    }
```
fabianfett, line 23: Would you mind creating a new test file for this test:
```
PostgresNIOTests/PostgresQueryMetadataTests.swift
```

Since it isn’t really related to PSQLData.

fabianfett (commented):
🚀 Thanks very much, for taking care of this. Let’s get this into the right form and land this!

Sources/PostgresNIO/PostgresDatabase+Query.swift
grennis: Yes, makes sense! Done

Tests/PostgresNIOTests/New/PSQLDataTests.swift
grennis: Fixed!

Sources/PostgresNIO/PostgresDatabase+Query.swift
grennis: I did do it different from your comment since I figured if the row count is present, we should parse it. But if that's not important (I really have no idea), I'm happy to just set rows to nil. Please let me know.

Sources/PostgresNIO/PostgresDatabase+Query.swift
```diff
@@ -74,7 +74,16 @@ public struct PostgresQueryMetadata {
...
+        case "SELECT":
+            // AWS Redshift does not return the actual row count as defined in the postgres wire spec:
+            // https://www.postgresql.org/docs/13/protocol-message-formats.html in section `CommandComplete`
+            guard parts.count == 1 || parts.count == 2 else {
+                return nil
+            }
+            self.command = .init(parts[0])
+            self.oid = nil
+            self.rows = parts.count == 2 ? Int(parts[1]) : 0
+        case "DELETE", "UPDATE", "MOVE", "FETCH", "COPY":
```
fabianfett, line 86: Small nit. Let’s make this a special case that we handle.

```suggestion
        case "SELECT" where parts.count == 1:
            // AWS Redshift does not return the actual row count as defined in the postgres wire spec for SELECT:
            // https://www.postgresql.org/docs/13/protocol-message-formats.html in section `CommandComplete`
            self.command = "SELECT"
            self.oid = nil
            self.rows = nil
        case "SELECT", "DELETE", "UPDATE", "MOVE", "FETCH", "COPY":
```

Tests/PostgresNIOTests/New/PSQLDataTests.swift
fabianfett: Awesome! Thanks!

Sources/PostgresNIO/PostgresDatabase+Query.swift
grennis: Thats much better! Thanks

Tests/PostgresNIOTests/New/PSQLMetadataTests.swift
```diff
@@ -0,0 +1,17 @@
+import NIO
+import XCTest
+@testable import PostgresNIO
+
+class PSQLMetadataTests: XCTestCase {
+    func testSelect() {
+        XCTAssertEqual(100, PostgresQueryMetadata(string: "SELECT 100")?.rows)
+        XCTAssertEqual(0, PostgresQueryMetadata(string: "SELECT")?.rows)
```
fabianfett, line 8: uff, this tests needs fixing as well!

```suggestion
        XCTAssertEqual(nil, PostgresQueryMetadata(string: "SELECT")?.rows)
```

fabianfett (approved):

Tests/PostgresNIOTests/New/PSQLMetadataTests.swift
grennis: Tests fixed!

0xTim (approved):
LGTM. It's a pity there's no Redshift CI available in GH actions so for now we'll have to trust manual testing

PR #169, opened 2021-08-12
Splits CI Jobs Into Unit And Integration Tests On Linux

### LINKED ISSUE #165: Update CI for wider test coverage
https://github.com/vapor/postgres-nio/pull/157#pullrequestreview-655859197

dnadoba:
This was prepared by #157.  
CI now also uses more swift compiler versions, namely `5.4`, `nightly-5.5` and `nightly-main`.
It also removes `amazonlinux2` and now only tests on `focal`.

fixes #165

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,5 @@
 name: test
-on: [ 'push' ]
+on: [ 'pull_request' ]
 env:
   LOG_LEVEL: notice
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,5 @@
 name: test
-on: [ 'pull_request' ]
+on: [ 'push' ]
 env:
   LOG_LEVEL: notice
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,8 +66,8 @@ jobs:
           - swift:5.2
           - swift:5.3
           - swift:5.4
-          - swiftlang:swift/5.5-nightly
-          - swiftlang:swift/main-nightly
+          - swiftlang:swift/nightly-5.5
+          - swiftlang:swift/nightly-main
         swiftos:
           #- xenial
           #- bionic
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,7 @@ jobs:
         swiftver:
           - 5.2
           - 5.3
+          - 5.4
         dbimage:
           - postgres:13
           - postgres:12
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -55,9 +56,35 @@ jobs:
...
+          - swift:5.4
+          - swiftlang/swift:nightly-5.5
+          - swiftlang/swift:nightly-main
+        swiftos:
+          #- xenial
+          #- bionic
+          - focal
+          #- centos7
+          #- centos8
+          #- amazonlinux2
```
0xTim, line 77: I think for the unit tests we should be testing against _all_ supported OS versions rather than just one.

.github/workflows/test.yml
```diff
@@ -55,9 +56,35 @@ jobs:
...
+  linux-unit-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        swiftver:
+          - swift:5.2
+          - swift:5.3
+          - swift:5.4
+          - swiftlang/swift:nightly-5.5
+          - swiftlang/swift:nightly-main
```
0xTim, line 70: This has traditionally caused issues with our CI as we find that bugs in nightly cause our tests to fail and block PRs. I think it would be good to have a separate workflow that runs daily that tests the nightlies

.github/workflows/test.yml
```diff
@@ -55,9 +56,35 @@ jobs:
...
+          #- centos7
+          #- centos8
+          #- amazonlinux2
+    container: ${{ format('{0}-{1}', matrix.swiftver, matrix.swiftos) }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Run tests with Thread Sanitizer
+        run: swift test --enable-test-discovery --sanitize=thread --filter=^PostgresNIOTests
```
0xTim, line 84: I wonder if it's worth trying to run these tests in parallel

0xTim (changes requested):
Thanks for doing this - this is an awesome start! We definitely need to expand the testing to more OSes. I'll let @gwynne chime in as well since she's been overhauling the rest of our CI

.github/workflows/test.yml
gwynne: The reason I disabled the other OS' was that there are several workarounds necessary for, e.g., SPM support on CentOS 7, SQLite support on AmazonLinux2, etc., plus the full test matrix with all versions _and_ OS' yields well over 100 tests _per push_ - it was just too many, and it was causing stalls for CI in other repos since the "workflows in progress at once" limit is per-org, not per-repo.

.github/workflows/test.yml
gwynne: This is something I was working on last time I had a chance to deal with the CI for more than trivial updates, yeah.

.github/workflows/test.yml
gwynne: (Yes, I realize the matrix given here would only yield 30 tests, but then also multiply that by the multiple postgresql versions - with 11, 12, and 13, that was already 90 tests right there. And then also two more (latest and latest-stable) per PSQL version for macOS gives 100, and at the time I was also trying to get both Catalina and Big Sur macOS tests going... it starts really adding up.)

.github/workflows/test.yml
gwynne: Anything that connects to the database can't run in parallel unless we start having tests generate randomly-named databases (or at least schemas).

0xTim: The PostgresNIOTests don't touch a real DB they're all in the integration tests

.github/workflows/test.yml
0xTim: Cool fine by me

0xTim (approved):

PR #173, opened 2021-09-18
Add conveniences to PSQLFrontendMessage

fabianfett:
### Motivation

We want to use the `NIOSingleStepByteToMessageDecoder` within the `PSQLChannelHandler` in the future. For this reason it is important that we will be able to decode PSQLFrontendMessages for tests in the future.

### Changes

- Rename protocol `PSQLFrontendMessagePayloadEncodable` to `PSQLMessagePayloadEncodable`
- `PSQLFrontendMessage.ID` is now `RawRepresentable`

### Result

Code that makes testing easier in the future.

Note: Please don't auto release this. Follow up prs incoming.

### DIFF
--- a/Sources/PostgresNIO/New/PSQLFrontendMessage.swift
+++ b/Sources/PostgresNIO/New/PSQLFrontendMessage.swift
@@ -20,8 +20,7 @@ enum PSQLFrontendMessage {
     case startup(Startup)
     case terminate
     
-    enum ID: RawRepresentable, Equatable {
-        typealias RawValue = UInt8
+    enum ID: UInt8, Equatable {
         
         case bind
         case close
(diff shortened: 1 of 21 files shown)


Sources/PostgresNIO/New/PSQLFrontendMessage.swift
```diff
@@ -22,7 +20,9 @@ enum PSQLFrontendMessage {
     case startup(Startup)
     case terminate
     
-    enum ID {
+    enum ID: RawRepresentable, Equatable {
+        typealias RawValue = UInt8
```
gwynne, line 24:
```suggestion
    enum ID: UInt8, Equatable {
```

gwynne (approved):
Just one style nit, otherwise looks great!

PR #175, opened 2021-09-18
Add PSQLBackendMessageEncoder

fabianfett:
### Motivation

To test a `PSQLChannelHandler`, that uses an internal `NIOSingleStepByteToMessageDecoder`, in an `EmbeddedChannel` we need to writeInbound bytes. To make this easier, this PR introduces a `PSQLBackendMessageEncoder` as a test util

### Changes

- Add `PSQLBackendMessageEncoder`
- Use `PSQLBackendMessageEncoder` in authentication tests

### Note

This pr only changes test code

### DIFF
--- a/Tests/PostgresNIOTests/New/Extensions/PSQLBackendMessageEncoder.swift
+++ b/Tests/PostgresNIOTests/New/Extensions/PSQLBackendMessageEncoder.swift
@@ -12,12 +12,10 @@ struct PSQLBackendMessageEncoder: MessageToByteEncoder {
     func encode(data message: PSQLBackendMessage, out buffer: inout ByteBuffer) throws {
         switch message {
         case .authentication(let authentication):
-            buffer.writeBackendMessageID(message.id)
-            self.encodePayload(authentication, into: &buffer)
+            self.encode(messageID: message.id, payload: authentication, into: &buffer)
             
         case .backendKeyData(let keyData):
-            buffer.writeBackendMessageID(message.id)
-            self.encodePayload(keyData, into: &buffer)
+            self.encode(messageID: message.id, payload: keyData, into: &buffer)
             
         case .bindComplete,
              .closeComplete,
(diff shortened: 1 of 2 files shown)


Tests/PostgresNIOTests/New/Extensions/PSQLBackendMessageEncoder.swift
```diff
@@ -0,0 +1,279 @@
...
+    
+    private struct EmptyPayload: PSQLMessagePayloadEncodable {
+        func encode(into buffer: inout ByteBuffer) {}
+    }
+    
+    private struct StringPayload: PSQLMessagePayloadEncodable {
+        var string: String
+        init(_ string: String) { self.string = string }
+        func encode(into buffer: inout ByteBuffer) {
+            buffer.writeNullTerminatedString(self.string)
```
gwynne, line 83: Okay, it's not your fault, but this is reminding me of a peeve that's been annoying the <censored> out of me for some time. Can we:
1) Move the multiple copies of various `ByteBuffer` utilities scattered around PostgresNIO, MySQLNIO, etc. into something they all use, like AsyncKit.
2) Fix the `writeFoo()` utilities to work the same way as `ByteBuffer`'s builtins - specifically in returning the number of bytes written, and preferably being based on `setFoo()` variants of themselves as well.
3) Having done those, no longer have to do things like `buffer.writerIndex - startIndex`.
4) Does this actually need to be an `encode()`-style method/protocol? Why can't it be something like `func write(into buffer: inout ByteBuffer) -> Int`? No generic-ness is lost, and there's no pretense of `Codable`-like semantics in a place where they have no real business being anyway.

(Again, there's nothing actually _wrong_ with the code you've written - it's just still trapped in some of the crummy code smells Tanner left lying all over the place 😰.)

Tests/PostgresNIOTests/New/Extensions/PSQLBackendMessageEncoder.swift
```diff
@@ -0,0 +1,279 @@
...
+
+    /// Called once there is data to encode.
+    ///
+    /// - parameters:
+    ///     - data: The data to encode into a `ByteBuffer`.
+    ///     - out: The `ByteBuffer` into which we want to encode.
+    func encode(data message: PSQLBackendMessage, out buffer: inout ByteBuffer) throws {
+        switch message {
+        case .authentication(let authentication):
+            buffer.writeBackendMessageID(message.id)
```
gwynne, line 15: There is nothing specifically wrong with having it this way, but it nags at me to have the same line repeated in (almost) every single `case` like this over and over. Is there any not-horrible way to factor it out? (I can't think of something off the top of my head, and it's certainly not worth spending any amount of time on.)

Tests/PostgresNIOTests/New/Extensions/PSQLBackendMessageEncoder.swift
```diff
@@ -0,0 +1,279 @@
...
+            return .parseComplete
+        case .portalSuspended:
+            return .portalSuspended
+        case .readyForQuery:
+            return .readyForQuery
+        case .rowDescription:
+            return .rowDescription
+        case .sslSupported,
+             .sslUnsupported:
+            preconditionFailure("Message has no id.")
```
gwynne, line 135: What guarantees in practical terms (a runtime crash is not ideal) that no one will ever ask for the `id` of a `ssl[Un]supported` message? It's feeling a lot like these two messages are really not of the same "type" as the rest.

gwynne (commented):
I'm neither approving nor requesting changes for now on this one - I'm not convinced any of my comments are really worth spending effort on, but I'd like to get the sense of your thoughts before moving on.

Tests/PostgresNIOTests/New/Extensions/PSQLBackendMessageEncoder.swift
fabianfett: I think we should keep `PostgresNIO` dependency free except for the NIO and Crypto dependencies.

Tests/PostgresNIOTests/New/Extensions/PSQLBackendMessageEncoder.swift
fabianfett: Transformed into one liners.

Tests/PostgresNIOTests/New/Extensions/PSQLBackendMessageEncoder.swift
fabianfett: It being in a test target :)

gwynne (approved):
LGTM!

PR #176, opened 2021-09-18
Add RowStreamStateMachine

fabianfett:
### Motivation

To best support back-pressure, we extract the necessary state machine into a dedicated substate machine.

### Changes

- Add RowStreamStateMachine

### Note

This pr does not include any tests and the new code will not be hit. We will add tests to the `ExtendedQueryStateMachine` that test the new functionality once we wire the new substate machine up.

### DIFF
--- a/Sources/PostgresNIO/New/Connection State Machine/RowStreamStateMachine.swift
+++ b/Sources/PostgresNIO/New/Connection State Machine/RowStreamStateMachine.swift
@@ -1,0 +1,163 @@
+import NIOCore
+
+/// A sub state for receiving data rows. Stores whether the consumer has either signaled demand and whether the
+/// channel has issued `read` events.
+///
+/// This should be used as a SubStateMachine in QuerySubStateMachines.
+struct RowStreamStateMachine {
+    
+    enum Action {
+        case read
+        case wait
+    }
+    
+    private enum State {
+        /// The state machines expects further writes to `channelRead`. The writes are appended to the buffer.
+        case waitingForRows(CircularBuffer<PSQLBackendMessage.DataRow>)
+        /// The state machines expects a call to `demandMoreResponseBodyParts` or `read`. The buffer is
+        /// empty. It is preserved for performance reasons.
+        case waitingForReadOrDemand(CircularBuffer<PSQLBackendMessage.DataRow>)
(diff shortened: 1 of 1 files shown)


Sources/PostgresNIO/New/Connection State Machine/RowStreamStateMachine.swift
```diff
@@ -0,0 +1,163 @@
...
+            
+        case .waitingForDemand(var buffer):
+            self.state = .modifying
+            buffer.append(newRow)
+            self.state = .waitingForRows(buffer)
+            
+        case .waitingForReadOrDemand(var buffer):
+            self.state = .modifying
+            buffer.append(newRow)
+            self.state = .waitingForRows(buffer)
```
gwynne, line 62: Feel free to "tweak" my "code style" (generous to call it that, I agree) as you find apropos, I mostly just care about the code de-duplication.
```suggestion
        case .waitingForRows(var buffer),
             // For all the following cases, please note:
             // Normally these code paths should never be hit. However there is one way to trigger
             // this:
             //
             // If the server decides to close a connection, NIO will forward all outstanding
             // `channelRead`s without waiting for a next `context.read` call. For this reason we might
             // receive new rows, when we don't expect them here.
             .waitingForRead(var buffer),
             .waitingForDemand(var buffer),
             .waitingForReadOrDemand(var buffer):
            self.state = .modifying
            buffer.append(newRow)
            self.state = .waitingForRows(buffer
```

Sources/PostgresNIO/New/Connection State Machine/RowStreamStateMachine.swift
```diff
@@ -0,0 +1,163 @@
...
+            // from the consumer
+            return .wait
+
+        case .modifying:
+            preconditionFailure("Invalid state: \(self.state)")
+        }
+    }
+
+    mutating func end() -> CircularBuffer<PSQLBackendMessage.DataRow> {
+        switch self.state {
```
gwynne, line 143: Yeah, like this! This is better than my suggestion above, do it this way there too!

gwynne (approved):
LGTM , one style nit I didn't even handle correctly anyway 😅

PR #177, opened 2021-09-18
Extract PSQLRow

fabianfett:
### Motivation

In #135, I did some naming things just wrong. `PSQLRows` is a stupid name. We should name it `PSQLRowStream`. `PSQLRows.Row` is a stupid name for a single table row. We should name it `PSQLRow`. Transforming an incoming data row packet to a `[PSQLData]` array early is expensive and stupid. Let's not do this anymore.

### Changes

Consume this pr in the three change steps (separate commit for each):

1. Extract `PSQLRows.Row` to `PSQLRows` (Got its own file). Stop the early `[PSQLData]` madness.
2. Rename `PSQLRows` to `PSQLRowStream`
3. Fix naming in integration tests to match `PSQLRowStream `

### Result

I feel less embarrassed about my code. It might be a little faster.

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection+Database.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection+Database.swift
@@ -78,7 +78,7 @@ internal enum PostgresCommands: PostgresRequest {
     }
 }
 
-extension PSQLRows {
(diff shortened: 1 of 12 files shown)


gwynne (approved):
LGTM

PR #178, opened 2021-09-18
Add PSQLFrontendMessageDecoder

fabianfett:
### Motivation

In tests we would like to use `EmbeddedChannel`, if we decode and encode in the `PSQLChannelHandler` using the `NIOSingleStepByteToMessageDecoder`, we need to decode messages the handler wrote to the channel from bytes.

### Changes

- Add `PSQLFrontendMessageDecoder`
- Add `ReverseByteToMessageHandler`

### Notes

The changes are mostly in the test target. Enabling easier tests in the future.

### DIFF
--- a/Tests/PostgresNIOTests/New/Extensions/PSQLFrontendMessageDecoder.swift
+++ b/Tests/PostgresNIOTests/New/Extensions/PSQLFrontendMessageDecoder.swift
@@ -1,0 +1,172 @@
+@testable import PostgresNIO
+
+struct PSQLFrontendMessageDecoder: NIOSingleStepByteToMessageDecoder {
+    typealias InboundOut = PSQLFrontendMessage
+    
+    private(set) var isInStartup: Bool
+    
+    init() {
+        self.isInStartup = true
+    }
+    
+    mutating func decode(buffer: inout ByteBuffer) throws -> PSQLFrontendMessage? {
+        // make sure we have at least one byte to read
+        guard buffer.readableBytes > 0 else {
+            return nil
+        }
+        
+        if self.isInStartup {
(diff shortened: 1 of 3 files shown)


fabianfett:
Urgh! I messed up GIT. Will need another force push.

Tests/PostgresNIOTests/New/Extensions/PSQLFrontendMessageDecoder.swift
```diff
@@ -0,0 +1,176 @@
...
+    }
+    
+    mutating func decode(buffer: inout ByteBuffer) throws -> PSQLFrontendMessage? {
+        // make sure we have at least one byte to read
+        guard buffer.readableBytes > 0 else {
+            return nil
+        }
+        
+        if self.isInStartup {
+            guard buffer.readableBytes >= 4 else {
```
gwynne, line 19: This and the other guards on readableBytes are redundant; `getInteger(at:as:)` already does this:

```swift
        guard let range = self.rangeWithinReadableBytes(index: index, length: MemoryLayout<T>.size) else {
            return nil
        }
```
So for that read you can just check that result for `nil` instead of having a separate check. All other `read*()` and `get*()` methods also do this, so you can simplify _and_ get rid of the `!`s.

This all being said, I very, very vaguely once recall there possibly having been a reason _not_ to do that... I have no idea what it was or if it was a real thing and not my fevered imagining, but if so, please tell me 😰

gwynne (changes requested):
Just the one commentary on dealing with the ByteBuffer, if there's a reason to leave it as is I'm not gonna try to argue the point 🙂

gwynne (approved):
LGTM

PR #179, opened 2021-09-18
Add BufferedMessageEncoder, fix PSQLFrontendMessageEncoder

fabianfett:
### Motivation

- We want to communicate as efficiently as possible to the server. For this reason we use message pipelining within a query.
- If we drop the current use of `MessageToByteHandler` and do the encoding in the `PSQLChannelHandler`, we can write to a single ByteBuffer and only forward this together with a flush

### Changes

- Add `BufferedMessageEncoder`
- Rename `PSQLFrontendMessage.Encoder` to `PSQLFrontendMessageEncoder` (got its own file)


gwynne (approved):
LGTM

PR #180, opened 2021-09-19
Batch rows for consumption

fabianfett:
### Motivation

To allow faster processing of incoming `DataRow`s, we should batch them up in channelRead events and forward them as a batch for consumption in `PSQLRowStream`. This work is the foundation for AsyncSequence support in the future.

### Modifications

- Extends `ExtendedQueryStateMachine` to use `RowStreamStateMachine` internally
- Refactor `PSQLRowStream` to work with batches of rows.


gwynne (approved):
LGTM; I could offer a few dozen style nits, but everything looks functionally great, and I don't care about the style enough to hold things up for it 🙂

PR #187, opened 2021-10-11
Fix PostgresNumeric.init crash at large number

### LINKED ISSUE #186: Runtime crush at PostgresNumeric.init
### Describe the bug

`PostgresNumeric` cannot decode large number (like `1234567890123`) and it causes runtime crash.

### To Reproduce

1. call this line 

```swift
PostgresNumeric(string: "1234567890123")
```

2. crash!

### Expected behavior

Can call initializer without crash.

### Environment

* Vapor Framework version: 4.48.8
* postgres-nio: 1.6.2

### Additional context

https://github.com/vapor/postgres-nio/blob/046d3ba1a40c5d2e2457177f4311b0b9c97f8945/Sources/PostgresNIO/Data/PostgresData%2BNumeric.swift#L270

this function does not work correctly and return values over `maxSize`.

```swift
let uuu = "1234567890123"
print(uuu.reverseChunked(by: 4)) // ["1", "2345", "67890", "123"]
```

then here, `Int16(67890)!` causes crash.

https://github.com/vapor/postgres-nio/blob/046d3ba1a40c5d2e2457177f4311b0b9c97f8945/Sources/PostgresNIO/Data/PostgresData%2BNumeric.swift#L91-L95

sidepelican:
Fix #186

Fix `reverseChunked` to comply with `maxSize`.

### DIFF
--- a/Sources/PostgresNIO/Data/PostgresData+Numeric.swift
+++ b/Sources/PostgresNIO/Data/PostgresData+Numeric.swift
@@ -268,16 +268,13 @@ private extension Collection {
     // splits the collection into chunks of the supplied size
     // if the collection is not evenly divisible, the first chunk will be smaller
     func reverseChunked(by maxSize: Int) -> [SubSequence] {
-        var lastDistance = 0
+        let firstChunkSize = self.count % maxSize
+        if firstChunkSize == 0 {
+            return self.chunked(by: maxSize)
+        }
         var chunkStartIndex = self.startIndex
-        return stride(from: 0, to: self.count, by: maxSize).reversed().map { current in
-            let distance = (self.count - current) - lastDistance
-            lastDistance = distance
-            let chunkEndOffset = Swift.min(
-                self.distance(from: chunkStartIndex, to: self.endIndex),
-                distance
-            )
-            let chunkEndIndex = self.index(chunkStartIndex, offsetBy: chunkEndOffset)
+        return stride(from: firstChunkSize, through: self.count, by: maxSize).map { current in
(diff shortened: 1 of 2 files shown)


omochi:
Hi, I am coworker with @sidepelican .
We are using vapor and postgres for our company business.
We depends on `numeric` type of postgres for accurate amount of money processing on DB.
I wish to grow vapor on enterprise use-cases 😁

gwynne (changes requested):
I suggest replacing both `chunked()` _and_ `reverseChunked()` with these simpler versions:

```swift
    /// Split the collection into chunks of `maxSize` length each. The last chunk may contain anywhere between 1 and maxSize elements. `maxSize` must be greater than zero.
    func chunked(by maxSize: Int) -> [SubSequence] {
        return stride(from: self.startIndex, to: self.endIndex, by: maxSize).map {
            self[$0 ..< (self.index($0, offsetBy: maxSize, limitedBy: self.endIndex) ?? self.endIndex)]
        }
    }

    /// Split the collection into chunks of `maxSize` length each. The **first** chunk may contain anywhere between 1 and maxSize elements. `maxSize` must be greater than zero.
    func reverseChunked(by maxSize: Int) -> [SubSequence] {
        return stride(from: self.endIndex, to: self.startIndex, by: -maxSize).map {
            self[(self.index($0, offsetBy: -maxSize, limitedBy: self.startIndex) ?? self.startIndex) ..< $0]
        }.reversed()
    }
```

ptoffy:
This was superseded by https://github.com/vapor/postgres-nio/pull/465

PR #191, opened 2021-11-11
Rewire CI a bit

gwynne:
- Use a better test matrix
- Add code coverage collection from the unit tests
- Disable TSan for Concurrency reasons
- Use rpath workaround for macOS tests
- Match things up in general with postgres-kit's CI

_Only affects CI, no release needed._


jdmcd (approved):
LGTM

PR #194, opened 2021-11-13
Add proper support for `Decimal`

madsodgaard:
Fixes an issue when trying to decode/encode `numeric` columns as `Decimal`, that would result in an error such as: `server: column "balance" is of type numeric but expression is of type text (transformAssignedExpr)`.


fabianfett:
Hi @madsodgaard,

generally I'm open to accept this change. However there are a number of things that I would like to see, if we want to continue:

1. Also add conformance of Decimal to PSQLEncodable and PSQLDecodable
2. Unit tests (locally confirming that encoding and decoding to the raw value works)
3. Integration tests (sending a bind payload with the value and retrieving it)

I'm happy to help at every step of the way, if you need help :)

madsodgaard:
@fabianfett I have added conformance to `PSQLCodable` and some unit tests and a integration test, let me know if anything is missing or can be improved

fabianfett (approved):
LGTM. Generally we should probably rework `PostgresNumeric` in the future... But let's ask @gwynne for a review as well!

gwynne (changes requested):
I'm concerned about the change in the public API behavior - will this break code at runtime for people who are currently storing `Decimal`s in `TEXT` columns? Can we add a test or two which explicitly demonstrates this scenario, for both reading (select) and writing (insert/update)? (Ideally we'd have the test both here at the raw Postgres level operating directly on these API and also in FluentBenchmarker ensuring the behavior is stable for the common usage.)

madsodgaard:
@gwynne It should not. I added a integration test to this library for testing string serialization. `FluentBenchmark` actually already contains a test for this in: [PerformanceTests](https://github.com/vapor/fluent-kit/blob/main/Sources/FluentBenchmark/Tests/PerformanceTests.swift). That is not the most explicit place for asserting this behaviour though.

gwynne (approved):
Okay, I'm sufficiently satisfied that this is backwards-compatible - I'm not 100%, but I'm mindful of being a bit paranoid, and given that text-column round-tripping still works, I can't think of any _actual_ objection that warrants holding back this long-overdue missing functionality 🙂

gwynne:
Merging as semver-minor - though there are no new public APIs, just in case my paranoia isn't completely pointless it gives just a tiny bit of extra wiggle room 🤷‍♀️

gwynne:
Thanks again for the excellent work on this, @madsodgaard!

PR #188, opened 2021-10-19
[DRAFT] Make async queries possible

fabianfett:
This is a very early draft enabling async queries on a Postgres connection. It allows async/await in Postgres:

```swift
// swift-tools-version:5.5
import PackageDescription

let package = Package(
    name: "postgres-perf",
    platforms: [
        .macOS(.v12)
    ],
    dependencies: [
        .package(url: "https://github.com/fabianfett/postgres-nio.git", .branch("ff-async2")),
    ],
    targets: [
        .executableTarget(name: "PostgresExample", dependencies: [
            .product(name: "PostgresNIO", package: "postgres-nio"),
        ]),
    ]
)
```

And then use it like this:

```swift
import NIO
import Logging
import PostgresNIO
import struct Foundation.Date

@main
enum AsyncExample {
    static func main() async throws {
        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
        let eventLoop = eventLoopGroup.next()
        defer { try! eventLoopGroup.syncShutdownGracefully() }
        
        var logger = Logger(label: "psql")
        logger.logLevel = .info
        
        let configuration = PSQLConnection.Configuration(
            host: "postgres", port: 5432,
            username: "postgres",
            database: "postgres",
            password: "postgres",
            tlsConfiguration: nil,
            coders: .foundation
        )
        let connection = try await PSQLConnection.connect(configuration: configuration, logger: logger, on: eventLoop)
        
        let start = Date()
        var i = 0
        var result = 0
        
        let stream = try await connection.query("SELECT * FROM large_test LIMIT 2000000", logger: logger)
        for try await (column1, column2, column3) in stream.decode(Int.self, Double.self, Double.self) {
            result &+= column1
            i &+= 1
        }
        
        try await connection.close()
    }
}
```

PR #198, opened 2021-11-19
DataRow without allocation; DataRow as Collection; RowDescription top level

fabianfett:
This is a cherry pick of #188.

### Modifications

- `DataRow` and `RowDescription` have been moved out of the `PSQLBackendMessage` namespace. This allows us to mark them as `@inlinable` or `@usableFromInline` at a later point, without marking everything in `PSQLBackendMessage` as `@inlinable`
- `DataRow` does not use an internal array for its columns anymore. Instead all read operations are directly done on its ByteBuffer slice.
- `DataRow` implements the `Collection` protocol now.

### Result

One allocation fewer per queried row.

### DIFF
--- a/Sources/PostgresNIO/New/Messages/DataRow.swift
+++ b/Sources/PostgresNIO/New/Messages/DataRow.swift
@@ -58,24 +58,9 @@ extension DataRow: Collection {
             self.offset = index
         }
         
-        static func == (lhs: Self, rhs: Self) -> Bool {
-            lhs.offset == rhs.offset
-        }
-        
+        // Only needed implementation for comparable. The compiler synthesizes the rest from this.
         static func < (lhs: Self, rhs: Self) -> Bool {
             lhs.offset < rhs.offset
-        }
-
-        static func <= (lhs: Self, rhs: Self) -> Bool {
-            lhs.offset <= rhs.offset
-        }
-
-        static func >= (lhs: Self, rhs: Self) -> Bool {
-            lhs.offset >= rhs.offset
-        }
-
-        static func > (lhs: Self, rhs: Self) -> Bool {
-            lhs.offset > rhs.offset
         }
     }
     
--- a/Sources/PostgresNIO/New/Messages/DataRow.swift
+++ b/Sources/PostgresNIO/New/Messages/DataRow.swift
@@ -1,34 +1,132 @@
 import NIOCore
 
-extension PSQLBackendMessage {
+/// A backend data row message.
+///
+/// - NOTE: This struct is not part of the ``PSQLBackendMessage`` namespace even
+///         though this is where it actually belongs. The reason for this is, that we want
+///         this type to be @usableFromInline. If a type is made @usableFromInline in an
+///         enclosing type, the enclosing type must be @usableFromInline as well.
+///         Not putting `DataRow` in ``PSQLBackendMessage`` is our way to trick
+///         the Swift compiler
+struct DataRow: PSQLBackendMessage.PayloadDecodable, Equatable {
     
-    struct DataRow: PayloadDecodable, Equatable {
+    var columnCount: Int16
+    
+    var bytes: ByteBuffer
+    
+    static func decode(from buffer: inout ByteBuffer) throws -> Self {
+        try buffer.ensureAtLeastNBytesRemaining(2)
+        let columnCount = buffer.readInteger(as: Int16.self)!
+        let firstColumnIndex = buffer.readerIndex
         
-        var columns: [ByteBuffer?]
-        
-        static func decode(from buffer: inout ByteBuffer) throws -> Self {
+        for _ in 0..<columnCount {
             try buffer.ensureAtLeastNBytesRemaining(2)
-            let columnCount = buffer.readInteger(as: Int16.self)!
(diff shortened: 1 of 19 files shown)


Sources/PostgresNIO/New/Messages/DataRow.swift
```diff
@@ -1,34 +1,132 @@
...
+            lhs.offset <= rhs.offset
+        }
+
+        static func >= (lhs: Self, rhs: Self) -> Bool {
+            lhs.offset >= rhs.offset
         }
+
+        static func > (lhs: Self, rhs: Self) -> Bool {
+            lhs.offset > rhs.offset
+        }
```
gwynne, line 79: Doesn't the runtime's default implementations of these operators for `Comparable` already provide these in terms of `<` and `==`? Also won't the compiler synthesize the `Equatable` conformance?

Sources/PostgresNIO/New/Messages/DataRow.swift
```diff
@@ -1,34 +1,132 @@
...
+    
+    func index(after index: ColumnIndex) -> ColumnIndex {
+        guard index < self.endIndex else {
+            preconditionFailure("index out of bounds")
+        }
+        var elementLength = Int(self.bytes.getInteger(at: index.offset, as: Int32.self)!)
+        if elementLength < 0 {
+            elementLength = 0
+        }
+        return ColumnIndex(index.offset + MemoryLayout<Int32>.size + elementLength)
```
gwynne, line 104: I don't know if the typical access pattern for a `DataRow` supports bothering with this, but wouldn't pre-calculating the buffer offsets in `.init()` then make it trivial to add `BidirectionalCollection` and `RandomAccessCollection` conformance? Or does caching the offsets itself carry enough overhead to negate the few allocations win?

Tests/PostgresNIOTests/New/Extensions/PSQLBackendMessage+Equatable.swift
```diff
@@ -1,4 +1,5 @@
 @testable import PostgresNIO
+import class Foundation.JSONEncoder
```
gwynne, line 2: Did Xcode annoyingly insert an unneeded `import` you didn't notice as it constantly does to me?

Tests/PostgresNIOTests/New/Extensions/PSQLBackendMessageEncoder.swift
```diff
@@ -188,19 +188,10 @@ extension PSQLBackendMessage.BackendKeyData: PSQLMessagePayloadEncodable {
...
-        for column in self.columns {
-            switch column {
-            case .none:
-                buffer.writeInteger(-1, as: Int32.self)
-            case .some(var writable):
-                buffer.writeInteger(Int32(writable.readableBytes))
-                buffer.writeBuffer(&writable)
-            }
-        }
+        buffer.writeInteger(Int16(self.columnCount))
```
gwynne, line 193: `self.columnCount` is already an `Int16`, so if you want to ensure being explicit about the type, why not use `.writeInteger(self.columnCount, as: Int16.self)` instead of ostensibly invoking `Int16.init<Int16>(_:)` (though the compiler hopefully inlines that away)?

Tests/PostgresNIOTests/New/Messages/DataRowTests.swift
```diff
@@ -20,18 +20,127 @@ class DataRowTests: XCTestCase {
...
+            nil,
+            ByteBuffer(),
+            ByteBuffer(repeating: 5, count: 10),
+            nil
+        )
+        
+        XCTAssertNil(dataRow.withContiguousStorageIfAvailable { _ -> Int in
+            XCTFail("DataRow does not have a contiguous storage")
+            return 123
+        })
```
gwynne, line 113:
```suggestion
        XCTAssertNil(dataRow.withContiguousStorageIfAvailable { _ in
            return XCTFail("DataRow should not be able to offer contiguous storage")
        })
```

gwynne (changes requested):
A few minor nits and questions, nothing serious.

Sources/PostgresNIO/New/Messages/DataRow.swift
fabianfett: We talked about this offline:

1. Currently we expose the user `PostgresRow`. `PostgresRow` translates the DataRow into an array of `PostgresData` in its initializer. For this reason `PostgresRow` is still a `RandomAccessCollection`, without any overhead.

2. In the future we probably want to support API's like proposed in #188:

    ```swift
    let stream = try await connection.query("SELECT * FROM large_test LIMIT 2000000", logger: logger)
    for try await (column1, column2, column3) in stream.decode(Int.self, Double.self, Double.self) {
        // do something with the row
    }
    ```

    In this scenario PSQLRow doesn't have to be a `RandomAccessCollection` and we shouldn't need to pay the price for it (another allocation).

Tests/PostgresNIOTests/New/Messages/DataRowTests.swift
fabianfett: The nil check only really works, if we return a value from the closure. If we return Void, the nil check doesn't do anything.

gwynne (approved):
LGTM now 🙂

PR #201, opened 2021-11-22
Refactor PSQLRowStream to make async/await easier

fabianfett:
### Motivation

`PSQLRowStream`'s current implementation is interesting. It should be better tested and easier to follow for async/await support later.

### Changes

- Make `PSQLRowStream`'s implementation more sensible
- Add unit tests for `PSQLRowStream`

### Result

Adding async/await support becomes easier.


gwynne (approved):
Looks fine, not sure I 100% understand the state machine, but when have I ever? 🤣

PR #203, opened 2021-11-22
Faster decoding, thanks to fewer bound checks.

fabianfett:
### Motivation

We like faster code. This removes a ton of unnecessary bounds checks.

### Result

Faster code

In draft since it requires a new NIO release.


gwynne (approved):
I love it, ship it! 🎉

PR #206, opened 2021-12-01
Add async option to PSQLRowStream

fabianfett:
### Motivation

To allow async/await querying we need to support AsyncSequences.

### Changes

- Adds `PSQLRowSequence`
- Adjusts `PSQLRowStream` to support `PSQLRowSequence`

### Result

- Internal async/await support added.


Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
+import NIOCore
+import NIOConcurrencyHelpers
+
+#if swift(>=5.5) && canImport(_Concurrency)
+struct PSQLRowSequence: AsyncSequence {
```
gwynne, line 5: Can you add a comment here explaining the use of the "value type wrapping a consumer with a reference type" pattern? You don't have to go and write one of those 5-page rambles the way I sometimes do (in fact, please, for the love of everything you hold dear- _don't_ do that! 😆), a couple lines would probably more than so.

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
...
+        switch completionAction {
+        case .succeed(let continuation):
+            continuation.resume(returning: nil)
+            
+        case .fail(let continuation, let error):
+            continuation.resume(throwing: error)
+            
+        case .none:
+            break
+        }
```
gwynne, line 143: Is there really no less mind-numbingly tedious way to forward the result to the completion than this? 😕 I'm not even snarking (for once 😅 - Having to scroll through this kind of boilerplate when dealing with bugs and such surprisingly demoralizing in Swift (at lest, for me).

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
...
+extension AsyncStreamConsumer {
+    struct StateMachine {
+        enum UpstreamState {
+            enum DemandState {
+                case canAskForMore
+                case waitingForMore(CheckedContinuation<PSQLRow?, Error>?)
+            }
+            
+            case initialized
+            case streaming(AdaptiveRowBuffer, PSQLRowStream, DemandState)
```
gwynne, line 226: Missing a doc comment for this case

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
...
+            /// The upstream has finished, but the downstream has not consumed all events.
+            case finished(AdaptiveRowBuffer, String)
+            /// The upstream has failed, but the downstream has not consumed the error yet.
+            case failed(Error)
+            /// The upstream has failed or finished and the downstream has consumed all events. Final state.
+            case consumed
+
+            case modifying
```
gwynne, line 234: This one too

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
...
+        }
+        
+        enum DownstreamState {
+            case sequenceCreated
+            case iteratorCreated
+        }
```
gwynne, line 240: Is another whole enum entirely necessary for something which can be represented by just a `Bool`? I recognize the potential value for readability, but a well-named boolean can cover this one decently enough IMO.

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
...
+        
+        var upstreamState: UpstreamState
+        var downstreamState: DownstreamState
+        
+        init() {
+            self.upstreamState = .initialized
+            self.downstreamState = .sequenceCreated
+        }
```
gwynne, line 248:
```suggestion
        var upstreamState = UpstreamState.initialized
        var downstreamState = DownstreamState.sequenceCreated
        
        init() {}
```

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
...
+            case finished(AdaptiveRowBuffer, String)
+            /// The upstream has failed, but the downstream has not consumed the error yet.
+            case failed(Error)
+            /// The upstream has failed or finished and the downstream has consumed all events. Final state.
+            case consumed
+
+            case modifying
+        }
+        
+        enum DownstreamState {
```
gwynne, line 237:
```suggestion
        enum DownstreamState: Hashable {
```
(for use further down)

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
...
+        
+        init() {
+            self.upstreamState = .initialized
+            self.downstreamState = .sequenceCreated
+        }
+        
+        mutating func buffered(_ buffer: CircularBuffer<DataRow>, upstream: PSQLRowStream) {
+            guard case .initialized = self.upstreamState else {
+                preconditionFailure("Invalid upstream state: \(self.upstreamState)")
+            }
```
gwynne, line 253:
```suggestion
            precondition(self.upstreamState == .initialized, "Invalid upstream state: \(self.upstreamState)")
```

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
...
+            let adaptive = AdaptiveRowBuffer(buffer)
+            self.upstreamState = .streaming(adaptive, upstream, buffer.isEmpty ? .waitingForMore(nil) : .canAskForMore)
+        }
+        
+        mutating func finished(_ buffer: CircularBuffer<DataRow>, commandTag: String) {
+            guard case .initialized = self.upstreamState else {
+                preconditionFailure("Invalid upstream state: \(self.upstreamState)")
+            }
```
gwynne, line 261:
```suggestion
            precondition(self.upstreamState == .initialized, "Invalid upstream state: \(self.upstreamState)")
```

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
...
+            let adaptive = AdaptiveRowBuffer(buffer)
+            self.upstreamState = .finished(adaptive, commandTag)
+        }
+        
+        mutating func failed(_ error: Error) {
+            guard case .initialized = self.upstreamState else {
+                preconditionFailure("Invalid upstream state: \(self.upstreamState)")
+            }
```
gwynne, line 269: And so on 😅

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,607 @@
...
+        
+        enum SequenceDeinitializedAction {
+            case cancelStream(PSQLRowStream)
+            case none
+        }
+        
+        mutating func sequenceDeinitialized() -> SequenceDeinitializedAction {
+            switch (self.downstreamState, self.upstreamState) {
+            case (.sequenceCreated, .initialized):
+                preconditionFailure()
```
gwynne, line 290:
```
error: no exact matches in call to global function 'preconditionFailure'
preconditionFailure()
^

note: candidate requires that 'message' not be unspecified (requirement specified as 'If it's not in the SSWG Guidelines, it should be!')
func preconditionFailure(_ message: @autoclosure () -> String = String(), file: StaticString = #file, line: UInt = #line) -> Never
     ^
```

gwynne (changes requested):
Aside from some minor nits, what amount of the state machine my brain can keep in its working set at one time seems fine, I have no issue with it.

Sources/PostgresNIO/New/PSQLRowSequence.swift
fabianfett: Short story is to prepare for eventual move only types. We only use the reference type internally to signal when the Sequence was released. Should I add this in code?

Sources/PostgresNIO/New/PSQLRowSequence.swift
fabianfett: I'm not sure I get what you would like to see here?

Sources/PostgresNIO/New/PSQLRowSequence.swift
fabianfett: As this is mostly a style nit, how serious are you about it?

Sources/PostgresNIO/New/PSQLRowSequence.swift
fabianfett: Yes! I prefer an explicit enum about a well documented bool, for a single reason... Expressive code is far better than documented unexpressive code. There is no chance of the expressive code becoming unsynced with the documentation.

Sources/PostgresNIO/New/PSQLRowSequence.swift
gwynne: I tend to consider `assertionFailure()` and `preconditionFailure()` to be there as fallbacks: "if for whatever reason your condition can't be expressed in an autoclosure, here's your escape hatch". I make a point of only using them if the check _can't_ be written as an inline boolean expression (or if doing so would be impractical, like nested ternaries or such). Aside the conceptual reason, IMO it makes it easier to tell the difference between enforcing invariants and normal code flow.

In this particular case, there's also the bloody awful backwards-arse `case .enumerator = value` syntax to reckon with - it has its place, primarily when dealing with associated values, but for trivial [in]equality enum comparisons I consider `Equatable` conformance to be a divine gift (yeah, I'm just a touch opinionated about the `case let` syntax, does it show? 😅). It especially bugs me when it becomes a reason for being forced to use `preconditionFailure()` 😆 

To answer your question, I mean, yeah, you're right, it's a style nit, I imagine it probably compiles down to the same IR in the end (wouldn't be surprised if it was a 1:1 correspondence instruction for instruction in fact) - but it hits on two significant questions of code readability and conceptual consistency simultaneously. I'm not gonna die on the hill, but this is one I definitely take seriously.

Sources/PostgresNIO/New/PSQLRowSequence.swift
gwynne: I wouldn't mind a short comment about it, don't need to get into detail.

Sources/PostgresNIO/New/PSQLRowSequence.swift
```diff
@@ -0,0 +1,609 @@
+import NIOCore
+import NIOConcurrencyHelpers
+
+#if swift(>=5.5) && canImport(_Concurrency)
+/// An async sequence of ``PSQLRow``s.
+///
+/// - NOTE:This is a struct to allow us to move to a move only type easily once they become available.
```
gwynne, line 7: I am legally (though IANAL) obligated to offer this utterly irrelevant correction:
```suggestion
/// - Note: This is a struct to allow us to move to a move only type easily once they become available.
```

Sources/PostgresNIO/New/PSQLRowSequence.swift
fabianfett: I like the more verbose way of doing things here... My brain hurts when we use short syntax, and I want to separate concerns explicitly. Since I have written most of this and am probably the one to fix any issues, I'd prefer if we keep it as is.

PR #207, opened 2021-12-13
Add missing debug logging of raw queries

gwynne:
Raw queries and their bindings are now logged at the `.debug` log level, as with MySQLNIO.


gwynne:
@fabianfett has correctly pointed out that this is the wrong place for this fix, and that MySQLNIO should not have this logging at this level either. Do not merge.

PR #209, opened 2022-02-02
Cleanup PSQLError

fabianfett:
### Motivation

If a postgres query fails, users want to see which query failed. For this reason it is highly valueable to add file and line information to the new PSQLError.

### Changes

- PSQLError has a file and line property
- PSQLError has a code property
- Removed decode methods from PSQLData as those don't work anyway
- Fixed tests

### Result

- Much nicer PSQLError
- More specific errors when decoding data
- Further perf improvements when decoding unlocked.


gwynne (changes requested):
While I'm generally loudly in favor of making errors more traceable and information-laden, I don't see this adding much value from an API standpoint, since it can only capture file/line information that end users aren't very likely to be able to do anything with at all - and I don't think it adds enough value to internal debugging to be worth it. I miss when Console had less logging in it 🤣

PR #211, opened 2022-02-13
PSQLDecodingError improvements

fabianfett:
Split out to #209.

### Motivation

Currently we hand a large `PSQLDecodingContext` into the `PSQLDecodable`'s `decode` method. Lets look at it, as it is today:

```swift 
struct PSQLDecodingContext {
    let jsonDecoder: PSQLJSONDecoder
    
    let columnIndex: Int
    let columnName: String
    
    let file: String
    let line: Int
    
    init(jsonDecoder: PSQLJSONDecoder, columnName: String, columnIndex: Int, file: String, line: Int) {
        self.jsonDecoder = jsonDecoder
        self.columnName = columnName
        self.columnIndex = columnIndex
        
        self.file = file
        self.line = line
    }
}
```

We use most of the `PSQLDecodingContext`'s properties to create good error messages for users. Let's look at an implementation example:

```swift
    static func decode(from buffer: inout ByteBuffer, type: PSQLDataType, format: PSQLFormat, context: PSQLDecodingContext) throws -> Bool {
        guard type == .bool else {
            throw PSQLCastingError.failure(targetType: Self.self, type: type, postgresData: buffer, context: context)
        }
        
        switch format {
        case .binary:
            guard buffer.readableBytes == 1 else {
                throw PSQLCastingError.failure(targetType: Self.self, type: type, postgresData: buffer, context: context)
            }
            
            switch buffer.readInteger(as: UInt8.self) {
            case .some(0):
                return false
            case .some(1):
                return true
            default:
                throw PSQLCastingError.failure(targetType: Self.self, type: type, postgresData: buffer, context: context)
            }
        case .text:
            guard buffer.readableBytes == 1 else {
                throw PSQLCastingError.failure(targetType: Self.self, type: type, postgresData: buffer, context: context)
            }
            
            switch buffer.readInteger(as: UInt8.self) {
            case .some(UInt8(ascii: "f")):
                return false
            case .some(UInt8(ascii: "t")):
                return true
            default:
                throw PSQLCastingError.failure(targetType: Self.self, type: type, postgresData: buffer, context: context)
            }
        }
    }
```

As we can see the `context` is only used to create errors. If library adopters want to implement decoding for their own types, they need to forward the `context` as well, if they want to create good errors.

However we, as the Postgres library authors, don't need to pass the decoding info to the users and expect them to return it to us.

All decoding metadata can be kept in our code. For this reason users should return a `PSQLDecodingError.Code` from the decode method and we can create a full PSQLDecodingError based on this information. This also reduces code size quite a bit.

### Changes

- Only return `PSQLDecodingError.Code` from `PSQLDecodable` `decode` function.
- Remove decode methods from `PSQLData` as those don't work anyway.
- Change `Optional` decoding to something actually working.

### Result

- We can remove unused properties from PSQLDecodingContext
- Easier code for adopters to write
- Less code to maintain


Sources/PostgresNIO/New/PSQLCodable.swift
```diff
@@ -33,6 +34,23 @@ protocol PSQLDecodable {
...
+
+extension PSQLDecodable {
+    @inlinable
+    public static func decodeRaw(from byteBuffer: inout ByteBuffer?, type: PSQLDataType, format: PSQLFormat, context: PSQLDecodingContext) throws -> Self {
+        switch byteBuffer {
+        case .some(var buffer):
+            return try self.decode(from: &buffer, type: type, format: format, context: context)
+        case .none:
+            throw PSQLCastingError.Code.missingData
+        }
```
gwynne, line 57:
```suggestion
        guard let buffer = byteBuffer else {
            throw PSQLCastingError.Code.missingData
        }
        return try self.decode(from: &buffer, type: type, format: format, context: context)
```
IMO the `guard` makes the intent clearer when reading the code, I tend to reserve switching over options for more complex scenarios. Not a big thing, just a style nit.

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -80,59 +80,55 @@ struct PSQLError: Error {
...
+
+        var base: Base
+
+        init(_ base: Base) {
+            self.base = base
+        }
+
+        static let missingData = Self.init(.missingData)
+        static let typeMismatch = Self.init(.typeMismatch)
+        static let failure = Self.init(.failure)
```
gwynne, line 99: Do I correctly recognize this pattern as being that used by the various `Error` subtypes in `Foundation` (`CocoaError` et al, the ones that use a bunch of underscored protocols to make it all play more nicely with `NSError` bridging)?

Either way, is there any benefit to this pattern beyond that of sidestepping the "is adding enum cases in the absence of library evolution a semver-major break?" issue? Given that Vapor's answer to that question thus far has pretty consistently been "no" aside from a couple of specific cases, if that's the intent here, I'd rather just have an un-nested enum. (Though I'm willing to potentially be argued out of it.)

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -80,59 +80,55 @@ struct PSQLError: Error {
...
+        case .failure:
+            return """
+                Failed to cast Postgres data type \(self.postgresType.description) to Swift type \(self.targetType).
+                """
+        }
+
+    }
+    
+    static func ==(lhs: PSQLCastingError, rhs: PSQLCastingError) -> Bool {
+        lhs.targetType == rhs.targetType
```
gwynne, line 131: Okay, definitely have to ask why only this field is being considered for equality in this case, it's not the most ideal implementation for users AFAICT. I'll return to this if it makes more sense to me after reading the full PR.

Sources/PostgresNIO/New/PSQLRow.swift
```diff
@@ -54,10 +54,9 @@ extension PSQLRow {
             columnIndex: index,
             file: file,
             line: line)
-        
-        guard var cellSlice = self.data[column: index] else {
-            throw PSQLCastingError.missingData(targetType: T.self, type: column.dataType, context: context)
-        }
+
+        // Safe to force unwrap here, as we have ensured above, that the row has enough columns
```
gwynne, line 58:
```suggestion
        // Safe to force unwrap here, as we have ensured above that the row has enough columns 
```
Look up "pedantic" in the dictionary, you'll see a picture of me. That picture will be placed nearby the textual definition and most likely illustrate me in the process of correcting some truly inane and unimportant grammatical error, or of significantly overexplaining a simple concept.

(So, what do you see a picture of if you look up "comedian"? 😇)

gwynne (changes requested):
No actual serious changes need making, just want to be sure I'm clear on what's going on in a couple of spots before moving ahead.

Sources/PostgresNIO/New/PSQLError.swift
fabianfett: Yes, that’s the adding an enum case is a breaking change is the only reason… I would like to build PostgresNIO in the most resilient way possible, as I consider this lib very foundational. The more headaches we can prevent the better.

And if we can build in a way that allows us to move forward without breaking change, why don’t we take the opportunity?

Sources/PostgresNIO/New/PSQLError.swift
fabianfett: Oh god. Good catch. Fixed.

Sources/PostgresNIO/New/PSQLRow.swift
fabianfett: fixed.

Sources/PostgresNIO/New/PSQLCodable.swift
fabianfett: fixed

gwynne (approved):
Looks g2g! 👍 ❤️

PR #213, opened 2022-02-14
Merge PSQLDataType with PostgresDataType

fabianfett:
### Motivation

While rewriting lot's of PostgresNIO internals last year, I created new types that very closely match existing types. This was a mistake. I should have tried everything to reuse existing types, and change them as needed.

### Changes

- Remove `PSQLDataType` and replace all uses with `PostgresDataType`

### Result


Sources/PostgresNIO/Data/PostgresDataType.swift
```diff
@@ -16,7 +16,7 @@ public enum PostgresFormatCode: Int16, Codable, CustomStringConvertible {
 
 /// The data type's raw object ID.
 /// Use `select * from pg_type where oid = <idhere>;` to lookup more information.
-public struct PostgresDataType: Codable, Equatable, ExpressibleByIntegerLiteral, CustomStringConvertible, RawRepresentable {
```
fabianfett, line 19: Note: I have moved the Codable and ExpressibleByIntegerLiteral further down and added a remove `TODO` for the next major version break.

Sources/PostgresNIO/Data/PostgresDataType.swift
```diff
@@ -227,3 +222,13 @@ public struct PostgresDataType: Codable, Equatable, ExpressibleByIntegerLiteral,
         return self.knownSQLName ?? "UNKNOWN \(self.rawValue)"
     }
 }
+
+// TODO: The Codable conformance does not make any sense. Let's remove this with next major break.
+extension PostgresDataType: Codable {}
+
+// TODO: The ExpressibleByIntegerLiteral conformance does not make any sense and is not used anywhere. Remove with next major break.
```
gwynne, line 229: My guess would be that Tanner probably used this during dev to be able to just drop raw data type codes in for quick testing 'cause it felt more convenient than having to add the `.init()` boilerplate (I can understand the feeling! 😅). Of course, then he either just forgot about it later or looked at it and went "well why not just leave it in" - he was happy to let me take the role of overzealous code neat freak 😆

(I'm 100% in favor of ditching it regardless of all that, all it does is make it easier to misuse the API.)

Sources/PostgresNIO/Data/PostgresDataType.swift
```diff
@@ -227,3 +222,13 @@ public struct PostgresDataType: Codable, Equatable, ExpressibleByIntegerLiteral,
         return self.knownSQLName ?? "UNKNOWN \(self.rawValue)"
     }
 }
+
+// TODO: The Codable conformance does not make any sense. Let's remove this with next major break.
```
gwynne, line 226: Was there a point in early Swift 5 before `RawRepresentable` automatically conferred `Codable`? That would make sense of why the redundant conformance was left in. (I mean heck, I tend to write out `Equatable, Hashable` by habit and then grumble at the compiler warning complaining `Equatable` is redundant...)

Though, given that the conformance is automatically bestowed regardless, does the removal of the explicit declaration actually have to wait for a major version break?

(Hm, do structs get it automatically? Or just enums? Either way, the enum in the other PR could lose it without a major version break at least.)

gwynne (approved):
Pretty much copy and paste what I wrote for #212, I love cleanup work ❤️

gwynne:
Sigh, merge conflicts. You'd think Git would be smart enough to be able to tell #212 and this are directly related 😒

PR #215, opened 2022-02-15
Remove PSQLJSONEncoder

fabianfett:
### Motivation

While rewriting lot's of PostgresNIO internals in the last year, I created new types that very closely match existing types. This was a mistake. I should have tried everything to reuse existing types, and change them as needed.

### Changes

⚠️ This PR extends API... minor version bump required.

- Remove `PSQLJSONEncoder`
- Extend `PostgresJSONEncoder` with an encode method based on `ByteBuffer`
- Provide a default implementation for the new encode method based on `ByteBuffer` to not break API
- Replace all uses of `PSQLJSONEncoder` with `PostgresJSONEncoder`

### Result


gwynne (approved):
Same as #212 and #213, why are you wasting my time reading my approval instead of merging lovely code cleanups faster than I can thumbs-up 'em? 🙃

PR #216, opened 2022-02-15
Remove PSQLJSONDecoder

fabianfett:
### Motivation

While rewriting lot's of PostgresNIO internals in the last year, I created new types that very closely match existing types. This was a mistake. I should have tried everything to reuse existing types, and change them as needed.

### Changes

⚠️ This PR extends API... minor version bump required.

- Remove `PSQLJSONDecoder`
- Extend `PostgresJSONDecoder` with a decode method based on `ByteBuffer`
- Provide a default implementation for the new decode method based on `ByteBuffer` to not break API
- Replace all uses of `PSQLJSONDecoder ` with `PostgresJSONDecoder`

### Result


gwynne (approved):
What I said on #215? That, but with more Infinity Improbability Drive side effects going on to drive home the idea that these are good PRs 😂

PR #220, opened 2022-02-17
Add PostgresCell

fabianfett:
### Motivation

In #219 we want to restructure PostgresRow and introduce PostgresRandomAccessRow. To make it very explicit, it would be cool if `PostgresRow` was implementing `Collection` and `PostgresRandomAccessRow` was implementing `RandomAccessCollection`. However to do this we need an Element type. This is what `PostgresCell` shall be.

### Changes

- Add `PostgresCell`
- Remove `PSQLData`

### Result

We have an awesome Element type for PostgresRow.


Sources/PostgresNIO/New/PostgresCell.swift
```diff
@@ -0,0 +1,47 @@
...
+
+struct PostgresCell: Equatable {
+    var bytes: ByteBuffer?
+    var dataType: PostgresDataType
+    var format: PostgresFormat
+
+    var columnName: String
+    var columnIndex: Int
+
+    /// use this only for testing
```
gwynne, line 11: There are no other initializers and the type is not `Codable` - this is the only way to create one of these. I'm wondering if this is potentially an example of splitting up the PRs _too_ much- if there're missing pieces that you couldn't even add via extension (such as non-delegating initializers), the PR is not complete 😕.

Or the comment may just be outdated, but then what would I make a fuss about? 😉

gwynne (changes requested):
Other than the one concern, I don't see anything that needs changing... is there a more specific reason to keep this PR separate other than general desire to cut down on large PRs? (A worthy goal, to be sure!)

Sources/PostgresNIO/New/PostgresCell.swift
fabianfett: The comment was outdated. Removed. 🤦‍♂️

gwynne (approved):
Merge away!

PR #221, opened 2022-02-18
Extend PostgresCastingError

fabianfett:
I forgot to add some important metadata to the `PostgresCastingError`. For this reason here it is... Also added some API comments that will come in handy once we make this type public.


gwynne (approved):
🎉 ❤️

PR #222, opened 2022-02-18
Add PSQLRow multi decode

fabianfett:
Follow up to #221 and required to make #219 easier to review.

### Motivation

We want to efficiently parse Postgres types back into Swift types. The best way to do this is by iterating the ByteBuffer exactly once and calling decode for every row.

### Changes

- Add a bash script that does the code generation for PSQLRow decode methods for us.

### Result

It becomes very cheap to decode a complete PSQLRow.


gwynne (approved):
I'll let the bash script pass 😉

PR #223, opened 2022-02-19
Add PostgresQuery

fabianfett:
### Motivation

Right now we encode bindings right before we send them out, in `PSQLFrontendMessageEncoder`. This means that a JSONEncoder must be set per connection. This is not an ideal design. Further we need to wrap the values that we want to bind in an existential `[PSQLEncodable]`.

I propose changing this API and introducing an explicit `PostgresQuery` type, that is `ExpressibleByStringInterpolation`.

### Changes

- Add `PostgresQuery`
- Add `PostgresBindings`
- Remove `PostgresJSONEncoder` from `PSQLFrontendMessageEncoder`

### Result

- No more existentials in the encoding path.
- Cleaner, less cluttered code.


Sources/PostgresNIO/Data/PostgresDataType.swift
```diff
@@ -26,7 +26,7 @@ public typealias PostgresFormatCode = PostgresFormat
 
 /// The data type's raw object ID.
 /// Use `select * from pg_type where oid = <idhere>;` to lookup more information.
-public struct PostgresDataType: RawRepresentable, Equatable, CustomStringConvertible {
+public struct PostgresDataType: RawRepresentable, Hashable, CustomStringConvertible {
```
fabianfett, line 29: @gwynne We extend public API here!

gwynne: Fine by me 🙂

gwynne (approved):
I have no problems with this as it stands on its own, though it may need some tweaking to be ideal for SQLKit/FluentKit interop, mostly in terms of the serialization of positional parameters. That can be addressed later, though.

PR #225, opened 2022-02-21
Rename PSQLCodable to PostgresCodable

fabianfett:
### Motivation

Same as #224.

### Modifications

- Rename `PSQLCodable` to `PostgresCodable`
- Rename `PSQLEncodable` to `PostgresEncodable`

### Result

More consistent future API

PR #224, opened 2022-02-19
Rename PSQLDecodable to PostgresDecodable

fabianfett:
### Motivation

The PSQL prefix is uncool for new public types.

### Changes

- Rename `PSQLDecodable` to `PostgresDecodable`

PR #226, opened 2022-02-21
Rename PSQLRowSequence to PostgresRowSequence

fabianfett:
### Motivation

Same as #224 and #225.

### Changes

- Rename `PSQLRowSequence` to `PostgresRowSequence`


gwynne (approved):
<no review message was provided; only a disturbance, as if a thousand repetitions of a sequence of letters suddenly cried out in terror and were shoved into a pull request...>

(Don't look at me like that, I've been awake long enough to keep making silly jokes in PR reviews in the first place, it wouldn't be any fun if they were _good_ jokes 😅 Besides, they're good PRs; the work getting done is phenomenal ❤️.)

PR #229, opened 2022-02-25
Deprecate unused postgres messages

fabianfett:
We should start to deprecate unused public API, that we want to remove with next major break.


gwynne (approved):
Delightful ❤️

PR #230, opened 2022-02-25
Cleanup encoding and decoding

fabianfett:
### Motivation

Simple code is better. 

We have an escape hatch to implement the encoding of optionals on PostgresEncodable. However with the introduction of `PostgresBindings`, we can deal with this special case directly. Sadly this is not possible for Decoding. However we can improve the usability of the PostgresDecodable protocol, by prefixing the special values with an underscore.

### Changes

- Rename `decodeRaw` on `PostgresDecodable` to `_decodeRaw`
- Rename `DecodableType` on `PostgresDecodable` to `_DecodableType`
- Remove Optional conformance to `PostgresEncodable`

### Result

Less and easier code.


gwynne (approved):
You're going for a 100% approval record, aren't you, Fabian? 🙂

I can't wait to start seeing some before and after performance numbers on this code!

PR #232, opened 2022-02-25
Add PostgresRowSequence multi decode

fabianfett:


gwynne (approved):
I'll do a PR to rewrite your cute Bash script in Swift later. BTW, a SPM build plugin for codegen would increase your cool factor a lot... 😉 Meanwhile, merge away!

PR #233, opened 2022-02-26
Add an async query API (internal for now)

fabianfett:
### Motivation

We like async code

### Changes

- Add an async query API to PostgresConnection (internal for now)
- Add first two async integration tests

### Result

- Async is closer to land publicly


gwynne (approved):
Okay, now you're just outright teasing me. I demand perf benchmarks, including before/after you started working on the package altogether!

PR #235, opened 2022-02-26
[PostgresRowSequence] Make StateMachine private

fabianfett:
### Motivation

- Elements that are private are easier for the compiler to optimize

### Changes

- Make `PostgresRowSequence.StateMachine` private
- Remove unnecessary whitespace

### Result

Code that is easier to optimize for the compiler


gwynne (approved):
Not sure what I’d request changes for in this one… not entirely sure I’m actually approving anything in the first place! 😆 This repo is easily the most active one in Vapor right now….

PR #236, opened 2022-02-26
Rename PSQLRow multi decode to PostgresRow multi decode

fabianfett:


gwynne (approved):
_Hello darkness, my old friend / I've come to talk with you again_
_Because a vision softly creeping / left its seeds while I was sleeping..._

PR #237, opened 2022-02-28
Explicit TLS config

fabianfett:
### Motivation

Currently users provide a TLSConfiguration to the PSQLConnection.Configuration to signal they want to require TLS. This is very implicit and it will create a new `NIOSSLContext` for every new connection (VERY, VERY expensive operation).

### Changes

- Make TLS config explicit in `PSQLConnection.Configuration`. The initial provided values are `.disable`, `.prefer`, `.require`
- If users use `.prefer` or `.require` they must provide a NIOSSLContext

### Result

- Faster more explicit code

### DIFF
--- a/Sources/PostgresNIO/New/PSQLConnection.swift
+++ b/Sources/PostgresNIO/New/PSQLConnection.swift
@@ -23,6 +23,30 @@ final class PSQLConnection {
                 self.password = password
             }
         }
+
+        struct TLS {
+            enum Base {
+                case disable
+                case prefer(NIOSSLContext)
+                case require(NIOSSLContext)
+            }
+
+            var base: Base
+
+            private init(_ base: Base) {
+                self.base = base
+            }
+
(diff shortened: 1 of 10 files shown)


Sources/PostgresNIO/New/PSQLConnection.swift
```diff
@@ -23,6 +23,30 @@ final class PSQLConnection {
...
+            static var disable: Self = Self.init(.disable)
+
+            static func prefer(_ sslContext: NIOSSLContext) -> Self {
+                self.init(.prefer(sslContext))
+            }
+
+            static func require(_ sslContext: NIOSSLContext) -> Self {
+                self.init(.require(sslContext))
+            }
+        }
```
fabianfett, line 49: @gwynne This will become public API eventually.

gwynne (approved):
LGTM

PR #238, opened 2022-02-28
Rename PSQLBackendMessage to PostgresBackendMessage

fabianfett:
### Motivation

We want to drop the PSQL prefix.

### Changes

- Rename `PSQLBackendMessage` to `PostgresBackendMessage`


gwynne (approved):
👍

PR #239, opened 2022-02-28
Rename PSQLFrontendMessage to PostgresFrontendMessage

fabianfett:
### Motivation

We want to drop the PSQL prefix.

### Changes

Rename `PSQLFrontendMessage` to `PostgresFrontendMessage`


gwynne (approved):
👍

PR #240, opened 2022-03-02
Merge PSQLConnection into PostgresConnection

fabianfett:
### Motivation

We want to evolve our current public API instead of just replacing it.

### Changes

- Merge PSQLConnection into PostgresConnection

### Result

A cleaner public API


gwynne (approved):
Love it ❤️

PR #241, opened 2022-03-03
Cleanup PostgresDecodable

fabianfett:
### Motivation

Before we make the `PostgresDecodable` public, let's revisit it. In Swift prefer initializer over static named methods for type creation.

### Changes

- Replace the `static func decode` with an `init`
- Cleanup the default implementations. Put the impl. into their correct protocol conformance.

### DIFF
--- a/Sources/PostgresNIO/New/PostgresCodable.swift
+++ b/Sources/PostgresNIO/New/PostgresCodable.swift
@@ -24,7 +24,7 @@ protocol PostgresDecodable {
     /// String? should be PostgresDecodable, String?? should not be PostgresDecodable
     associatedtype _DecodableType: PostgresDecodable = Self
 
-    /// Decode an entity from the `byteBuffer` in postgres wire format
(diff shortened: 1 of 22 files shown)


Sources/PostgresNIO/New/PostgresCodable.swift
```diff
@@ -35,12 +35,12 @@ protocol PostgresDecodable {
     ///   - context: A `PSQLDecodingContext` providing context for decoding. This includes a `JSONDecoder`
     ///              to use when decoding json and metadata to create better errors.
     /// - Returns: A decoded object
-    static func decode<JSONDecoder: PostgresJSONDecoder>(
+    init<JSONDecoder: PostgresJSONDecoder>(
```
fabianfett, line 38: @gwynne This is the real meat of this PR. Everything else is cleanup.

gwynne (approved):
👍

PR #242, opened 2022-03-03
Add Sendable checking

fabianfett:
### Motivation

For Swift 5.6 we want to add Sendable checking

### Changes

- Add Sendable protocol requirement where needed.

### Result

Safer code for anyone.


gwynne (approved):
It's a start 😉

PR #244, opened 2022-03-06
Make new Postgres decoding public

fabianfett:
### Motivation

After a long time in the works, we finally want to make the new Postgres decoding public.

### Changes

- Add new `protocol PostgresArrayDecodable` that inherits from `PostgresDecodable`. A type that conforms to `PostgresArrayDecodable` can be decoded into an array of its own type from a Postgres array.
- Make PostgresCell decode method public
- Make PostgresRow decode methods public (we use `@_alwaysEmitIntoClient` here, to keep binary sizes small)
- Mark a number of things `@usableFromInlineable` to allow better perf

### Result

Faster code for users!


gwynne (approved):
I like it 😁

PR #245, opened 2022-03-06
New PostgresConnection connect API

fabianfett:
### Motivation

Our current connection creation API is very weird. Adopters first need to connect to a server and then manually call authenticate on the just created connection to auth and select the database.

### Changes

- Add new connect method that allows users to connect and auth with a single call
- Add new connect async/await API

### Result

Easier to use public interface

### Open questions:

- Parameter order in new connect method.
- Splitting up PostgresConnection.Configuration into three subtypes.


dnadoba (approved):
LGTM

gwynne (approved):
LGTM!

PR #247, opened 2022-03-08
Make PostgresRowSequence public

fabianfett:
### Motivation

For the new async interface we need the `PostgresRowSequence` to be public

### Changes

- Make `PostgresRowSequence` public
- Rename `PostgresRowSequence.Iterator` to `PostgresRowSequence.AsyncIterator` (Having an `Iterator` and an `AsyncIterator` in the same type is confusing)
- Make `PostgresRowSequence` decode methods public

### Result

Only need to make encoding public


gwynne (approved):
This is all adding up to something pretty awesome, isn't it? 👍

PR #250, opened 2022-03-09
Make DataRow Sendable

fabianfett:
### Motivation

For Swift 5.6, we want to support Sendable checks. `DataRow` generates a lot of warnings in CI today.

### Changes

- Make `DataRow` `Sendable`

### DIFF
--- a/Sources/PostgresNIO/New/Messages/DataRow.swift
+++ b/Sources/PostgresNIO/New/Messages/DataRow.swift
@@ -1,4 +1,8 @@
+#if swift(>=5.6)
+@preconcurrency import NIOCore
+#else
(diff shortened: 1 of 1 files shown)


Sources/PostgresNIO/New/Messages/DataRow.swift
```diff
@@ -116,3 +120,7 @@ extension DataRow {
         return self[byteIndex]
     }
 }
+
+#if swift(>=5.6)
```
gwynne, line 124: Why 5.6 rather than 5.5?

gwynne (approved):
Marching onward!

PR #252, opened 2022-03-09
Rename PostgresBackendMessageDecoder

fabianfett:
Same old story: Drop PSQL prefix, use Postgres prefix.


gwynne (approved):
Same old story, same old yawn+approve 🙂

PR #253, opened 2022-03-09
Add support for Network.framework

fabianfett:
Add support for Network.framework

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -1,5 +1,8 @@
 import NIOCore
 import NIOConcurrencyHelpers
+#if canImport(Network)
+import NIOTransportServices
+#endif
 import NIOSSL
 import Logging
 import NIOPosix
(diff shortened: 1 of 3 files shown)


Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -271,6 +275,23 @@ public final class PostgresConnection {
         }
     }
 
+    static func makeBootstrap(
+        on eventLoop: EventLoop,
+        configuration: PostgresConnection.InternalConfiguration
+    ) -> NIOClientTCPBootstrapProtocol {
+        #if canImport(Network)
```
gwynne, line 282: I would prefer if clients were able to override this decision.

gwynne (changes requested):

Sources/PostgresNIO/Connection/PostgresConnection.swift
gwynne: Withdrawn, I didn't realize it was using the validating initializers. Good to go!

gwynne (approved):

PR #258, opened 2022-03-12
Allow unsafe raw SQL injection in PostgresQuery

fabianfett:
### Motivation

Sometimes developers want to have the freedom to inject values into a PostgresQuery that are not bindable. Examples for this can be tableNames or columnNames.

### Changes

- Allow adopters to interpolate raw sql into the PostgresQuery

### Result

More use-cases covered. Make unsafe code explicit unsafe.

### Open question

What is the best unsafe marker word? Right now I have `unsafeSQLInjection`. I like this because it makes potential users google for SQL injection... This is rather long. Other previously discussed options are: `raw`, `unsafeRaw`.

### DIFF
--- a/Sources/PostgresNIO/New/PostgresQuery.swift
+++ b/Sources/PostgresNIO/New/PostgresQuery.swift
@@ -67,6 +67,11 @@ extension PostgresQuery {
             try self.binds.append(value, context: context)
             self.sql.append(contentsOf: "$\(self.binds.count)")
         }
+
+        @inlinable
+        public mutating func appendInterpolation(unescaped interpolated: String) {
+            self.sql.append(contentsOf: interpolated)
+        }
     }
 }
 
--- a/Tests/PostgresNIOTests/New/PostgresQueryTests.swift
+++ b/Tests/PostgresNIOTests/New/PostgresQueryTests.swift
@@ -3,16 +3,18 @@ import XCTest
 
 final class PostgresQueryTests: XCTestCase {
 
-    func testStringInterpolationWithOptional() throws {
+    func testStringInterpolationWithOptional() {
         let string = "Hello World"
         let null: UUID? = nil
         let uuid: UUID? = UUID()
 
-        let query: PostgresQuery = try """
+        var query: PostgresQuery?
+        XCTAssertNoThrow(query = try """
             INSERT INTO foo (id, title, something) SET (\(uuid), \(string), \(null));
             """
+        )
 
-        XCTAssertEqual(query.sql, "INSERT INTO foo (id, title, something) SET ($1, $2, $3);")
+        XCTAssertEqual(query?.sql, "INSERT INTO foo (id, title, something) SET ($1, $2, $3);")
 
         var expected = ByteBuffer()
         expected.writeInteger(Int32(16))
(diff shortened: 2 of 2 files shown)


Sources/PostgresNIO/New/PostgresQuery.swift
```diff
@@ -67,6 +67,11 @@ extension PostgresQuery {
             try self.binds.append(value, context: context)
             self.sql.append(contentsOf: "$\(self.binds.count)")
         }
+
+        @inlinable
+        public mutating func appendInterpolation(unsafeSQLInjection interpolated: String) {
```
gwynne, line 72: I think the word `Injection` is overkill in this case - `unsafe` is at least theoretically supposed to be a sufficiently big signal flare in _any_ Swift code to cover it. I'd tend towards making the label `unsafeSQL` and then clearly defining SQL injection and its dangers (in short form, not one of my rants!! 😆) as a documentation comment.

Tests/PostgresNIOTests/New/PostgresQueryTests.swift
```diff
@@ -79,4 +76,19 @@ final class PostgresQueryTests: XCTestCase {
 
         XCTAssertEqual(query.binds.bytes, expected)
     }
+
+    func testUnsafeSQLInjection() {
```
gwynne, line 80: This is just me being silly, but you know, if you're gonna test SQL injection, why not at least have the test do a classical injection? Say, `#"SELECT foo FROM dual LIMIT 0; INSERT INTO users (username, password, isSuperuserOverAllTheInternet) VALUES ("Good ol' Bobby Tables", "easy password", 1); -- PWNED"#` 😇

gwynne (changes requested):

Sources/PostgresNIO/New/PostgresQuery.swift
fabianfett: However unsafeSQL doesn't give users such clear google results. SQL injection results are very clear...

0xTim: Yeah I agree `unsafeSQLInjection` is confusing. We can add it as a comment if we're concerned about searching. I like `\(unsafeRaw: "")` or `\(raw: "")` - lines up with the rest of the cases across Vapor where we allow unsafe injection of variables and still makes it clear

adam-fowler: I think the full name is too wordy. Surely we can detail the issues with sql injection in the comments. I would go further than @gwynne i don't think SQL is required. It should be obvious it is SQL as it is inside a SQL query. So `\(unsafe: ...)` or `\(unsafeRaw...)` should be enough.

Jerry-Carter: I’ll put in a vote for ‘unsafeRaw’ in that it suggests direct SQL phrases without escaping or other processing for safety.

fabianfett: I talked to @weissi about this yesterday, and we came up with `unescaped`, which I like the most so far. @0xTim, @gwynne, @adam-fowler, @Jerry-Carter What do you think about this?

0xTim: Yeah I'm happy with that

adam-fowler: Sounds good

0xTim (approved):
LGTM

Sources/PostgresNIO/New/PostgresQuery.swift
gwynne: I want to argue for `unsafeUnescaped`, but not enough to die on the hill; I can live with this 🙂

gwynne (approved):

PR #261, opened 2022-03-13
Continue to make forward progress when cancelled

fabianfett:
### Motivation

With the async/await interface we introduced the option to ignore a rowStream (by dropping the `AsyncSequence` or `AsyncIterator`). In those cases we must fail the stream and drain the connection ASAP. Otherwise we will lock up the connection.

### Changes

- Drain connection, if the row stream was cancelled

### Result

Connection is reusable, even if previous queries were cancelled.

### Note:

Needs more unit tests

### DIFF
--- a/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
+++ b/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
@@ -842,7 +842,15 @@ struct ConnectionStateMachine {
     // MARK: Consumer
     
     mutating func cancelQueryStream() -> ConnectionAction {
-        preconditionFailure("Unimplemented")
+        guard case .extendedQuery(var queryState, let connectionContext) = self.state, !queryState.isComplete else {
+            preconditionFailure("Tried to cancel stream, without active query")
+        }
+
+        return self.avoidingStateMachineCoW { machine -> ConnectionAction in
+            let action = queryState.cancel()
+            machine.state = .extendedQuery(queryState, connectionContext)
+            return machine.modify(with: action)
+        }
     }
     
     mutating func requestQueryRows() -> ConnectionAction {
@@ -1074,6 +1082,8 @@ extension ConnectionStateMachine {
             return true
         case .failedToAddSSLHandler:
             return true
+        case .queryCancelled:
+            return false
         case .server(let message):
             guard let sqlState = message.fields[.sqlState] else {
                 // any error message that doesn't have a sql state field, is unexpected by default.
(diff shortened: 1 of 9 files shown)


Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -842,7 +842,15 @@ struct ConnectionStateMachine {
     // MARK: Consumer
     
     mutating func cancelQueryStream() -> ConnectionAction {
-        preconditionFailure("Unimplemented")
+        guard case .extendedQuery(var queryState, let connectionContext) = self.state, !queryState.isComplete else {
+            preconditionFailure("Tried to cancel stream, without active query")
```
gwynne, line 846:
```suggestion
            preconditionFailure("Tried to cancel stream without active query")
```
Yes, I'm _this_ pedantic about grammar nits 🙂

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -15,6 +15,8 @@ struct ExtendedQueryStateMachine {
         /// used after receiving a `bindComplete` message
         case bindCompleteReceived(ExtendedQueryContext)
         case streaming([RowDescription.Column], RowStreamStateMachine)
+        /// A state that is used, if the current query was cancelled, and we want to drain rows from the connection ASAP
```
gwynne, line 18:
```suggestion
        /// Indicates that the current query was cancelled and we want to drain rows from the connection ASAP
```

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -71,6 +75,44 @@ struct ExtendedQueryStateMachine {
...
+             .parseCompleteReceived(let queryContext),
+             .parameterDescriptionReceived(let queryContext),
+             .rowDescriptionReceived(let queryContext, _),
+             .noDataMessageReceived(let queryContext),
+             .bindCompleteReceived(let queryContext):
+            guard !self.isCancelled else {
+                return .wait
+            }
+
+            self.isCancelled = true
```
gwynne, line 94: If you're `guard`ing that `self.isCancelled` is true, you don't need to then set it `true` immediately afterward.

gwynne (changes requested):
Two utterly pointless grammar nits just to annoy you 🙂, plus one minor functional nit, marking as changes requested in case there was an actual logic error there.

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
fabianfett: Not sure I or you misread the code here: I think, I'm guarding that `isCancelled` is `false`, to then set it to `true`.

gwynne: Whoops, that was my misread! And even more embarrassing 'cause I did the same thing a couple lines down and then caught myself on that one 😅

gwynne (approved):
My bad on the misread, approved!

PR #264, opened 2022-03-14
Start work on async listen

fabianfett:
### Motivation

We want to translate all our APIs to async.

### Changes

- Add `listen(_: String) -> PostgresNotificationSequence` function

### Result

Async listen is possible.


pepicrft:
What's the state of this @fabianfett?

fabianfett:
Hi @pepicrft, thanks for reaching out. I haven't touched this for quite a while and I'd be happy for someone else to drive this forward. Are you interested? 

If that's the case I will write down my thoughts on this, to make getting started easier.

pepicrft:
> Hi @pepicrft, thanks for reaching out. I haven't touched this for quite a while and I'd be happy for someone else to drive this forward. Are you interested?

Very interested indeed. Would you brain-dumping everything in this issue?

fabianfett:
Hi @pepicrft, 

I sat down last night and tried to move this forward so that you can have an idea of where I want this to go. Please be well aware that the code is totally undertested and that likely only the happy path works. I'd really appreciate help here. However if you say this is already "too" done, and you are not interested in helping out anymore that is understandable. In that case you have a very bar bones implementation that you might be able to use.

But first let me discuss design goals:

1. We want to support listening to the same channel multiple times
2. We want to cancel listening on the connection (by telling the server `UNLISTEN`) when all listeners have cancelled
3. Users shall not be required to send `LISTEN` and `UNLISTEN` commands themselves to start the listening
4. If a connection dies all listeners shall be closed by throwing an error

Next let me explain, what types I've added and how all of this works together:

1. In `PostgresConnection` we now have an `internalListenID` atomic. We use this to generate unique ids for NotificationListeners. This is important to support correct cancellation.
5. The `PostgresConnection` now communicates with the `PostgresChannelHandler` using a `HandlerTask`. The `HandlerTask` is a superset of the existing `PSQLTask` and adds `startListening` and `cancelListening`
6. All state handling code that is required for Listening has moved from `PostgresConnection` into the `PostgresChannelHandler`
7. Inside the `PostgresChannelHandler` we use a `ListenStateMachine` to handle all state handling of listening to a given channel
8. The `ListenStateMachine` has a sub statemachine for each channel called `ChannelState`
9. The `PostgresChannelHandler` translates actions from the `ListenStateMachine` into tasks that need to be performed on the connection
    - `makeStartListeningQuery`
    - `startListenCompleted`
    - `makeStopListeningQuery`
    - `forwardNotificationToListeners`
10. The `NotificationListener` is a stateful reference type that represents the different states a single Listen command can be in. It is synchronized by using the connections `EventLoop` as synchronization primitive.
    - It supports the new API `.streamInitialized`, `.streamListening`
    - As well as the old existing interface `.closure`
11. We return an `PostgresNotificationSequence` which is a wrapper around an `AsyncThrowingStream`. We use the wrapper approach here to ensure that we can eventually back `PostgresNotificationSequence` with another `AsyncSequence`, without needing to change public API.

Feel free to use the commit in this PR and add further commits (by opening a new PR), that finalize the implementation and add tests. For an example on how to test a state machine, check out the `ConnectionStateMachineTests` in `PostgresNIOTests`.

How does all of this sound to you @pepicrft?

Small note: I'll be on vacation starting tomorrow until July 3. For this reason I might be slow to respond.

pepicrft:
> How does all of this sound to you @pepicrft?

Sounds perfect and thanks a lot for the thorough update on the state of things. I won't be able to dedicate myself full-time to it, but I hope to make some progress before you are back from vacation 😊.
I'll open a separate PR that we can sync on when you are back.

Tests/PostgresNIOTests/New/PostgresChannelHandlerTests.swift
```diff
@@ -6,17 +6,24 @@ import NIOEmbedded
 @testable import PostgresNIO
 
 class PostgresChannelHandlerTests: XCTestCase {
-    
+
+    var eventLoop: EmbeddedEventLoop!
+
+    override func setUp() {
+        self.eventLoop = EmbeddedEventLoop()
+        super.setUp()
```
gwynne, line 14:
```suggestion
        super.setUp()
        self.eventLoop = EmbeddedEventLoop()
```

Sources/PostgresNIO/New/NotificationListener.swift
```diff
@@ -0,0 +1,156 @@
...
+
+        case .done:
+            break // ignore
+        }
+    }
+}
+
+
+#if swift(<5.9)
+// Async stream API backfil
```
gwynne, line 144:
```suggestion
// Async stream API backfill
```

Sources/PostgresNIO/New/NotificationListener.swift
```diff
@@ -0,0 +1,156 @@
...
+        self.eventLoop = eventLoop
+        self.state = .streamInitialized(checkedContinuation)
+    }
+
+    init(
+        channel: String,
+        id: Int,
+        eventLoop: EventLoop,
+        context: PostgresListenContext,
+        closure: @escaping (PostgresListenContext, PostgresMessage.NotificationResponse) -> Void
```
gwynne, line 36: Should this be `@Sendable`?

gwynne (approved):
A few minor nits, otherwise lgtm

fabianfett:
API break checker reports a false positive:
```
1 breaking change detected in PostgresNIO:
  💔 API breakage: func PostgresConnection.addListener(channel:handler:) is now with @preconcurrency
```

PR #265, opened 2022-03-16
Add EventLoop API that uses PostgresQuery

fabianfett:
### Motivation

I talked to @adam-fowler and he mentioned that it would be great if we would also offer a `PostgresQuery` based API for usage with EL.

### Changes

- Add better EL APIs based on `PostgresQuery`

### Result

Nicer API even for users that stay on the EL.


gwynne (approved):
Agree! 🙂

PR #266, opened 2022-03-16
Remove connection state machine log

fabianfett:
### Motivation

In trace mode we log every new state machine state. This is not necessary, since our state machine can be better described through all the inputs it receives. We log those in trace mode as well. When running a query with 2m rows, I can see in Instruments that we spend about 100ms in the trace log statement that I want to remove here – even if we don't use trace logging.

### Changes

- Remove unnecessary state log

### Motivation

A bit faster code, less verbose output


gwynne (approved):
I would've wanted this removed even if there wasn't a perf drop from it - Fluent's trace logging is way too noisy historically as it is, and this certainly fits that category! 🙂

PR #267, opened 2022-03-16
Use Int(exactly:) instead of restricting to 64 bit platforms

### LINKED ISSUE #199: should use the `Int(exactly:)` constructor instead of restricting itself to 64 bit platforms
https://github.com/vapor/postgres-nio/blob/d97f26858003d87a1d2b52c0061b243035620996/Sources/PostgresNIO/Data/PostgresData%2BInt.swift#L47

I think

```
buffer.readInteger(as: UInt32.self).flatMap { Int(exactly($0) }
```

should be much better

```
  5> Int32(exactly: UInt32.max) 
$R4: Int32? = nil
  6> Int32(exactly: 1)
$R5: Int32? = 1
```

fabianfett:
Fixes #199. cc @weissi

### DIFF
--- a/Sources/PostgresNIO/Data/PostgresData+Int.swift
+++ b/Sources/PostgresNIO/Data/PostgresData+Int.swift
@@ -1,7 +1,6 @@
 extension PostgresData {
     public init(int value: Int) {
-        assert(Int.bitWidth == 64)
-        self.init(type: .int8, value: .init(integer: value))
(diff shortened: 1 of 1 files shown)


weissi (approved):
lgtm, thanks

gwynne (approved):
Win. Made a very similar change in MySQLNIO recently; Tanner was really way too willing to just crash everything at the slightest provocation 🤦‍♀️

PR #275, opened 2022-03-25
bump target version to 12 and swift tools version to 5.5 so it compli…

wibed:
…es with the concurrency api

bump target version to 12 and swift tools version to 5.5 so it complies with the concurrency api


0xTim:
This is not needed. Any async/await APIs are wrapped in availability checks

Issue #273, opened 2022-03-24
Expose connectTimeout as a configuration option


trasch:
**Is your feature request related to a problem? Please describe.**
Currently, the connection timeout (default is 10 seconds) for new Postgres connections is hidden away in `NIOTSConnectionBootstrap` or `ClientBootstrap`. It would be nice to be able to change this via a configuration option.

**Describe the solution you'd like**
A configuration option `connectTimeout`.

**Additional context**
Depending on use case and environment, the default 10 seconds might be too short, or (in my case) way too long.

PR #276, opened 2022-03-25
#273: Expose connectTimeout as a configuration option

trasch:
This exposes the `connectTimeout` property in the configuration.

```
let config = PostgresConnection.Configuration(
   connection: .init(
     host: "localhost",
     port: 5432,
     connectTimeout: .seconds(5)
   ),
   authentication: .init(
     username: "my_username",
     database: "my_database",
     password: "my_password"
   ),
   tls: .disable
)
```


Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -78,9 +78,15 @@ public final class PostgresConnection {
...
-            public init(host: String, port: Int = 5432) {
+            /// Specifies a timeout to apply to a connection attempt.
+            ///
+            /// - Default: 10 seconds
+            public var connectTimeout: TimeAmount
+
+            public init(host: String, port: Int = 5432, connectTimeout: TimeAmount = .seconds(10)) {
                 self.host = host
                 self.port = port
+                self.connectTimeout = connectTimeout
```
fabianfett, line 89:
```suggestion
            public init(host: String, port: Int = 5432) {
                self.host = host
                self.port = port
                self.connectTimeout = .seconds(10)
```
We sadly can not add another parameter to the initializer and not be source breaking. To work around this all properties are `vars` by default. This way, if users want to override the value, they can just use the property after the init.

fabianfett (commented):
@trasch sorry for tkaing so long to come back to you!

There is one direct code comment, and would you be open to add a test case for it?

Sources/PostgresNIO/Connection/PostgresConnection.swift
trasch: No problem, but I have to admit that I don't understand why this is source breaking... 🙂 Shouldn't a new parameter with a default value at the end of the parameter list compile everywhere without problems?

fabianfett: Yes, this API breaking change is very hard to see, but it occurs in this case:

Existing API:
```swift
struct Foo {
    init(bar: String) {

    }

    static func test() {
        let make = Foo.init(bar:)
        let foo = make("hello")
        // ...
    }
}
```

Adding a defaulted value breaks this code however:
```swift
struct Foo {
    init(bar: String, test: String = "foo") {

    }

    static func test() {
        let make = Foo.init(bar:) // <--- ❌ compiler error
        let foo = make("hello")
        // ...
    }
}
```

trasch:
@fabianfett 

> would you be open to add a test case for it

I would, and maybe you can give me a hint if there is already some code somewhere that opens a blocking port with which I could test this?

fabianfett:
@trasch after browsing around a bit, this seems to be quite hard to test... Sorry for making you look into it. Basically we would test nio functionality. If the CI passes, we should be good to go!

trasch:
@fabianfett 

> @trasch after browsing around a bit, this seems to be quite hard to test... Sorry for making you look into it. Basically we would test nio functionality. If the CI passes, we should be good to go!

Indeed, testing this would basically be replicating `NIOTSEndToEndTests.testBasicConnectionTimeout()` from the `NIOTransportServices` library - which I found a minute after asking you for help... :-D

Thanks for reviewing this!

fabianfett (approved):
Awesome! Thanks for the contribution @trasch!

Issue #279, opened 2022-04-25
Make `PostgresRowSequence.collect` public


fabianfett:
1. Checkout the code base:
2. Find the following code:
    ```swift
    extension PostgresRowSequence {
        func collect() async throws -> [PostgresRow] {
            var result = [PostgresRow]()
            for try await row in self {
                result.append(row)
            }
            return result
        }
    }
    ```
3. Make the function `public`
4. [Create a Pull Request](https://www.freecodecamp.org/news/how-to-make-your-first-pull-request-on-github-3/)

If you need any help, please let us know!

fabianfett:
Marked this issue as a assigned to me to move it to In Progress. I can not tag @nicholas-otto.

PR #281, opened 2022-04-26
Update PostgresRowSequence.swift

nicholas-otto:
As mentioned in issue #279 by @fabianfett
As requested in recent issue attached above, the function should be made public. That is the only change made in this pull request and should solve the issue!


fabianfett (approved):
Great. Thanks for tackling this. Since this is a public API change let’s also ask @gwynne for an opinion.

gwynne (approved):
Seems fine to me! Unless there was some pressing reason it wasn't `public` before which still applies at the present time, I say go for it and thank you! 🙂

Issue #278, opened 2022-04-25
Rename `PostgresCastingError` and make public


fabianfett:
1. Checkout the code base
2. Rename `PostgresCastingError` to `PostgresDecodingError`
3. Make `PostgresDecodingError` public
4. Make `PostgresDecodingError.Code` public
5. Make the static properties on `PostgresDecodingError.Code` public
6. Make the properties on `PostgresDecodingError` public
7. Write an extension to `PostgresDecodingError` that conforms to `CustomStringConvertible`, move the description property there.
8. Write a test that checks, if the description is always displayed as "Database error".
9. [Create a Pull Request](https://www.freecodecamp.org/news/how-to-make-your-first-pull-request-on-github-3/)

If you need any help, please let us know!

fabianfett:
Marked this issue as a assigned to me to move it to In Progress. I can not tag @nicholas-otto.

PR #282, opened 2022-04-26
Patch 2

nicholas-otto:
Proposed changes in issue #278 by @fabianfett

The changes included are renaming PostgresCastingError to PostgresDecodingError and making both it and PostgresDecodingError.code public. This also applies to the corresponding properties associated with them. Finally the description property which prevents security vulnerabilities to be linked is moved into an extension that conforms to CustomStringConvertible, which should be done as it is standard practice in swift. Thank you so much for the first issues work as this was my first time contributing (along with the pull request #281 )to an open source project and made me much more comfortable and familiar with the idea and excites me about the future possibilities!

### DIFF
--- a/Sources/PostgresNIO/New/PSQLError.swift
+++ b/Sources/PostgresNIO/New/PSQLError.swift
@@ -11,6 +11,7 @@ struct PSQLError: Error {
         case unsupportedAuthMechanism(PSQLAuthScheme)
         case authMechanismRequiresPassword
         case saslError(underlyingError: Error)
+        case invalidCommandTag(String)
 
         case queryCancelled
         case tooManyParameters
@@ -58,6 +59,10 @@ struct PSQLError: Error {
     
     static func sasl(underlying: Error) -> PSQLError {
         Self.init(.saslError(underlyingError: underlying))
+    }
+
+    static func invalidCommandTag(_ value: String) -> PSQLError {
+        Self.init(.invalidCommandTag(value))
     }
 
     static var queryCancelled: PSQLError {
--- a/Sources/PostgresNIO/New/PSQLError.swift
+++ b/Sources/PostgresNIO/New/PSQLError.swift
@@ -20,7 +20,7 @@ struct PSQLError: Error {
         case connectionError(underlying: Error)
         case uncleanShutdown
         
-        case casting(PostgresCastingError)
+        case casting(PostgresDecodingError)
     }
     
     internal var base: Base
@@ -92,9 +92,9 @@ struct PSQLError: Error {
 
 /// An error that may happen when a ``PostgresRow`` or ``PostgresCell`` is decoded to native Swift types.
 @usableFromInline
-struct PostgresCastingError: Error, Equatable {
+public struct PostgresDecodingError: Error, Equatable {
     @usableFromInline
-    struct Code: Hashable, Error {
+    public struct Code: Hashable, Error {
         enum Base {
             case missingData
             case typeMismatch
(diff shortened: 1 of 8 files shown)


Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -92,9 +92,9 @@ struct PSQLError: Error {
 
 /// An error that may happen when a ``PostgresRow`` or ``PostgresCell`` is decoded to native Swift types.
 @usableFromInline
```
fabianfett, line 94: `@usableFromInline` is not needed anymore. Please remove it.

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -92,9 +92,9 @@ struct PSQLError: Error {
...
-struct PostgresCastingError: Error, Equatable {
+public struct PostgresDecodingError: Error, Equatable {
     @usableFromInline
```
fabianfett, line 96: `@usableFromInline` is not needed anymore. Please remove it.

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -108,33 +108,33 @@ struct PostgresCastingError: Error, Equatable {
         }
 
         @usableFromInline
-        static let missingData = Self.init(.missingData)
+        public static let missingData = Self.init(.missingData)
         @usableFromInline
```
fabianfett, line 112: `@usableFromInline` is not needed anymore. Please remove it.

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -108,33 +108,33 @@ struct PostgresCastingError: Error, Equatable {
...
-        static let typeMismatch = Self.init(.typeMismatch)
+        public static let typeMismatch = Self.init(.typeMismatch)
         @usableFromInline
```
fabianfett, line 114: `@usableFromInline` is not needed anymore. Please remove it.

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -108,33 +108,33 @@ struct PostgresCastingError: Error, Equatable {
...
+    public let postgresData: ByteBuffer?
 
     /// The file the casting/decoding was attempted in
-    let file: String
+    public let file: String
     /// The line the casting/decoding was attempted in
-    let line: Int
+    public let line: Int
     
     var description: String {
```
fabianfett, line 139: You have moved this as intended into its own extension. I think we should remove it here.

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -180,6 +180,15 @@ struct PostgresCastingError: Error, Equatable {
     }
 }
 
+extension PostgresDecodingError: CustomStringConvertible {
+    var description: String {
```
fabianfett, line 184: This property must now be marked `public`.

We even do have a test case for it. However the test case has one big bug, that we should fix now (in `PostgresErrorTests.swift`):

```swift
    func testPostgresCastingErrorDescription() {
        let error = PostgresCastingError(
            code: .typeMismatch,
            columnName: "column",
            columnIndex: 0,
            targetType: String.self,
            postgresType: .text,
            postgresFormat: .binary,
            postgresData: ByteBuffer(string: "hello world"),
            file: "foo.swift",
            line: 123
        )

        XCTAssertNotEqual("\(error)", "Database error") // <-- Remove the Not 🤦‍♂️
    }
```

FYI: @gwynne I guess bug of a century.

fabianfett (commented):
This is a great start! And I'm excited that I can help you contribute to open source. If you have any further questions please reach out. There are a couple of small issues here, that we need to address. Most of them are my mistake, I didn't anticipate that they would be showing up.

Apparently a large number of `PostgresCastingError` is still left in the code:

I guess we will find most of them in: `dev/generate-postgresrow-multi-decode.sh`:
1. Replace `PostgresCastingError` in line 68 and 69 with `PostgresDecodingError`
2. Navigate to your `PostgresNIO` folder in terminal and run `dev/generate-postgresrow-multi-decode.sh`. This will recreate a lot of code generated code, that currently uses `PostgresCastingError`

I also have a few small other comments inline:

fabianfett:
@nicholas-otto Why did you close this one? And what is #285?

PR #286, opened 2022-04-26
Updates PostgresCastingError to PostgresDecodingError

nicholas-otto:
This updates from PostgresCastingError to PostgresDecodingError and changes associated files. This includes the update to the PostgresErrorTests.swift file in its testPostgresCastingErrorDescription() and the resolution of @usableFromInline tags. I believe it addresses all the issues from the previous pull request Patch 2 #282 and is in response to issue #278.


Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -68,33 +68,31 @@ struct PSQLError: Error {
...
         Self.init(.uncleanShutdown)
     }
 }
 
 /// An error that may happen when a ``PostgresRow`` or ``PostgresCell`` is decoded to native Swift types.
-@usableFromInline
-struct PostgresCastingError: Error, Equatable {
-    @usableFromInline
-    struct Code: Hashable, Error {
+struct PostgresDecodingError: Error, Equatable {
```
fabianfett, line 94: we are missing a `public` here.

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -167,7 +155,7 @@ struct PostgresCastingError: Error, Equatable {
     }
 
     @usableFromInline
-    static func ==(lhs: PostgresCastingError, rhs: PostgresCastingError) -> Bool {
+    static func ==(lhs: PostgresDecodingError, rhs: PostgresDecodingError) -> Bool {
```
fabianfett, line 158: we are missing a `public` here.

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -180,6 +168,14 @@ struct PostgresCastingError: Error, Equatable {
     }
 }
 
+extension PostgresDecodingError: CustomStringConvertible {
+  var description: String {
+    // This may seem very odd... But we are afraid that users might accidentally send the
+    // unfiltered errors out to end-users. This may leak security relevant information. For this
+    // reason we overwrite the error description by default to this generic "Database error"
+    "Database error"
+  }
```
fabianfett, line 177: We are missing a `public` here. Also this is indented with only 2 spaces.

dev/generate-postgresrow-multi-decode.sh
```diff
@@ -65,8 +65,8 @@ function gen() {
         echo -n ", r$(($n))"
     done
     echo ")"
-    echo "        } catch let code as PostgresCastingError.Code {"
-    echo "            throw PostgresCastingError("
+    echo "        } catch let code as PostgresDecodingError.Code {"
+    echo "            throw PostgresDecodingError("
```
fabianfett, line 69: This fix looks correct!

Sources/PostgresNIO/New/PostgresRow-multi-decode.swift
```diff
@@ -3,7 +3,11 @@
 extension PostgresRow {
     @inlinable
     @_alwaysEmitIntoClient
-    public func decode<T0: PostgresDecodable, JSONDecoder: PostgresJSONDecoder>(_: (T0).Type, context: PostgresDecodingContext<JSONDecoder>, file: String = #file, line: Int = #line) throws -> (T0) {
+-n     public func decode<T0: PostgresDecodable
+-n , JSONDecoder: PostgresJSONDecoder>(_: (T0
+-n ).Type, context: PostgresDecodingContext<JSONDecoder>, file: String = #file, line: Int = #line) throws
+-n  -> (T0
+) {
```
fabianfett, line 10: Why does echo suddenly print the `-n`. @gwynne do you have an idea why this might be happening?

fabianfett (commented):
We are getting closer 🎉. We are missing three publics, at one point the indentations is wrong, and for whatever the reason the script printed a lot of `-n`.

nicholas-otto:
I believe the pull request has been updated with the previous changes including a fix to the -n issue caused by dev/generate-postgresrow-multi-decode.sh

fabianfett:
Awesome! this looks much better already! However CI is still failing. Looks like a find & replace for the remaining usages of: `PostgresCastingError` would be the right approach now!

fabianfett:
@nicholas-otto We are nearly there! Looks like you ran into a small typo: `PostgresDecoingError` instead of `PostgresDecodingError`. If the project builds for you locally you should be good. You can call `swift build --build-tests` or just build in Xcode.

fabianfett (approved):
Great change! Thank you!

PR #296, opened 2022-06-02
Make backend key data optional

rausnitz:
This allows the `ConnectionStateMachine` to be in state `readyForQuery` even if the `BackendKeyData` has not been received.

The use case that prompted this change is integration with Amazon RDS Proxy. When connecting to Postgres via RDS Proxy, any query fails with error `PSQLError(base: PostgresNIO.PSQLError.Base.unexpectedBackendMessage(.readyForQuery(.idle)))`.

According to the RDS Proxy [docs](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/rds-proxy.html#rds-proxy.limits), one limitation of RDS Proxy is "For PostgreSQL, RDS Proxy doesn't currently support canceling a query from a client." The Postgres [docs](https://www.postgresql.org/docs/current/protocol-flow.html) explain BackendKeyData's purpose as: "This message provides secret-key data that the frontend must save if it wants to be able to issue cancel requests later. The frontend should not respond to this message, but should continue listening for a ReadyForQuery message."

So, it seems that RDS Proxy for Postgres has no use for BackendKeyData and does not provide it.

Allowing `nil` for `BackendKeyData` in this function resolves the `PSQLError` error we were seeing.

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -387,7 +387,6 @@ extension PostgresConnection {
         to socketAddress: SocketAddress,
         tlsConfiguration: TLSConfiguration? = nil,
         serverHostname: String? = nil,
-        requireBackendKeyData: Bool = true,
         logger: Logger = .init(label: "codes.vapor.postgres"),
         on eventLoop: EventLoop
     ) -> EventLoopFuture<PostgresConnection> {
@@ -406,7 +405,7 @@ extension PostgresConnection {
                 connection: .resolved(address: socketAddress, serverName: serverHostname),
                 authentication: nil,
                 tls: tls,
-                requireBackendKeyData: requireBackendKeyData
+                requireBackendKeyData: true
             )
 
             return PostgresConnection.connect(
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -77,13 +77,23 @@ public final class PostgresConnection {
             ///
             /// - Default: 5432
             public var port: Int
-
-            public init(host: String, port: Int = 5432) {
+            
+            /// Require connection to provide `BackendKeyData`
+            ///
+            /// - Default: true
+            public var requireBackendKeyData: Bool
+            
+            public init(
+                host: String,
+                port: Int = 5432,
+                requireBackendKeyData: Bool = true
+            ) {
                 self.host = host
                 self.port = port
+                self.requireBackendKeyData = requireBackendKeyData
             }
         }
-
+        
         public var connection: Connection
 
         /// The authentication properties to send to the Postgres server during startup auth handshake
@@ -377,6 +387,7 @@ extension PostgresConnection {
         to socketAddress: SocketAddress,
         tlsConfiguration: TLSConfiguration? = nil,
         serverHostname: String? = nil,
+        requireBackendKeyData: Bool = true,
         logger: Logger = .init(label: "codes.vapor.postgres"),
         on eventLoop: EventLoop
     ) -> EventLoopFuture<PostgresConnection> {
@@ -394,7 +405,8 @@ extension PostgresConnection {
             let configuration = PostgresConnection.InternalConfiguration(
                 connection: .resolved(address: socketAddress, serverName: serverHostname),
                 authentication: nil,
-                tls: tls
+                tls: tls,
+                requireBackendKeyData: requireBackendKeyData
             )
 
             return PostgresConnection.connect(
@@ -756,6 +768,8 @@ extension PostgresConnection {
         var authentication: Configuration.Authentication?
 
         var tls: Configuration.TLS
+        
+        let requireBackendKeyData: Bool
     }
 }
 
@@ -764,6 +778,7 @@ extension PostgresConnection.InternalConfiguration {
         self.authentication = config.authentication
         self.connection = .unresolved(host: config.connection.host, port: config.connection.port)
         self.tls = config.tls
+        self.requireBackendKeyData = config.connection.requireBackendKeyData
     }
 }
 
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -89,7 +89,7 @@ public final class PostgresConnection {
                 self.port = port
             }
         }
-        
+
         public var connection: Connection
 
         /// The authentication properties to send to the Postgres server during startup auth handshake
@@ -764,7 +764,7 @@ extension PostgresConnection {
 
         var tls: Configuration.TLS
         
-        let requireBackendKeyData: Bool
+        var requireBackendKeyData: Bool
     }
 }
 
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -78,19 +78,15 @@ public final class PostgresConnection {
             /// - Default: 5432
             public var port: Int
             
-            /// Require connection to provide `BackendKeyData`
+            /// Require connection to provide `BackendKeyData`.
+            /// For use with Amazon RDS Proxy, this must be set to false.
             ///
             /// - Default: true
-            public var requireBackendKeyData: Bool
+            public var requireBackendKeyData: Bool = true
             
-            public init(
-                host: String,
-                port: Int = 5432,
-                requireBackendKeyData: Bool = true
-            ) {
+            public init(host: String, port: Int = 5432) {
                 self.host = host
                 self.port = port
-                self.requireBackendKeyData = requireBackendKeyData
             }
         }
         
--- a/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
+++ b/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
@@ -5,8 +5,8 @@ struct ConnectionStateMachine {
     typealias TransactionState = PostgresBackendMessage.TransactionState
     
     struct ConnectionContext {
-        let processID: Int32
-        let secretKey: Int32
+        let processID: Int32?
+        let secretKey: Int32?
         
         var parameters: [String: String]
         var transactionState: TransactionState
@@ -543,14 +543,9 @@ struct ConnectionStateMachine {
     mutating func readyForQueryReceived(_ transactionState: PostgresBackendMessage.TransactionState) -> ConnectionAction {
         switch self.state {
         case .authenticated(let backendKeyData, let parameters):
-            guard let keyData = backendKeyData else {
-                // `backendKeyData` must have been received, before receiving the first `readyForQuery`
-                return self.closeConnectionAndCleanup(.unexpectedBackendMessage(.readyForQuery(transactionState)))
-            }
-            
             let connectionContext = ConnectionContext(
-                processID: keyData.processID,
-                secretKey: keyData.secretKey,
+                processID: backendKeyData?.processID,
+                secretKey: backendKeyData?.secretKey,
                 parameters: parameters,
                 transactionState: transactionState)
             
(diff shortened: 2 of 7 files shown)


fabianfett (commented):
Thanks for bringing this up! I think this is something that we should definitely support. However, I think, we should make all of this way more explicit than currently proposed:

1. We should add a setting to `PostgresConnection.Configuration.Conntection` that allow connections to not provide `BackendKeyData`. A possible name should be: `allowNotProvidingBackendKeyData`. By default this property should be set to false.
2. We need to inject the value of the property into the StateMachine.
3. If we receive `readyForQuery` without having received `backendKeyData` before, we should check, if the property `allowNotProvidingBackendKeyData` is true and only progress in this case.
4. We need unit test cases that cover both flows.

Is this something you feel comfortable tackling? I'm here to help you at any step on the way.

rausnitz:
Can do. I'll work on those changes. Thanks!

rausnitz:
Actually, handling this in the unit tests was straightforward, so I went ahead and updated it.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -77,13 +77,23 @@ public final class PostgresConnection {
...
+            public var requireBackendKeyData: Bool
+            
+            public init(
+                host: String,
+                port: Int = 5432,
+                requireBackendKeyData: Bool = true
+            ) {
                 self.host = host
                 self.port = port
+                self.requireBackendKeyData = requireBackendKeyData
```
fabianfett, line 93:
```suggestion
            public init(host: String, port: Int = 5432) {
                self.host = host
                self.port = port
                self.requireBackendKeyData = true
```

Sadly the proposed change is an API breaking change, because of this...

Existing API:
```swift
struct Foo {
    init(bar: String) {}

    static func test() {
        let make = Foo.init(bar:)
        let foo = make("hello")
        // ...
    }
}
```

Adding a defaulted value breaks this code however:
```swift
struct Foo {
    init(bar: String, test: String = "foo") {}

    static func test() {
        let make = Foo.init(bar:) // <--- ❌ compiler error
        let foo = make("hello")
        // ...
    }
}
```

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -77,13 +77,23 @@ public final class PostgresConnection {
...
             }
         }
-
+
```
fabianfett, line 96: Revert change here.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -377,6 +387,7 @@ extension PostgresConnection {
         to socketAddress: SocketAddress,
         tlsConfiguration: TLSConfiguration? = nil,
         serverHostname: String? = nil,
+        requireBackendKeyData: Bool = true,
```
fabianfett, line 390: This would be a breaking change. For now, let's not add the option here. Users who want to the new feature, shall use async await API, or create a separate PR.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -394,7 +405,8 @@ extension PostgresConnection {
             let configuration = PostgresConnection.InternalConfiguration(
                 connection: .resolved(address: socketAddress, serverName: serverHostname),
                 authentication: nil,
-                tls: tls
+                tls: tls,
+                requireBackendKeyData: requireBackendKeyData
```
fabianfett, line 409: Just set to true. See above comment.
```suggestion
                requireBackendKeyData: true
```

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -756,6 +768,8 @@ extension PostgresConnection {
         var authentication: Configuration.Authentication?
 
         var tls: Configuration.TLS
+        
+        let requireBackendKeyData: Bool
```
fabianfett, line 772:
```suggestion
        var requireBackendKeyData: Bool
```

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -113,17 +113,20 @@ struct ConnectionStateMachine {
     }
     
     private var state: State
+    private let requireBackendKeyData: Bool
     private var taskQueue = CircularBuffer<PSQLTask>()
     private var quiescingState: QuiescingState = .notQuiescing
     
-    init() {
+    init(requireBackendKeyData: Bool = true) {
```
fabianfett, line 120: We should not default this property here. Whenever we use the state machine we should be explicit.

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -543,14 +546,13 @@ struct ConnectionStateMachine {
...
-                // `backendKeyData` must have been received, before receiving the first `readyForQuery`
+            if self.requireBackendKeyData && backendKeyData == nil {
                 return self.closeConnectionAndCleanup(.unexpectedBackendMessage(.readyForQuery(transactionState)))
             }
             
             let connectionContext = ConnectionContext(
-                processID: keyData.processID,
-                secretKey: keyData.secretKey,
+                processID: backendKeyData?.processID,
+                secretKey: backendKeyData?.secretKey,
```
fabianfett, line 555: just pass in `backendKeyData`. See comment regarding making this a single property on `ConnectionContext`.

fabianfett (commented):
This already looks really good. Don't get distracted by the amount of code comments... This is awesome work!

Would you mind addressing the comments and add a test case for the new code path in `ConnectionStateMachineTests.swift`?

Two examples tests are here:

```
    func testBackendKeyAndParameterStatusReceivedAfterAuthenticated() {
        var state = ConnectionStateMachine(.authenticated(nil, [:]))
        
        XCTAssertEqual(state.backendKeyDataReceived(.init(processID: 2730, secretKey: 882037977)), .wait)
        
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "DateStyle", value: "ISO, MDY")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "application_name", value: "")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "server_encoding", value: "UTF8")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "integer_datetimes", value: "on")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "client_encoding", value: "UTF8")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "TimeZone", value: "Etc/UTC")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "is_superuser", value: "on")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "server_version", value: "13.1 (Debian 13.1-1.pgdg100+1)")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "session_authorization", value: "postgres")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "IntervalStyle", value: "postgres")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "standard_conforming_strings", value: "on")), .wait)
        
        XCTAssertEqual(state.readyForQueryReceived(.idle), .fireEventReadyForQuery)
    }
    
    func testReadyForQueryReceivedWithoutBackendKeyAfterAuthenticated() {
        var state = ConnectionStateMachine(.authenticated(nil, [:]))
        
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "DateStyle", value: "ISO, MDY")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "application_name", value: "")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "server_encoding", value: "UTF8")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "integer_datetimes", value: "on")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "client_encoding", value: "UTF8")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "TimeZone", value: "Etc/UTC")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "is_superuser", value: "on")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "server_version", value: "13.1 (Debian 13.1-1.pgdg100+1)")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "session_authorization", value: "postgres")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "IntervalStyle", value: "postgres")), .wait)
        XCTAssertEqual(state.parameterStatusReceived(.init(parameter: "standard_conforming_strings", value: "on")), .wait)
        
        XCTAssertEqual(state.readyForQueryReceived(.idle),
                       .closeConnectionAndCleanup(.init(action: .close, tasks: [], error: PSQLError.unexpectedBackendMessage(.readyForQuery(.idle)), closePromise: nil)))
    }
```

Again, thanks a lot for driving this forward...

Sources/PostgresNIO/Connection/PostgresConnection.swift
rausnitz: Hadn't thought of this scenario! Will address along with the other comments. Thank you.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -77,13 +77,23 @@ public final class PostgresConnection {
             ///
             /// - Default: 5432
             public var port: Int
-
-            public init(host: String, port: Int = 5432) {
+            
+            /// Require connection to provide `BackendKeyData`
```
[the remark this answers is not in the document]
rausnitz: This issue may go beyond RDS Proxy to include any Postgres-compatible service that does not support canceling a query, so I will make a note of that general issue as well. (One of my colleagues noticed that CockroachDB used to have a similar issue with a Postgres library in Python until they added a workaround: https://github.com/cockroachdb/cockroach/pull/60281.)

I'm wondering if we'd be better off with something like `canCancelQueries` instead of `requireBackendKeyData`. The behavior wouldn't have to change, but the purpose would be clearer to anyone configuring their Postgres integration. The `BackendKeyData` functionality in Postgres is pretty obscure.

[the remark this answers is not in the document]
fabianfett: The problem I have with `canCancelQueries` is that it would make users think they can not cancel queries in Swift. However this is not true. After a cancellation we can just ignore further data... or we just close the connection. Cancellation is not required to go trough the crazy Postgres way... However server cancellation works if we get backendKeyData.

So we could name this property `mustSupportServerSideCancellation`. Not sure if that makes anything clearer to users... 

I agree that this behavior goes beyond RDS proxy (for this reason the property shall not be called `supportRDSProxy`), but it is a use-case for it. For this reason I think it might make adopters life easier if they search for rds proxy in the code base.

fabianfett:
This looks really good now. Only thing left is merge of main into your branch!

rausnitz:
> This looks really good now. Only thing left is merge of main into your branch!

Done!

fabianfett (commented):
One Nit remaining...

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -5,9 +5,7 @@ struct ConnectionStateMachine {
     typealias TransactionState = PostgresBackendMessage.TransactionState
     
     struct ConnectionContext {
-        let processID: Int32
-        let secretKey: Int32
-        
+        let backendKeyData: BackendKeyData?
```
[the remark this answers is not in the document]
rausnitz: Neat, I didn't know that was better for the compiler.

fabianfett (approved):
Awesome patch @rausnitz! Thanks a ton!

rausnitz:
Thanks for the thorough and fast reviews @fabianfett!

PR #298, opened 2022-06-07
Increase bind parameters limit

MahdiBM:
# The problem
Right now PostgresNIO doesn't allow queries with bind parameters count more than Int16.max (32768).
This is despite Postgres 10+ supporting up to ~~(UInt16.max - 1)~~ UInt16.max parameters (65535).
For reference for Postgres 14, see https://www.postgresql.org/docs/14/postgres-fdw.html and search the page for "batch_size". I couldn't find documentation on this on lower Postgres versions, but manually tested the PR (see below).

# Solution
This PR increases that limit to 65535 which is more appropriate.

# Tests
I don't think there is a need for adding a test, but I changed one of the test's to reflect the changes. We'll see if there is anything I missed, when the CI runs the test.

Other than that, I've manually tested this on Postgres Docker images 10, 11, 12, 13, 14 (The currently supported-by-Postgres versions).

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -309,7 +309,7 @@ public final class PostgresConnection {
     private func queryStream(_ query: PostgresQuery, logger: Logger) -> EventLoopFuture<PSQLRowStream> {
         var logger = logger
         logger[postgresMetadataKey: .connectionID] = "\(self.id)"
-        guard query.binds.count <= Int(Int16.max) else {
+        guard query.binds.count < Int(UInt16.max) else {
             return self.channel.eventLoop.makeFailedFuture(PSQLError.tooManyParameters)
         }
 
@@ -341,7 +341,7 @@ public final class PostgresConnection {
     }
 
     func execute(_ executeStatement: PSQLExecuteStatement, logger: Logger) -> EventLoopFuture<PSQLRowStream> {
-        guard executeStatement.binds.count <= Int(Int16.max) else {
+        guard executeStatement.binds.count < Int(UInt16.max) else {
             return self.channel.eventLoop.makeFailedFuture(PSQLError.tooManyParameters)
         }
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
@@ -498,7 +498,7 @@ extension PostgresConnection {
         var logger = logger
         logger[postgresMetadataKey: .connectionID] = "\(self.id)"
 
-        guard query.binds.count <= Int(Int16.max) else {
+        guard query.binds.count < Int(UInt16.max) else {
             throw PSQLError.tooManyParameters
         }
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -309,7 +309,7 @@ public final class PostgresConnection {
     private func queryStream(_ query: PostgresQuery, logger: Logger) -> EventLoopFuture<PSQLRowStream> {
         var logger = logger
         logger[postgresMetadataKey: .connectionID] = "\(self.id)"
-        guard query.binds.count < Int(UInt16.max) else {
+        guard query.binds.count <= Int(UInt16.max) else {
             return self.channel.eventLoop.makeFailedFuture(PSQLError.tooManyParameters)
         }
 
@@ -341,7 +341,7 @@ public final class PostgresConnection {
     }
 
     func execute(_ executeStatement: PSQLExecuteStatement, logger: Logger) -> EventLoopFuture<PSQLRowStream> {
-        guard executeStatement.binds.count < Int(UInt16.max) else {
+        guard executeStatement.binds.count <= Int(UInt16.max) else {
             return self.channel.eventLoop.makeFailedFuture(PSQLError.tooManyParameters)
         }
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
@@ -498,7 +498,7 @@ extension PostgresConnection {
         var logger = logger
         logger[postgresMetadataKey: .connectionID] = "\(self.id)"
 
-        guard query.binds.count < Int(UInt16.max) else {
+        guard query.binds.count <= Int(UInt16.max) else {
             throw PSQLError.tooManyParameters
         }
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
--- a/Tests/IntegrationTests/PSQLIntegrationTests.swift
+++ b/Tests/IntegrationTests/PSQLIntegrationTests.swift
@@ -337,7 +337,7 @@ final class IntegrationTests: XCTestCase {
         let eventLoop = eventLoopGroup.next()
         
         try await withTestConnection(on: eventLoop) { connection in
-            // Mac binds limit is UInt16.max which is 65535 which is 3 * 5 * 17 * 257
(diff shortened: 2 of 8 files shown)


Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -309,7 +309,7 @@ public final class PostgresConnection {
     private func queryStream(_ query: PostgresQuery, logger: Logger) -> EventLoopFuture<PSQLRowStream> {
         var logger = logger
         logger[postgresMetadataKey: .connectionID] = "\(self.id)"
-        guard query.binds.count <= Int(Int16.max) else {
+        guard query.binds.count < Int(UInt16.max) else {
```
fabianfett, line 312: `UInt16.max` is `65535`. Given that `65535` are allowed, shouldn't this be `<=`?

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -341,7 +341,7 @@ public final class PostgresConnection {
     }
 
     func execute(_ executeStatement: PSQLExecuteStatement, logger: Logger) -> EventLoopFuture<PSQLRowStream> {
-        guard executeStatement.binds.count <= Int(Int16.max) else {
+        guard executeStatement.binds.count < Int(UInt16.max) else {
```
fabianfett, line 344: Same as above?

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -498,7 +498,7 @@ extension PostgresConnection {
         var logger = logger
         logger[postgresMetadataKey: .connectionID] = "\(self.id)"
 
-        guard query.binds.count <= Int(Int16.max) else {
+        guard query.binds.count < Int(UInt16.max) else {
```
fabianfett, line 501: Same as above.

fabianfett (commented):
@MahdiBM thanks for bringing this up. And sorry for taking so long to get back to you. Since we are testing the boundaries of Postgres here, would you mind adding an integration test, that ensures that Postgres supports what we allow?

fabianfett:
The best way to integration test this will be a test case like this:

```
    func testBindMaximumParameters() async throws {
        defer { XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully()) }
        let eventLoop = eventLoopGroup.next()

        try await withTestConnection(on: eventLoop) { connection in
            var binds = PostgresBindings()
            let max = 1664 //Int16.max - 1
            let range = (0..<Int(max))
            for num in range {
                try binds.append(num, context: .default)
            }

            let query = PostgresQuery(unsafeSQL: "SELECT \(range.map({"$\($0 + 1)"}).joined(separator: ","))", binds: binds)
            let rows = try await connection.query(query, logger: .psqlTest)
            var iterator = rows.makeAsyncIterator()
            let firstRow = try await iterator.next()
            XCTAssertEqual(firstRow?.count, Int(max))
            let done = try await iterator.next()
            XCTAssertNil(done)
        }
    }
```

However we will need to extend this a bit:

- Apparently a row in postgres has a max of `1664` columns
  - This means to test the max binding we need to create a table and then do a multirow insert into it.
  - The test case should create a table with 1000 columns first (can all be integer) 
- Create a `PSQLQuery` manually, with our new max parameter binding and insert 66 rows.
- There are not enough available parameters for the last row (short of 465 parameters). Just fill those spaces with the numbers directly and no parameter bindings there.
- Cleanup the table after the test...

Sources/PostgresNIO/Connection/PostgresConnection.swift
MahdiBM: Correct, I forgot everything starts from 0 😅

MahdiBM:
I corrected the equality checks and added the integration test.

Please let me know if there are any problems or if any improvements can be made.

MahdiBM:
Looking at your original test example, i think it would have been better if i asserted that the inserted rows count is what we expect after the insertion query.
Right now, if there are no errors thrown from the insertion query, the test assumes everything has been successful. I think (?) this could go wrong?

MahdiBM:
I went ahead and added a query to assert that all the rows were inserted successfully. Since all columns are `NOT NULL`, that would guarantee that the max binds test has been successful. 

As an alternative, in the Postgres documentation I read that the `INSERT` command returns the count of inserted rows, but when i tried to use that in the test, i realized there is no nice way of accessing the a command's tag, so basically the count of inserted rows after a `INSERT` command is inaccessible and we need another query to check for that.

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -329,4 +329,60 @@ final class IntegrationTests: XCTestCase {
...
+        let eventLoop = eventLoopGroup.next()
+        
+        try await withTestConnection(on: eventLoop) { connection in
+            // Mac binds limit is UInt16.max which is 65535 which is 3 * 5 * 17 * 257
+            // Max columns limit is 1664, so we will only make 5 * 257 columns which is less
+            // Then we will insert 3 * 17 rows
+            // In the insertion, there will be a total of 3 * 17 * 5 * 257 == UInt16.max bindings
+            // If the test is successful, it means Postgres supports UInt16.max bindings
+            let columnsCount = 5 * 257
+            let rowsCount = 3 * 17
```
fabianfett, line 346: Huge props to the prime factorization here!

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -329,4 +329,60 @@ final class IntegrationTests: XCTestCase {
...
+            for _ in (0..<rowsCount) {
+                for num in (0..<columnsCount) {
+                    try binds.append(num, context: .default)
+                }
+            }
+            XCTAssertEqual(binds.count, Int(UInt16.max))
+            
+            let insertionValues = (0..<rowsCount).map { rowIndex in
+                let indices = (0..<columnsCount).map { columnIndex -> String in
+                    "$\(rowIndex * 1_000 + columnIndex + 1)"
```
fabianfett, line 367: Where is the `1_000` coming from? I would suspect this to be `1285` (5 * 257 which is the columns per row).

MahdiBM: I'll need to fix that. I probably forgot about it after the first stages of writing the test function.

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -329,4 +329,60 @@ final class IntegrationTests: XCTestCase {
...
+                }
+                return "(\(indices.joined(separator: ", ")))"
+            }.joined(separator: ", ")
```
fabianfett, line 370: Since we know, we want to write exactly `65535` values, it might be easier to just reach for:
```
let insertionValues = (0..<UInt16.max).map { "$\(Int($0) + 1)" }.joined(separator: ", ")
```

fabianfett (commented):
This is really awesome. One question remaining

Tests/IntegrationTests/PSQLIntegrationTests.swift
MahdiBM: Good idea!

MahdiBM: However, doesn't seem to work since we'll be missing some parentheses.
Currently, and correctly, the values will look like `($1, $2, ...), ($x, $x+1, ...), ...` but with the simplified calculation, we'll be missing the parentheses, which postgres doesn't like.

fabianfett: yes makes sense!

fabianfett (approved):

PR #299, opened 2022-06-26
Use provided logger consistently in `PostgresConnection.send(_:logger:)`

gwynne:
Probably affects no one whatsoever, since using a non-default logger per-query isn't actually implemented anywhere that I've seen, but might as well be consistent just the same.

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -601,7 +601,7 @@ extension PostgresConnection: PostgresDatabase {
             }
 
         case .prepareQuery(let request):
-            resultFuture = self.prepareStatement(request.query, with: request.name, logger: self.logger).map {
+            resultFuture = self.prepareStatement(request.query, with: request.name, logger: logger).map {
                 request.prepared = PreparedQuery(underlying: $0, database: self)
             }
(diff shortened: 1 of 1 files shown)


0xTim (approved):

fabianfett (changes requested):
I'm pretty sure this is not needed.

https://github.com/vapor/postgres-nio/blob/2fa1cb1c136e4256daee191068b7795625733c2a/Sources/PostgresNIO/Connection/PostgresDatabase%2BPreparedQuery.swift#L8

gwynne:
@fabianfett It's less about whether it's actually needed and more about being consistent with the usage elsewhere in the same method 🤷‍♀️. I mean, at the end of the day, there's no reason that the send methods of all the Fluent drivers take a logger parameter to begin with; nothing ever passes anything other than the existing logger to that parameter, it's not exposed to any of the higher laters, and it's not required by any protocol. It's just another of the weird warts on Fluent's API design. (And furthermore, on general principle of crabbiness, I accuse Mongo of being at fault, despite having made no effort whatsoever to find evidence to that effect 😆.)

That all being said, why not be consistent?

fabianfett (approved):

fabianfett:
I'm stupid. You are right...

PR #300, opened 2022-07-07
Fix new NIO warnings

fabianfett:
This PR brings us up to speed with current Swift NIO main branch and fixes the warnings that will show up once a new NIO release is cut.


gwynne (approved):
LGTM so far

PR #301, opened 2022-07-09
Adding uds connection support to remove the need for deprecated socket address apis

MahdiBM:
### Problem

The ability to connect through a socket address was deprecated in recent changes.

### Changes

This PR adds a new way to connect through a socket address.
The only drawback is deprecation of `host` and `post` in `PostgresConnection.Configuration.Connection`. 
They will continue to work as properly as possible.

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -68,15 +68,55 @@ public final class PostgresConnection {
         }
 
         public struct Connection {
+            
+            enum Base {
+                case unresolved(host: String, port: Int)
+                case resolved(address: SocketAddress)
+            }
+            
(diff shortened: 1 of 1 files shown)


gwynne (approved):
Looks great to me! @fabianfett?

MahdiBM:
For more info, I did notice that Fabian mentioned he would rather the public API only accept a uds and internally resolve it to a `SocketAddress`, but when looking at the postgres-kit, i didn't know how that would solve the problem there. A few minutes ago i took another look at postgres-kit and finally noticed it will solve the problem there too, though will need some more modifications in the postgres-kit side. 
the new commit will make it so the `SocketAddress` is not exposed anymore.

perhaps a better way to name this PR is now "Adding uds connection support to remove the need for deprecated socket address apis".

PR #305, opened 2022-09-27
Replace Lock with new NIOLock

fabianfett:
SwiftNIO 2.42.0 has deprecated Lock and replaced it with a new NIOLock. This commit removes all uses of Lock and replaces them with NIOLock. Further, require new package versions.


0xTim (approved):
LGTM

PR #306, opened 2022-10-03
Update test CI

MahdiBM:
The test CI could use a bit of an update :)
NIO ~~has recently dropped~~ will soon drop support for swift 5.4, so we should drop swift 5.4 too.
This also updates integration-tests and api-breakage jobs to use swift 5.7.
**EDIT 1**:
Now this PR also removes the codecov checks from nightly builds because codecov doesn't seem to get along with nightlies.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ jobs:
         run: swift test --enable-test-discovery --filter=^PostgresNIOTests --sanitize=thread --enable-code-coverage
       - name: Submit coverage report to Codecov.io
         if: "!contains(matrix.container, 'nightly')"
-        uses: vapor/swift-codecov-action@v0.2.1
+        uses: vapor/swift-codecov-action@v0.2
         with:
           cc_flags: 'unittests'
           cc_env_vars: 'SWIFT_VERSION,SWIFT_PLATFORM,RUNNER_OS,RUNNER_ARCH'


fabianfett (approved):
LGTM. We need a code owner to merge and adjust the protected branch here though. Pinging @gwynne, @0xTim.

.github/workflows/test.yml
```diff
@@ -29,7 +27,8 @@ jobs:
       - name: Run unit tests with code coverage and Thread Sanitizer
         run: swift test --enable-test-discovery --filter=^PostgresNIOTests --sanitize=thread --enable-code-coverage
       - name: Submit coverage report to Codecov.io
-        uses: vapor/swift-codecov-action@v0.2
+        if: "!contains(matrix.container, 'nightly')"
+        uses: vapor/swift-codecov-action@v0.2.1
```
gwynne, line 31: The `v0.2` tag points to the latest "0.2.x" version, not `0.2.0`; it has the effect of SPM's `upToNextMinor()`, very much on purpose, please keep it that way 🙂

.github/workflows/test.yml
```diff
@@ -12,14 +12,12 @@ jobs:
...
-        swift:
-          - swift:5.4
-          - swift:5.5
-          - swift:5.6
-          - swiftlang/swift:nightly-main
-        os:
-          - focal
-    container: ${{ format('{0}-{1}', matrix.swift, matrix.os) }}
+        container:
+          - swift:5.5-bionic
```
gwynne, line 16: What is the benefit to testing on `bionic`?

gwynne (changes requested):

.github/workflows/test.yml
MahdiBM: Good to know!

.github/workflows/test.yml
fabianfett: We have a nice mix of oses and swift versions...

gwynne (approved):

PR #307, opened 2022-10-03
Remove need to supply PostgresDecodingContext on decode

fabianfett:
### Motivation

To use a customized JSONDecoder, users have the option to supply a `PostgresDecodingContext`. Currently we require users to set explicitly on every row decode call. This makes the API unnecessarily verbose. Since the `PostgresDecodingContext` is generic we can not default the parameter. Instead we need add proxy functions so that we can default the parameter manually.

### Changes

- Add decode methods omitting the `PostgresDecodingContext`


gwynne (approved):
LGTM

fabianfett:
The API breakage detector reports a false positive here. The decode methods have been moved from `PostgresRowSequence` to `extension AsyncSequence where Element == PostgresRow`, which `PostgresRowSequence` conforms to.

PR #313, opened 2022-10-12
Deprecating PostgresDataConvertible, PostgresMessageType

fabianfett:
### Motivation

We want to inform users of patterns that will be going away.

### Changes

- deprecate `PostgresDataConvertible`
- deprecate `PostgresMessageType`
- deprecate `PostgresJSONCodable`
- deprecate `PostgresJSONBCodable`

### DIFF
--- a/Sources/PostgresNIO/Data/PostgresData+Array.swift
+++ b/Sources/PostgresNIO/Data/PostgresData+Array.swift
@@ -1,14 +1,14 @@
 import NIOCore
 
 extension PostgresData {
-    public init<T>(array: [T])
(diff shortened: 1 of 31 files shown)


gwynne (changes requested):
Deprecating `PostgresDataConvertible` would cause `postgres-kit` to emit multiple deprecation warnings in anything using it (typically via Fluent) which users would have no way to silence. @0xTim any thoughts?

gwynne (approved):
This PR will be unblocked by completion of https://github.com/vapor/postgres-kit/issues/237, at that time there are no further objections to merging it.

PR #314, opened 2022-10-21
Add method to connect using existing channel

tkrajacic:
Add a `connect` method that allows using an existing `Channel` for the connection.

This makes it possible to for example create an SSH tunnel and establish a Postgres connection on the SSH Channel.


fabianfett:
Superseeded by https://github.com/vapor/postgres-nio/pull/335.

PR #316, opened 2022-10-26
AsyncStreamConsumer should stream DataRows.

fabianfett:
### Motivation

We want to replace our `AsyncStreamConsumer` implementation with the new `NIOThrowingAsyncSequenceProducer`. To do this we must ensure that the producing and consuming side of our AsyncSequence operate on the same type.

### Changes

- Create the `PostgresRow` later


gwynne (approved):
LGTM

PR #317, opened 2022-10-26
Use NIOThrowingAsyncSequenceProducer

fabianfett:
### Motivation

We should use components from NIO where possible and not rely on our own. Learnings from PostgresNIO have influenced  the new very shiny `NIOThrowingAsyncSequenceProducer`. We should adopt sooner rather than later.

### Changes

- Use `NIOThrowingAsyncSequenceProducer` and drop our own `AsyncStreamConsumer`

### Result

We support task cancellation on `iterator.next()`

### DIFF
--- a/Sources/PostgresNIO/New/PSQLRowStream.swift
+++ b/Sources/PostgresNIO/New/PSQLRowStream.swift
@@ -1,10 +1,5 @@
-#if swift(>=5.6)
 import NIOCore
 import Logging
-#else
-@preconcurrency import NIOCore
-@preconcurrency import Logging
-#endif
 
 final class PSQLRowStream {
     private typealias AsyncSequenceSource = NIOThrowingAsyncSequenceProducer<DataRow, Error, AdaptiveRowBuffer, PSQLRowStream>.Source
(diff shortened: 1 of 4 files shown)


gwynne (approved):
LGTM

Sources/PostgresNIO/New/PSQLRowStream.swift
```diff
@@ -128,10 +139,10 @@ final class PSQLRowStream {
 
     private func cancel0() {
         switch self.downstreamState {
-        case .asyncSequence(let consumer, let dataSource):
-            let error = PSQLError.connectionClosed
+        case .asyncSequence(let source, let dataSource):
+            let error = CancellationError()
```
FranzBusch, line 143: I don't fully agree with the usage of `CancellationError` here. It should only get thrown if the task is cancelled; however, in your code path it code be thrown when the source terminated the sequence where this would be quite confusing for the user. In general, you don't have to call `source.finish` as a result of `didTerminate` at all.

FranzBusch (commented):
Nice! Glad this all worked out perfectly. I left a small comment inline

FranzBusch (approved):

PR #321, opened 2022-11-10
PostgresCodable should be a typealias

fabianfett:
### Motivation

`PostgresCodable` only marks a type as `PostgresEncodable` and `PostgresDecodable`. For this reason this should not be an extra protocol. Instead it should be a typealias. (See same pattern in Codable – which also is just a typealias).

Lucky for us we never made `PostgresCodable` public so far.

### Changes

- `PostgresCodable` is a typealias


gwynne (approved):
Nice!!! I didn't even notice that. (Not that `Codable` has much track record for getting things right, but this is one case where it definitely makes more sense this way.)

PR #322, opened 2022-11-10
Add new protocol PostgresNonThrowingEncodable

fabianfett:
### Motivation

Currently creating a `PostgresQuery` through String interpolation always throws. The reason to support throwing in `PostgresEncodable` is, that the `Encodable` protocol enforces throwing (and we do want to support JSON). But for most of the primary types throwing is totally useless.

### Changes

- We introduce the protocol `PostgresNonThrowingEncodable`, which inherits from `PostgresEncodable` but enforces that the `encode` method does not `throw`
- We add a separate entry points for `PostgresNonThrowingEncodable` in `PostgresQuery` and `PostgresBinding` to reduce the number of places where we throw

### Result

- way less throwing in places where we don't actually `throw`

### DIFF
--- a/Sources/PostgresNIO/New/Data/Array+PostgresCodable.swift
+++ b/Sources/PostgresNIO/New/Data/Array+PostgresCodable.swift
@@ -124,6 +124,43 @@ extension Array: PostgresEncodable where Element: PostgresArrayEncodable {
     }
 }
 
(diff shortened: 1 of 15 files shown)


gwynne (approved):
\*mutter mutter grumble\*

PR #324, opened 2022-11-13
fix en- and decoding of NULL values in arrays

### LINKED ISSUE #323: Arrays with `NULL` values get encoded/decoded incorrectly
### Describe the bug

When having an optional array (such as `[Int32?]`) as a bind, then any `nil` values it contains get encoded incorrectly which causes to the following error:
> insufficient data left in message (pq_copymsgbytes)

When receiving arrays that contain `NULL`s from the database and decoding them, then `postgres-nio` will fail an assertion and crash.
https://github.com/vapor/postgres-nio/blob/607279d6893dd0332263489cc725255bfcfe520f/Sources/PostgresNIO/Data/PostgresData%2BArray.swift#L83
### To Reproduce

Running this query will throw the error:
```swift
conn.query(
    """
    select $1::int8[] as array
    """,
    [PostgresData(array: [1, nil, 3] as [Int?])]
)
```

And this snippet will crash:
```swift
let rows = try conn.query(
    """
    select'{1, NULL, 3}'::int8[] as array
    """
).wait()
_ = rows.first?.makeRandomAccess()[data: "array"].array(of: Int?.self)
```

### Expected behavior

Using optional arrays in binds doesn't throw an error, and reading arrays containing `NULL`s doesn't lead to a crash.

### Environment

* Vapor Framework version: PostgresNIO 1.12.1
* Postgresql version: PostgreSQL 14.5 (Homebrew) on x86_64-apple-darwin21.6.0, compiled by Apple clang version 14.0.0 (clang-1400.0.29.102), 64-bit
* OS version: MacOS 13.0

### Additional context
After digging around the source code, I believe the encoding error stems from these lines:
https://github.com/vapor/postgres-nio/blob/607279d6893dd0332263489cc725255bfcfe520f/Sources/PostgresNIO/Data/PostgresData%2BArray.swift#L28-L35
Instead of zero, `nil` values should be encoded as a two's complement -1.

The assertion failure is caused by a flag that Postgres sets for arrays containing `NULL`s.

ehpi:
Fixes #323

Fixes the en- and decoding of arrays containing `NULL` values.

### DIFF
--- a/Sources/PostgresNIO/Data/PostgresData+Array.swift
+++ b/Sources/PostgresNIO/Data/PostgresData+Array.swift
@@ -13,7 +13,7 @@ extension PostgresData {
         var buffer = ByteBufferAllocator().buffer(capacity: 0)
         // 0 if empty, 1 if not
         buffer.writeInteger(array.isEmpty ? 0 : 1, as: UInt32.self)
-        // b
+        // b - this gets ignored by psql
         buffer.writeInteger(0, as: UInt32.self)
         // array element type
         buffer.writeInteger(elementType.rawValue)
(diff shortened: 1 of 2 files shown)


fabianfett (approved):
This is awesome! Thanks!

PR #329, opened 2023-02-19
Update API Docs Workflow

0xTim:
The API docs workflow is now working so we can roll this out.

Because of the way exporting works on Linux and how DocC consumes it, I've ensured everything compiles when nothing is exported and wrapped the exports in our DocC flag


fabianfett (approved):
LGTM! Thanks

PR #330, opened 2023-03-05
Add support for int8range and int8range[]

rausnitz:
This PR adds support for the `int8range` Postgres type and the corresponding Swift type `Range<Int64>`. Postgres comes with built-in support for `int8range` and five other range types: https://www.postgresql.org/docs/current/rangetypes.html.

The test case I created in this PR demonstrates the new functionality. In short, the range `'[-9223372036854775808, 9223372036854775807)'::int8range` gets decoded to `-9223372036854775808..<9223372036854775807` and the range array `ARRAY['[0, 1)'::int8range, '[10, 11)'::int8range]` gets decoded to `[0..<1, 10..<11]`.

For now, I only wrote code for `int8range` and `int8range[]` because that's what I personally had a need for, but I'd be happy to add support for the other range types (and `ClosedRange` too), once I get confirmation that there is interest in adding this stuff to PostgresNIO.

Since these types are standard in Postgres, I think they belong in PostgresNIO.

### DIFF
--- a/Sources/PostgresNIO/Data/PostgresData+Range.swift
+++ b/Sources/PostgresNIO/Data/PostgresData+Range.swift
@@ -17,7 +17,7 @@ extension PostgresData {
         switch self.formatCode {
         case .binary:
             switch self.type {
-            case PostgresDataType(3926):
+            case .int8Range:
                 guard value.readInteger(as: Int8.self) == 2 else {
                     return nil
                 }
--- a/Sources/PostgresNIO/Data/PostgresData+Range.swift
+++ b/Sources/PostgresNIO/Data/PostgresData+Range.swift
@@ -1,0 +1,66 @@
+extension PostgresData {
+    public init(int8Range: Range<Int64>) {
+        var buffer = ByteBufferAllocator().buffer(capacity: 25)
+        buffer.writeInteger(2, as: Int8.self)
+        buffer.writeInteger(8, as: Int32.self)
+        buffer.writeInteger(int8Range.lowerBound)
+        buffer.writeInteger(8, as: Int32.self)
+        buffer.writeInteger(int8Range.upperBound)
+        self.init(type: .int8Range, formatCode: .binary, value: buffer)
+    }
+    
+    public var int8Range: Range<Int64>? {
+        guard var value = self.value else {
+            return nil
+        }
+        
+        switch self.formatCode {
+        case .binary:
+            switch self.type {
+            case PostgresDataType(3926):
+                guard value.readInteger(as: Int8.self) == 2 else {
+                    return nil
+                }
+
+                guard value.readInteger(as: Int32.self) == 8 else {
+                    return nil
+                }
+
+                guard let lowerBound: Int64 = value.readInteger() else {
+                    return nil
+                }
+
+                guard value.readInteger(as: Int32.self) == 8 else {
+                    return nil
+                }
+
+                guard let upperBound: Int64 = value.readInteger() else {
+                    return nil
+                }
+                
+                return lowerBound..<upperBound
+            default:
+                return nil
+            }
+        case .text:
+            return nil
+        }
+    }
+}
+
+extension Range: PostgresDataConvertible where Element == Int64 {
+    public static var postgresDataType: PostgresDataType {
+        return .int8Range
+    }
+    
+    public init?(postgresData: PostgresData) {
+        guard let range: Range<Int64> = postgresData.int8Range else {
+            return nil
+        }
+        self = range
+    }
+
+    public var postgresData: PostgresData? {
+        return .init(int8Range: self)
+    }
+}
--- a/Sources/PostgresNIO/Data/PostgresData.swift
+++ b/Sources/PostgresNIO/Data/PostgresData.swift
@@ -70,6 +70,10 @@ public struct PostgresData: CustomStringConvertible, CustomDebugStringConvertibl
             description = self.array(of: String.self)?.description
         case .jsonbArray:
             description = self.array?.description
+        case .int8Range:
+            description = self.int8Range?.description
+        case .int8RangeArray:
+            description = self.array(of: Range<Int64>.self)?.description
         default:
             if self.type.isUserDefined {
                 // custom type
--- a/Tests/IntegrationTests/PostgresNIOTests.swift
+++ b/Tests/IntegrationTests/PostgresNIOTests.swift
@@ -341,6 +341,23 @@ final class PostgresNIOTests: XCTestCase {
         XCTAssertEqual(UUID(uuidString: row?[data: "id"].string ?? ""), UUID(uuidString: "123E4567-E89B-12D3-A456-426655440000"))
     }
 
+    func testInt8Range() {
+        var conn: PostgresConnection?
+        XCTAssertNoThrow(conn = try PostgresConnection.test(on: eventLoop).wait())
+        defer { XCTAssertNoThrow( try conn?.close().wait() ) }
+        struct Model: Decodable {
+            let range: Range<Int64>
+        }
+        var results: PostgresQueryResult?
+        XCTAssertNoThrow(results = try conn?.query("""
+        SELECT
+            '[-9223372036854775808, 9223372036854775807)'::int8range AS range
+        """).wait())
+        XCTAssertEqual(results?.count, 1)
+        let row = results?.first?.makeRandomAccess()
+        XCTAssertEqual(row?[data: "range"].int8Range, -9223372036854775808..<9223372036854775807)
+    }
+
     func testDates() {
         var conn: PostgresConnection?
         XCTAssertNoThrow(conn = try PostgresConnection.test(on: eventLoop).wait())
--- a/Tests/IntegrationTests/PostgresNIOTests.swift
+++ b/Tests/IntegrationTests/PostgresNIOTests.swift
@@ -354,8 +354,19 @@ final class PostgresNIOTests: XCTestCase {
             '[-9223372036854775808, 9223372036854775807)'::int8range AS range
         """).wait())
         XCTAssertEqual(results?.count, 1)
-        let row = results?.first?.makeRandomAccess()
+        var row = results?.first?.makeRandomAccess()
         XCTAssertEqual(row?[data: "range"].int8Range, -9223372036854775808..<9223372036854775807)
+
+        XCTAssertNoThrow(results = try conn?.query("""
+        SELECT
+            ARRAY[
+                '[0, 1)'::int8range,
+                '[10, 11)'::int8range
+            ] AS ranges
+        """).wait())
+        XCTAssertEqual(results?.count, 1)
+        row = results?.first?.makeRandomAccess()
+        XCTAssertEqual(row?[data: "ranges"].array(of: Range<Int64>.self), [0..<1, 10..<11])
     }
 
     func testDates() {
--- a/Sources/PostgresNIO/Data/PostgresDataType.swift
+++ b/Sources/PostgresNIO/Data/PostgresDataType.swift
@@ -122,6 +122,8 @@ public struct PostgresDataType: RawRepresentable, Hashable, CustomStringConverti
     public static let jsonbArray = PostgresDataType(3807)
     /// `3926`
     public static let int8Range = PostgresDataType(3926)
+    /// `3927` _int8range
+    public static let int8RangeArray = PostgresDataType(3927)
 
     /// The raw data type code recognized by PostgreSQL.
     public var rawValue: UInt32
@@ -187,6 +189,8 @@ public struct PostgresDataType: RawRepresentable, Hashable, CustomStringConverti
         case .uuidArray: return "UUID[]"
         case .jsonb: return "JSONB"
         case .jsonbArray: return "JSONB[]"
+        case .int8Range: return "INT8RANGE"
+        case .int8RangeArray: return "INT8RANGE[]"
         default: return nil
         }
     }
@@ -208,6 +212,7 @@ public struct PostgresDataType: RawRepresentable, Hashable, CustomStringConverti
         case .jsonb: return .jsonbArray
         case .text: return .textArray
         case .varchar: return .varcharArray
+        case .int8Range: return .int8RangeArray
         default: return nil
         }
     }
@@ -230,6 +235,7 @@ public struct PostgresDataType: RawRepresentable, Hashable, CustomStringConverti
         case .jsonbArray: return .jsonb
         case .textArray: return .text
         case .varcharArray: return .varchar
+        case .int8RangeArray: return .int8Range
         default: return nil
         }
     }
--- a/Sources/PostgresNIO/Data/PostgresDataType.swift
+++ b/Sources/PostgresNIO/Data/PostgresDataType.swift
@@ -120,6 +120,8 @@ public struct PostgresDataType: RawRepresentable, Hashable, CustomStringConverti
     public static let jsonb = PostgresDataType(3802)
     /// `3807` _jsonb
     public static let jsonbArray = PostgresDataType(3807)
+    /// `3926`
+    public static let int8Range = PostgresDataType(3926)
 
     /// The raw data type code recognized by PostgreSQL.
     public var rawValue: UInt32


fabianfett (commented):
@rausnitz Thanks so much for raising this. Do you want to also add an implementation for `PostgresEncodable`, `PostgresDecodable` which will be the way forward. We plan to deprecate `PostgresData` sooner rather than later.

rausnitz:
> The part that could become an issue is that I specified the `Range` type in `PostgresData.init(int8Range:)` and `PostgresData.int8Range`. If we wanted to make those generic in the future after this code has been released with `Range` specified, it would probably be a breaking change at that point. But it's a bit tricky to use the generic `RangeExpression` because of the associated type, and I didn't know if it was worth coming up with a solution if `PostgresData` is going away soon anyway.

Update: The latest commit gets around this issue by adding `PostgresInt8RangeExpression`.

rausnitz:
The commit that I just pushed makes it straightforward to add the other Postgres range types in the future, not limited to `int8range`.

rausnitz:
In the latest commit, I took a different approach that is more straightforward and easier to extend: I created a `PostgresRange` struct type that contains the properties necessary to represent all the Swift `RangeExpression` types, with any kind of associated `Bound`. This turned out to be a lot cleaner and simpler than using protocols to represent Postgres ranges. The new `PostgresRange` type handles all the encoding to and decoding from binary. So, adding support for more `RangeExpression` types and `Bound` types should be very easy, both in the old `PostgresData` way and the new `PostgresCodable` way.

This PR now covers two `RangeExpression` types (`Range` and `ClosedRange`) with two `Bound` types (`Int32` and `Int64`). I can implement all the others supported by Postgres if @fabianfett approves of what I've got so far.

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        guard let postgresRange = PostgresRange<Bound>(
+            from: &buffer,
+            format: format
+        ) else {
+            throw PostgresDecodingError.Code.failure
+        }
+
+        guard let lowerBound: Bound = postgresRange.lowerBound,
```
fabianfett, line 43: Do we have to check if the range is inclusive here?

Sources/PostgresNIO/Data/PostgresData+Range.swift
```diff
@@ -0,0 +1,244 @@
...
+
+public protocol PostgresRangeBound: PostgresDataConvertible, Comparable {
+    /// the Postgres range type associated with this bound
+    static var rangeType: PostgresDataType { get }
+    
+    /// If a Postgres range type has a well-defined step,
+    /// Postgres automatically converts it to a canonical form.
+    /// Types such as `int4range` get converted to upper-bound-exclusive.
+    /// This method is needed when converting an upper bound to inclusive.
+    var stepDown: (() -> Self)? { get }
```
fabianfett, line 13: Why is that a property and not a simple method? Why should this be optional?

```swift
func decrement() -> Self
```

Sources/PostgresNIO/Data/PostgresData+Range.swift
```diff
@@ -0,0 +1,244 @@
...
+    public static var rangeType: PostgresDataType { return .int8Range }
+
+    public var stepDown: (() -> Int64)? {
+        { self - 1 }
+    }
+}
+
+// MARK: PostgresRange
+
+public struct PostgresRange<B: PostgresRangeBound>: CustomStringConvertible {
```
fabianfett, line 34: Do we need to make this type public?

Sources/PostgresNIO/Data/PostgresData+Range.swift
```diff
@@ -0,0 +1,244 @@
+import NIOCore
+
+// MARK: PostgresRangeBound
+
+public protocol PostgresRangeBound: PostgresDataConvertible, Comparable {
```
fabianfett, line 5: can we split this into `PostgresRangeEncodable` and `PostgresRangeDecodable`? Why does this inherit `Comparable`?

Sources/PostgresNIO/Data/PostgresData+Range.swift
```diff
@@ -0,0 +1,244 @@
...
+            upperBound = steppedDownUpperBound
+        }
+
+        self = lowerBound...upperBound
+    }
+
+    public var postgresData: PostgresData? {
+        return PostgresData(range: PostgresRange<Bound>.init(closedRange: self))
+    }
+}
```
fabianfett, line 239: Would it be okay for you if we don't add conformance to `PostgresDataConvertible`? We have already started deprecating this API in #313.

Sources/PostgresNIO/New/Data/Array+PostgresCodable.swift
```diff
@@ -85,6 +85,18 @@ extension UUID: PostgresArrayEncodable {
...
+
+extension Range: PostgresArrayEncodable where Bound: PostgresRangeBound {
+    public static var psqlArrayType: PostgresDataType { Bound.rangeType.arrayType! }
+}
+
+extension ClosedRange: PostgresArrayDecodable where Bound: PostgresRangeBound {}
+
+extension ClosedRange: PostgresArrayEncodable where Bound: PostgresRangeBound {
+    public static var psqlArrayType: PostgresDataType { Bound.rangeType.arrayType! }
+}
```
fabianfett, line 98: Having the separation into `PostgresRangeEncodable` and `PostgresRangeDecodable` will make this much nicer.

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
```diff
@@ -0,0 +1,104 @@
+import NIOCore
+
+extension RangeExpression where Bound: PostgresRangeBound {
+    public static var psqlFormat: PostgresFormat { return .binary }
+}
```
fabianfett, line 5: Can we link the postgres docs here, why this is the case?

fabianfett (commented):
Sorry for taking so long to come back to you. This already is in a nice shape. However there are a few things that I would like to bring into the same shape as the other Encodable, Decodable implementations.

rausnitz:
@fabianfett Thanks for the review! I'll respond to your comments in the next few days.

Sources/PostgresNIO/Data/PostgresData+Range.swift
rausnitz: I just pushed a commit where I limited the usage of `Comparable`: https://github.com/vapor/postgres-nio/pull/330/commits/3b2989981c31866306debab7eb6a87832ee02497. The reason it's needed at all is because in Swift's `RangeExpression` types, the `Bound` has to be `Comparable`: https://developer.apple.com/documentation/swift/rangeexpression/bound.

Sources/PostgresNIO/Data/PostgresData+Range.swift
rausnitz: See new commit https://github.com/vapor/postgres-nio/pull/330/commits/52f1d292197f6043bb242855cd6a3a084b864046

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
rausnitz: See new commit https://github.com/vapor/postgres-nio/pull/330/commits/5aa2b1924a7af050efba4d7e05a5936cd26ecee7

Sources/PostgresNIO/Data/PostgresData+Range.swift
rausnitz: I changed it to a method here: https://github.com/vapor/postgres-nio/pull/330/commits/1b3d009c04c6bab14af5ba48d402d1c4a9a86137. (I forget why I made it a property initially).

I also changed it from returning an optional value to making the function throw, which I think is clearer. The reason it can't always return a value is because some Postgres range types are "continuous" and don't have a well-defined step. This includes the built-in Postgres types numrange, tsrange, and tstzrange. Section 8.17.7 of the docs here explains it: https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-DISCRETE.

I originally attempted to do something with Swift's `Strideable` but couldn't come up with a solution. Here's essentially what I would like to do in the `ClosedRange` conformance to `PostgresDecodable`, though this code isn't valid Swift:
```
if let strideable = upperBound as? Strideable {
    upperBound = strideable.advanced(by: -1)
}
```

Sources/PostgresNIO/Data/PostgresData+Range.swift
rausnitz: If I remove `public`, I see build errors such as `generic struct 'PostgresRange' is internal and cannot be referenced from an '@inlinable' function` on this line: https://github.com/vapor/postgres-nio/pull/330/files/61b4d5939e43d88ce92406c07645427dd561ac4e#diff-5c0292b699e71becd70812877bc88e53d4e1c18d061b195dacfcdfbb0d8f8faaR17. I could try removing `@inlinable` but the other existing conformances to `PostgresDecodable` and `PostgresEncodable` use it so I figured I should leave it.

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
rausnitz: I'm actually not sure, though I did notice that all the other implementations of `PostgresEncodable` return `.binary` as the value for their `psqlFormat`.

Is there a scenario where we want to enable the option of `.text` even though `.binary` works?

Sources/PostgresNIO/Data/PostgresData+Range.swift
rausnitz: I replaced `PostgresRangeBound` with `PostgresRangeEncodable` and `PostgresRangeDecodable` here: https://github.com/vapor/postgres-nio/pull/330/commits/9b8b3ec7a9de3031df8a84b54050d1317cfb61b4.

Sources/PostgresNIO/Data/PostgresData+Range.swift
rausnitz: Another solution would be to add `@usableFromInline` in a bunch of places, such as:
```
@usableFromInline
struct PostgresRange<B> {
    @usableFromInline let lowerBound: B?
    @usableFromInline let upperBound: B?
    @usableFromInline let isLowerBoundInclusive: Bool
    @usableFromInline let isUpperBoundInclusive: Bool

    init(
        lowerBound: B?,
        upperBound: B?,
        isLowerBoundInclusive: Bool,
        isUpperBoundInclusive: Bool
    ) {
        self.lowerBound = lowerBound
        self.upperBound = upperBound
        self.isLowerBoundInclusive = isLowerBoundInclusive
        self.isUpperBoundInclusive = isUpperBoundInclusive
    }
}
```

rausnitz:
@fabianfett This is ready for another review.

In the new changes, one minor thing I haven't been able to figure out is why this line https://github.com/vapor/postgres-nio/pull/330/files#diff-b065d49f245f03d7edb908c32f395835e399ecc588321d03b6142e2a42dd606aR75 requires `PostgresEncodable & PostgresNonThrowingEncodable` instead of just `PostgresNonThrowingEncodable`.

Removing `PostgresEncodable` gives me this build error:
```
conditional conformance of type 'PostgresRange<B>' to protocol 'PostgresNonThrowingEncodable' does not imply conformance to inherited protocol 'PostgresEncodable'
```

I'm not sure why this is, considering `PostgresNonThrowingEncodable` inherits from `PostgresEncodable` and I know the implementation covers `PostgresEncodable` because I don't have to change any of the implementation (other than declaring it `PostgresEncodable`) to get it to build.

Sources/PostgresNIO/Data/PostgresData+Range.swift
fabianfett: yes please let's go for the second option! I really, really only want to add things to the public API that we need.

Sources/PostgresNIO/Data/PostgresData+Range.swift
fabianfett: Lovely!

Sources/PostgresNIO/Data/PostgresData.swift
```diff
@@ -58,6 +58,10 @@ public struct PostgresData: CustomStringConvertible, CustomDebugStringConvertibl
             var value = value
             value.moveReaderIndex(forwardBy: 1)
             description = String(decoding: value.readableBytesView, as: UTF8.self)
+        case .int4Range:
+            description = try? PostgresRange<Int32>(from: &value, type: type, format: formatCode, context: .default).description
+        case .int8Range:
+            description = try? PostgresRange<Int64>(from: &value, type: type, format: formatCode, context: .default).description
```
fabianfett, line 64: since we don't want to support this in PostgresData anymore, is this still needed?

Sources/PostgresNIO/Data/PostgresData.swift
```diff
@@ -70,6 +74,10 @@ public struct PostgresData: CustomStringConvertible, CustomDebugStringConvertibl
             description = self.array(of: String.self)?.description
         case .jsonbArray:
             description = self.array?.description
+        case .int4RangeArray:
+            description = try? [PostgresRange<Int32>](from: &value, type: type, format: formatCode, context: .default).description
+        case .int8RangeArray:
+            description = try? [PostgresRange<Int64>](from: &value, type: type, format: formatCode, context: .default).description
```
fabianfett, line 80: since we don't want to support this in PostgresData anymore, is this still needed?

Sources/PostgresNIO/Data/PostgresData+Range.swift
```diff
@@ -0,0 +1,133 @@
+import NIOCore
+
+// MARK: PostgresRange
+
+public struct PostgresRange<B> {
```
fabianfett, line 5: we should make this `@usableFromInline` instead of public. This way it is easier for us to change, if new requirements appear when we add conformance for dates and other stuff.

fabianfett (commented):
Thanks, this already looks much better! Let's try to keep the new external API as small as possible.

Sources/PostgresNIO/Data/PostgresData.swift
rausnitz: Removed here, along with two protocol conformances that enabled it: https://github.com/vapor/postgres-nio/pull/330/commits/24c92570c4c7537923ee723b9fb0efd15c7dbdc1

Sources/PostgresNIO/Data/PostgresData.swift
rausnitz: Removed here, along with two protocol conformances that enabled it: https://github.com/vapor/postgres-nio/pull/330/commits/24c92570c4c7537923ee723b9fb0efd15c7dbdc1

Sources/PostgresNIO/Data/PostgresData+Range.swift
rausnitz: See new commit https://github.com/vapor/postgres-nio/pull/330/commits/23cd10f201ff3f10344eabc25c455cdbb2111a3e

rausnitz:
> Thanks, this already looks much better! Let's try to keep the new external API as small as possible.

Agreed. I really appreciate the detailed reviews.

rausnitz:
Now that I've removed all the code related to `PostgresData`, should I rename `PostgresData+Range.swift` and should I move the code in that file elsewhere?

Tests/IntegrationTests/PostgresNIOTests.swift
```diff
@@ -341,6 +341,85 @@ final class PostgresNIOTests: XCTestCase {
...
+        SELECT
+            '[\(Int32.min), \(Int32.max))'::int4range AS range
+        """).wait())
+        XCTAssertEqual(results?.count, 1)
+        var row = results?.first?.makeRandomAccess()
+        let expectedRange: Range<Int32> = Int32.min..<Int32.max
+        var expectedRangeBinaryValue = ByteBuffer()
+        expectedRange.encode(into: &expectedRangeBinaryValue, context: .default)
+        var rangeBytes: ByteBuffer = try XCTUnwrap(row?["range"].bytes)
+        let postgresRange: PostgresRange<Int32>? = try PostgresRange<Int32>.init(from: &rangeBytes, type: .int4Range, format: .binary, context: .default)
```
fabianfett, line 362: can we try to go to to the final Range directly?

Thinking of:
```swift
  let decodedClosedRange = try row.decode(ClosedRange<Int32>.self)
  let decodedOpenRange = try row.decode(Range<Int32>.self)
```

Tests/IntegrationTests/PostgresNIOTests.swift
```diff
@@ -341,6 +341,85 @@ final class PostgresNIOTests: XCTestCase {
         XCTAssertEqual(UUID(uuidString: row?[data: "id"].string ?? ""), UUID(uuidString: "123E4567-E89B-12D3-A456-426655440000"))
     }
 
+    func testInt4Range() throws {
```
fabianfett, line 344: we should add tests with which we also test the encoding...

fabianfett (commented):
This gets nicer and nicer. Quite close by now.

rausnitz:
I ended up with build errors from https://github.com/vapor/postgres-nio/pull/330/commits/0dc332a85f32e0aae28a84a629f20da1b508d902 so I reverted in https://github.com/vapor/postgres-nio/pull/330/commits/1d3cf9fb824d30caaeda3486c984097c49e753f6.

Sources/PostgresNIO/Data/PostgresData+Range.swift
```diff
@@ -0,0 +1,130 @@
+import NIOCore
```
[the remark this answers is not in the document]
rausnitz: See new commit https://github.com/vapor/postgres-nio/pull/330/commits/1372252907a313fee9ab602821f572d1d3310aaf

fabianfett:
> I ended up with build errors from https://github.com/vapor/postgres-nio/commit/0dc332a85f32e0aae28a84a629f20da1b508d902 so I reverted in https://github.com/vapor/postgres-nio/commit/1d3cf9fb824d30caaeda3486c984097c49e753f6.

Where do you run into those issues? What are the errors?

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
```diff
@@ -0,0 +1,133 @@
...
+
+/// A type that can be decoded into a Swift RangeExpression type from a Postgres range where it is the bound type
+public protocol PostgresRangeDecodable: PostgresDecodable {
+    /// If a Postgres range type has a well-defined step,
+    /// Postgres automatically converts it to a canonical form.
+    /// Types such as `int4range` get converted to upper-bound-exclusive.
+    /// This method is needed when converting an upper bound to inclusive.
+    /// It should throw if the type lacks a well-defined step.
+    func stepDown() throws -> Self
```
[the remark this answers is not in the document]
rausnitz: Works for me. See new commit https://github.com/vapor/postgres-nio/pull/330/commits/7e8d4fb2489e8917cc4288f053cec8a16e1c6783.

We now have code that reads `upperBound.upperBoundExclusiveToUpperBoundInclusive()`, which feels a bit awkward, but it does the job.

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
```diff
@@ -0,0 +1,133 @@
...
+// MARK: Bound conformances
+
+extension Int32: PostgresRangeEncodable {
+    public static var psqlRangeType: PostgresDataType { return .int4Range }
+}
+
+extension Int32: PostgresRangeDecodable {
+    public func stepDown() -> Self {
+        return self - 1
+    }
```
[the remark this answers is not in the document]
rausnitz: Done in https://github.com/vapor/postgres-nio/commit/7e8d4fb2489e8917cc4288f053cec8a16e1c6783

rausnitz:
> Where do you run into those issues? What are the errors?

- `property 'boundType' is internal and cannot be referenced from an '@inlinable' function`
- `let '_isLowerBoundInclusive' is private and cannot be referenced from an '@inlinable' function`
- `let '_isUpperBoundInclusive' is private and cannot be referenced from an '@inlinable' function`
- `initializer 'init(lowerBound:upperBound:isLowerBoundInclusive:isUpperBoundInclusive:)' is internal and cannot be referenced from an '@inlinable' function`

To resolve them, I can make those properties and the initializer all `@usableFromInline`. I'd also make `_isLowerBoundInclusive` and `_isUpperBoundInclusive` internal, which probably means I should drop the `_` prefix (assuming that convention is for `private` properties).

Tests/IntegrationTests/PostgresNIOTests.swift
rausnitz: Yes that is much better. See new commit https://github.com/vapor/postgres-nio/pull/330/commits/26b59f03b0db71fabc822bc0ee33e8d20c09b00d.

Tests/IntegrationTests/PostgresNIOTests.swift
rausnitz: Added to `Tests/PostgresNIOTests/New/Data/Range+PSQLCodableTests.swift` in https://github.com/vapor/postgres-nio/pull/330/commits/26b59f03b0db71fabc822bc0ee33e8d20c09b00d.

rausnitz:
> To resolve them, I can make those properties and the initializer all `@usableFromInline`. I'd also make `_isLowerBoundInclusive` and `_isUpperBoundInclusive` internal, which probably means I should drop the `_` prefix (assuming that convention is for `private` properties).

I went ahead and did this in https://github.com/vapor/postgres-nio/pull/330/commits/8dc614690b13e1f1ca5ab7a902f2f57e4333bdb0, but not sure what to do with `let _isLowerBoundInclusive` and `let _isUpperBoundInclusive`. The code works, but now that they're not `private`, should I make them part of a type instead of making them global?

rausnitz:
I added a new commit https://github.com/vapor/postgres-nio/pull/330/commits/aecfdc14c6551a5704043f567962a0b54028987f to handle empty ranges, since Postgres treats them as a special case:
```
-- includes no points (and will be normalized to 'empty')
SELECT '[4,4)'::int4range;
```
Source: https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-IO.

This means the `PostgresDecodable` implementation won't find any bound values when parsing the binary. I think it's better to define a `defaultBoundValueForEmptyRange` than to throw an error.

fabianfett (commented):
Really polishing now!

Sources/PostgresNIO/New/Data/Array+PostgresCodable.swift
```diff
@@ -85,6 +85,18 @@ extension UUID: PostgresArrayEncodable {
     public static var psqlArrayType: PostgresDataType { .uuidArray }
 }
 
+extension Range: PostgresArrayDecodable where Bound: PostgresRangeDecodable {}
+
+extension Range: PostgresArrayEncodable where Bound: PostgresRangeEncodable {
+    public static var psqlArrayType: PostgresDataType { Bound.psqlRangeType.arrayType! }
+}
```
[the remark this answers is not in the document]
rausnitz: See new commit https://github.com/vapor/postgres-nio/pull/330/commits/71e10426d1c4140502054aa670be96874b02e645

Tests/IntegrationTests/PostgresNIOTests.swift
rausnitz: Added integration tests here https://github.com/vapor/postgres-nio/pull/330/commits/42174af06fcf7af5547dc08f990e288490fc0460.

I started to convert tests to async, but I wasn't sure how to handle the many instances of:
```
defer { XCTAssertNoThrow( try conn?.close().wait() ) }
```

If I use `await` instead of `wait()`, then I can't use `XCTAssertNoThrow`. If I don't use `XCTAssertNoThrow`, then the `defer` block won't build due to `errors cannot be thrown out of a defer body`. If I catch the error, I still get `'async' call cannot occur in a defer body`.

Any thoughts about how you'd like that handled?

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
```diff
@@ -0,0 +1,280 @@
...
+            lowerBound.encode(into: &byteBuffer, context: context)
+        }
+        if let upperBound: B = self.upperBound {
+            byteBuffer.writeInteger(boundMemorySize)
+            upperBound.encode(into: &byteBuffer, context: context)
+        }
+    }
+}
+
+extension PostgresRange where B: Comparable {
```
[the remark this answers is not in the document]
rausnitz: The two initializers in this extension require it.

The `Bound` value in `Range` and `ClosedRange` must be `Comparable`. Removing `where B: Comparable` here causes a build error.

See https://developer.apple.com/documentation/swift/range#declaration and https://developer.apple.com/documentation/swift/closedrange#declaration.

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
```diff
@@ -0,0 +1,280 @@
...
+    /// If a Postgres range type has a well-defined step,
+    /// Postgres automatically converts it to a canonical form.
+    /// Types such as `int4range` get converted to upper-bound-exclusive.
+    /// This method is needed when converting an upper bound to inclusive.
+    /// It should throw if the type lacks a well-defined step.
+    func upperBoundExclusiveToUpperBoundInclusive() throws -> Self
+
+    /// Postgres does not store any bound values for empty ranges,
+    /// but Swift requires a value to initialize an empty Range<Bound>.
+    static var defaultBoundValueForEmptyRange: Self { get }
```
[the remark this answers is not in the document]
rausnitz: How about `valueForDecodedEmptyRange` or `valueForEmptyRange`? The terms "default" and "fallback" imply that this value is only used if Postgres fails to provide a value, when we know Postgres will never provide a value.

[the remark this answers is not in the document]
fabianfett: I like `valueForEmptyRange`.

Tests/PostgresNIOTests/New/Data/Range+PSQLCodableTests.swift
```diff
@@ -0,0 +1,77 @@
...
+
+class Range_PSQLCodableTests: XCTestCase {
+    func testInt32RangeRoundTrip() {
+        let value: Range<Int32> = Int32.min..<Int32.max
+
+        var buffer = ByteBuffer()
+        value.encode(into: &buffer, context: .default)
+        XCTAssertEqual(Range<Int32>.psqlType, .int4Range)
+        XCTAssertEqual(buffer.readableBytes, 17)
+
```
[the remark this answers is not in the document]
rausnitz: See new commit https://github.com/vapor/postgres-nio/pull/330/commits/41c2846aae5197cf1ecfb39b54efe73c91996f9e. Is this what you wanted to see tests for?

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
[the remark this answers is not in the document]
rausnitz: Done in https://github.com/vapor/postgres-nio/pull/330/commits/884e8078ecfc8f6804387f2758e9f454598d3c15

fabianfett (commented):
Lovely. Can't believe how good this has already gotten. Sadly there is a bit more.

rausnitz:
> Lovely. Can't believe how good this has already gotten. Sadly there is a bit more.

I'm enjoying the work and welcome the thorough reviews. I'll have the next batch in soon.

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
```diff
@@ -0,0 +1,297 @@
...
+    /// It should throw if the type lacks a well-defined step.
+    func upperBoundExclusiveToUpperBoundInclusive() throws -> Self
+
+    /// Postgres does not store any bound values for empty ranges,
+    /// but Swift requires a value to initialize an empty Range<Bound>.
+    static var valueForEmptyRange: Self { get }
+}
+
+/// A type that can be encoded into a Postgres range array type where it is the bound type
+public protocol PostgresRangeArrayEncodable {
```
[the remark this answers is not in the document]
rausnitz: See new commit https://github.com/vapor/postgres-nio/pull/330/commits/e3c891fd9640b53b543a4caa2a28c5c40f6c256c

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
```diff
@@ -0,0 +1,297 @@
...
+    static var psqlRangeArrayType: PostgresDataType { get }
+}
+
```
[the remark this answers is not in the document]
rausnitz: Makes sense. Done in https://github.com/vapor/postgres-nio/pull/330/commits/e3c891fd9640b53b543a4caa2a28c5c40f6c256c.

Sources/PostgresNIO/New/Data/Array+PostgresCodable.swift
```diff
@@ -85,6 +85,18 @@ extension UUID: PostgresArrayEncodable {
     public static var psqlArrayType: PostgresDataType { .uuidArray }
 }
 
+extension Range: PostgresArrayDecodable where Bound: PostgresRangeDecodable {}
+
+extension Range: PostgresArrayEncodable where Bound: PostgresRangeEncodable & PostgresRangeArrayEncodable {
```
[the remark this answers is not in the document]
rausnitz: Done in https://github.com/vapor/postgres-nio/commit/e3c891fd9640b53b543a4caa2a28c5c40f6c256c

Sources/PostgresNIO/New/Data/Array+PostgresCodable.swift
```diff
@@ -85,6 +85,18 @@ extension UUID: PostgresArrayEncodable {
     public static var psqlArrayType: PostgresDataType { .uuidArray }
 }
 
+extension Range: PostgresArrayDecodable where Bound: PostgresRangeDecodable {}
```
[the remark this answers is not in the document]
rausnitz: Done in https://github.com/vapor/postgres-nio/commit/e3c891fd9640b53b543a4caa2a28c5c40f6c256c

Tests/IntegrationTests/PostgresNIOTests.swift
rausnitz: Thanks, done in https://github.com/vapor/postgres-nio/pull/330/commits/618a9d700c5b97ab247671e4467aabc41b3299b9.

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
```diff
@@ -0,0 +1,304 @@
...
+// MARK: PostgresRange
+
+@usableFromInline
+struct PostgresRange<B> {
+    @usableFromInline let lowerBound: B?
+    @usableFromInline let upperBound: B?
+    @usableFromInline let isLowerBoundInclusive: Bool
+    @usableFromInline let isUpperBoundInclusive: Bool
+
+    @usableFromInline
```
fabianfett, line 77:
```suggestion
    @inlinable
```

fabianfett (approved):
Lovely! We should fix the one `@usableFromInline` to `@inlinable`. However I would love if @gwynne could take a look as well.

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
```diff
@@ -0,0 +1,304 @@
...
+        }
+        
+        self = lowerBound...upperBound
+    }
+}
+
+// MARK: Private
+@usableFromInline let _isEmpty: UInt8 = 0x01
+@usableFromInline let _isLowerBoundInclusive: UInt8 = 0x02
+@usableFromInline let _isUpperBoundInclusive: UInt8 = 0x04
```
gwynne, line 304: Can these go in an enum or struct?

gwynne (changes requested):
This is excellent work! I had only the one concern, see below.

Sources/PostgresNIO/New/Data/Range+PostgresCodable.swift
rausnitz: See new commit https://github.com/vapor/postgres-nio/pull/330/commits/a6e1dd8666588a946553d4fabd3d41bacb10392d

gwynne (approved):

rausnitz:
@fabianfett Thanks for bearing with me through the iterations!

PR #331, opened 2023-03-09
Fix Decodable autoconformance to PostgresDecodable

jtouzy:
The default implementation of `PostgresDecodable` on `Decodable` types does not work automatically when used in other packages because it's not public. Just a simple visibility change does fix this.


fabianfett (approved):
Great finding. Thanks!

PR #302, opened 2022-07-10
Adding uds connection support to remove the need for deprecated socke…

MahdiBM:
Sorry for the noises, didn't know renaming a branch will close its PR.

This is duplicate of https://github.com/vapor/postgres-nio/pull/301.

### The Problem

Connecting through sockets has been deprecated in recent changes, which limits the ability to use uds for connections.

### The Solution

This PR will add the ability to use uds using the new apis.

### More

to not cause confusion: I rebased the commits into 1 since some changes from first commits were reversed.

PR #335, opened 2023-04-13
Add support for UDS and existing Channels

### LINKED ISSUE #259: Support establishing connections over existing NIO channels
It is useful to be able to ask MySQLNIO to initiate a connection using a preexisting NIO channel. The most obvious use case is SSH tunneling. This should mostly be a matter of simply exposing the appropriate API as public.

See also vapor/mysql-nio#76.

/cc @fabianfett

### LINKED ISSUE #268: The new connect syntax appears to be missing an option for unix domain sockets
The new connect syntax calls one of the following swift-nio connect functions:
`func connect(host: String, port: Int) -> EventLoopFuture<Channel>`
`func connect(to address: SocketAddress) -> EventLoopFuture<Channel>`

it dosen't call the third:
`func connect(unixDomainSocketPath: String) -> EventLoopFuture<Channel>`

gwynne:
Extends `PostgresConnection` to enable connecting to Postgres servers over UNIX domain sockets or using a preexisting `Channel`.

In the process, most of the existing structure of `PostgresConnection.Configuration` has been deprecated in favor of new APIs with more options and cleaner design. The deprecation messages provide instructions on migrating existing code. List of deprecations and their replacements (all listed APIs should be assumed to have a `PostgresConnection.` prefix):

- `Configuration.Connection` and `.connection` -> `Configuration.host`, `.port`, `.unixDomainSocket`, and `.establishedChannel` properties, and respective initializers.
- `Configuration.Authentication` and `.authentication` -> `Configuration.username`, `.password`, and `.database` properties.
- New `Configuration.Options` structure and `.options` property, containing the `connectTimeout` and `requireBackendKeyData` properties previously found on `Configuration.Connection`.

_Note:_ The integration tests for UNIX domain sockets will be skipped unless `POSTGRES_SOCKET` is set to a socket path in the test environment. See the updated `test.yml` for examples of how to do so correctly.

Closes #259
Closes #268

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
         shell: bash
         run: |
           coverage=$( [[ '${{ matrix.coverage }}' == 'true' ]] && echo -n '--enable-code-coverage' || true )
-          swift test --enable-test-discovery --filter=^PostgresNIOTests --sanitize=thread ${coverage}
+          swift test --filter=^PostgresNIOTests --sanitize=thread ${coverage}
       - name: Submit coverage report to Codecov.io
         if: ${{ matrix.coverage }}
         uses: vapor/swift-codecov-action@v0.2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,6 +58,7 @@ jobs:
             dbauth: trust
     container:
       image: swift:5.8-jammy
+      volumes: [ 'pgrunshare:/var/run/postgresql' ]
     runs-on: ubuntu-latest
     env:
       LOG_LEVEL: debug
@@ -74,10 +75,12 @@ jobs:
       POSTGRES_HOSTNAME: 'psql-a'
       POSTGRES_HOSTNAME_A: 'psql-a'
       POSTGRES_HOSTNAME_B: 'psql-b'
+      POSTGRES_SOCKET: '/var/run/postgresql/.s.PGSQL.5432'
       POSTGRES_HOST_AUTH_METHOD: ${{ matrix.dbauth }}
     services:
       psql-a:
         image: ${{ matrix.dbimage }}
+        volumes: [ 'pgrunshare:/var/run/postgresql' ]
         env:
           POSTGRES_USER: 'test_username'
           POSTGRES_DB: 'test_database'
@@ -86,6 +89,7 @@ jobs:
           POSTGRES_INITDB_ARGS: --auth-host=${{ matrix.dbauth }}
       psql-b:
         image: ${{ matrix.dbimage }}
+        volumes: [ 'pgrunshare:/var/run/postgresql' ]
         env:
           POSTGRES_USER: 'test_username'
           POSTGRES_DB: 'test_database'
@@ -134,6 +138,7 @@ jobs:
       POSTGRES_PASSWORD: 'test_password'
       POSTGRES_DB: 'postgres'
       POSTGRES_HOST_AUTH_METHOD: ${{ matrix.dbauth }}
+      POSTGRES_SOCKET: '/tmp/.s.PGSQL.5432'
     steps:
       - name: Select latest available Xcode
         uses: maxim-lobanov/setup-xcode@v1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,14 @@ jobs:
           - swift:5.8-jammy
           - swiftlang/swift:nightly-5.9-jammy
           - swiftlang/swift:nightly-main-jammy
+        include:
+          - coverage: true
+          # https://github.com/apple/swift-package-manager/issues/5853
+          - container: swift:5.8-jammy
+            coverage: false
+          # https://github.com/apple/swift/issues/65064
+          - container: swiftlang/swift:nightly-main-jammy
+            coverage: false
     container: ${{ matrix.container }}
     runs-on: ubuntu-latest
     env:
@@ -29,9 +37,12 @@ jobs:
       - name: Check out package
         uses: actions/checkout@v3
       - name: Run unit tests with code coverage and Thread Sanitizer
-        run: swift test --enable-test-discovery --filter=^PostgresNIOTests --sanitize=thread --enable-code-coverage
+        shell: bash
+        run: |
+          coverage=$( [[ '${{ matrix.coverage }}' == 'true' ]] && echo -n '--enable-code-coverage' || true )
+          swift test --enable-test-discovery --filter=^PostgresNIOTests --sanitize=thread ${coverage}
       - name: Submit coverage report to Codecov.io
-        if: ${{ !contains(matrix.container, '5.8') }}
+        if: ${{ matrix.coverage }}
         uses: vapor/swift-codecov-action@v0.2
         with:
           cc_flags: 'unittests'
--- a/Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
@@ -1,5 +1,5 @@
 import NIOCore
-@_implementationOnly import NIOPosix // inet_pton() et al.
+import NIOPosix // inet_pton() et al.
 import NIOSSL
 
 extension PostgresConnection {
@@ -51,12 +51,12 @@ extension PostgresConnection {
 
             // MARK: Implementation details
             
-            fileprivate enum Base {
+            enum Base {
                 case disable
                 case prefer(NIOSSLContext)
                 case require(NIOSSLContext)
             }
-            fileprivate let base: Base
+            let base: Base
             private init(base: Base) { self.base = base }
         }
         
@@ -70,23 +70,20 @@ extension PostgresConnection {
             /// ``PostgresConnection/Configuration/init(establishedChannel:username:password:database:)``.)
             public var connectTimeout: TimeAmount
             
-            /// The name to use for SNI (Server Name Indication) when TLS is enabled. Defaults to none (but see below).
+            /// The server name to use for certificate validation and SNI (Server Name Indication) when TLS is enabled.
+            /// Defaults to none (but see below).
             ///
             /// > When set to `nil`:
             /// If the connection is made to a server over TCP using
             /// ``PostgresConnection/Configuration/init(host:port:username:password:database:tls:)``, the given `host`
             /// is used, unless it was an IP address string. If it _was_ an IP, or the connection is made by any other
             /// method, SNI is disabled.
-            ///
-            /// - Warning: This name is not validated in any way; in particular, there is no attempt to check that it
-            ///   matches the server's TLS certificate, nor is it in any way impacted by the validation setting in the
-            ///   TLS configuration (see ``PostgresConnection/Configuration/TLS-swift.struct``).
             public var tlsServerName: String?
             
             /// Whether the connection is required to provide backend key data (internal Postgres stuff).
             ///
             /// This property is provided for compatibility with Amazon RDS Proxy, which requires it to be `false`.
-            /// If you are not using Amazon RDS Proxy, you should probably leave this set to `true` (the default).
+            /// If you are not using Amazon RDS Proxy, you should leave this set to `true` (the default).
             public var requireBackendKeyData: Bool
             
             /// Create an options structure with default values.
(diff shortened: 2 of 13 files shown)


gwynne:
This PR is intended to subsume both #302 and #314, lmk if there's anything missing.

0xTim:
Resolves #259

gwynne:
CI failure on `nightly-main` is tracked by https://github.com/apple/swift/issues/65064

gwynne:
Also resolves #268

gwynne:
Resolves #268

.github/workflows/test.yml
```diff
@@ -27,7 +31,7 @@ jobs:
       - name: Run unit tests with code coverage and Thread Sanitizer
         run: swift test --enable-test-discovery --filter=^PostgresNIOTests --sanitize=thread --enable-code-coverage
       - name: Submit coverage report to Codecov.io
-        if: "!contains(matrix.container, 'nightly')"
+        if: ${{ !contains(matrix.container, '5.8') }}
```
fabianfett, line 34: why did we change this to `5.8`?

MahdiBM: The reason we used 'nightly' was that Codecov seemed to have problems with nightly builds. 
We can either try to see if Codecov now works fine with nightly builds, or let it stay disabled to make sure it doesn't fail the CI. 
I think Gwynne thought now that 5.8 is released, we can change that to '5.8', but if i were to guess i would say the problems we were having were nightly related, not 5.8's fault.

Sources/PostgresNIO/New/Data/UUID+PostgresCodable.swift
```diff
@@ -1,4 +1,5 @@
 import NIOCore
+import NIOFoundationCompat
```
fabianfett, line 2: why did that happen?

Tests/IntegrationTests/PostgresNIOTests.swift
```diff
@@ -31,6 +31,25 @@ final class PostgresNIOTests: XCTestCase {
...
+    
+    func testConnectUDSAndClose() throws {
+        try XCTSkipUnless(env("POSTGRES_SOCKET") != nil)
+        var conn: PostgresConnection?
+        XCTAssertNoThrow(conn = try PostgresConnection.testUDS(on: eventLoop).wait())
+        XCTAssertNoThrow(try conn?.close().wait())
+    }
+    
+    func testConnectEstablishedChannelAndClose() throws {
+        let channel = try ClientBootstrap(group: self.group).connect(to: PostgresConnection.address()).wait()
```
fabianfett, line 43: do we need
```
        try XCTSkipUnless(env("POSTGRES_SOCKET") != nil)
```
here as well?

.github/workflows/test.yml
MahdiBM: I see Gwynne is tracking the Codecov issue already, based on her comment above. Then i guess the options are to either remove that 'if' or change it back.

0xTim (approved):
LGTM - only thing to mention that I couldn't quite tell is that it looks like all the tests have been updated to use the new APIs. We should probably have a test just to ensure the deprecated stuff still works

MahdiBM (approved):

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -68,31 +68,209 @@ public final class PostgresConnection {
...
             ///
-            /// - Default: 5432
-            public var port: Int
+            /// - Parameters:
+            ///   - channel: The ``NIOCore/Channel`` to use. The channel must already be active and
+            ///     connected to an endpoint.
+            ///   - tlsHostname: See ``tlsHostname``.
+            ///   - requireBackendKeyData: See ``requireBackendKeyData``.
+            public static func establishedChannel(
+                channel: Channel,
```
fabianfett, line 143: if we use `establishedChannel` then we should drop the `channel` parameter name

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -269,8 +447,16 @@ public final class PostgresConnection {
...
-            case .unresolved(let host, let port):
+            case .unresolvedTCP(let host, let port):
                 connectFuture = bootstrap.connect(host: host, port: port)
+            case .unresolvedUDS(let path, _):
+                connectFuture = bootstrap.connect(unixDomainSocketPath: path)
+            case .bootstrapped(let channel, _):
+                guard channel.isActive else {
+                    return eventLoop.makeFailedFuture(PSQLError.channel(underlying: ChannelError.alreadyClosed))
+                }
+                // TODO: Are there drawbacks to creating a bootstrap we don't end up using?
```
fabianfett, line 458: I think we are fine here. The bootstrap is minimal anyway

.github/workflows/test.yml
gwynne: The issue was that 5.8 changed the behavior of the `--show-codecov-path` flag. This was fixed, but the fix didn't make it in until 5.9. See https://github.com/apple/swift-package-manager/issues/5853 for details.

Sources/PostgresNIO/New/Data/UUID+PostgresCodable.swift
gwynne: It's required for `ByteBuffer.writeUUIDBytes(_:)` and `.readUUIDBytes()`. The need for the import becomes explicit when building in `BUILDING_DOCC` mode (which basically removes `Exports.swift` and thus all uses of `@_exported`). This was our workaround for the issues DocC has with re-exported imports - `@_documentation(visibility:)` didn't come along until later.

Tests/IntegrationTests/PostgresNIOTests.swift
gwynne: That's only for the UDS tests; for the established channel test I just mimic the channel bootstrap part of opening a TCP connection.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -68,31 +68,209 @@ public final class PostgresConnection {
...
+            /// - Parameters:
+            ///   - host: The hostname to connect to.
+            ///   - port: The TCP port to connect to (defaults to 5432).
+            ///   - connectTimeout: See ``connectTimeout``.
+            ///   - requireBackendKeyData: See ``requireBackendKeyData``.
+            public static func tcp(
+                host: String,
+                port: Int = 5432,
+                connectTimeout: TimeAmount = .seconds(10),
+                requireBackendKeyData: Bool = true
```
fabianfett, line 102: as discussed yesterday, we should drop the tcp prefix where possible.

fabianfett (changes requested):

Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
```diff
@@ -0,0 +1,318 @@
...
+            }
+            let base: Base
+            init(base: Base, tls: TLS) { (self.base, self.tls) = (base, tls) }
+        }
+        
+        // MARK: - Authentication
+
+        /// Contains authentication information for a connection.
+        public struct Authentication {
+            // TODO: Make the properties of this structure immutable (removing setters breaks public API).
```
fabianfett, line 159: Why do we want to do that?

Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
```diff
@@ -0,0 +1,318 @@
...
+            /// - Warning: This name is not validated in any way; in particular, there is no attempt to check that it
+            ///   matches the server's TLS certificate, nor is it in any way impacted by the validation setting in the
+            ///   TLS configuration (see ``PostgresConnection/Configuration/Server-swift.struct/TLS-swift.struct``).
+            public let tlsServerName: String?
+            
+            /// Whether the connection is required to provide backend key data (internal Postgres stuff).
+            ///
+            /// This property is provided for compatibility with Amazon RDS Proxy, which requires it to be `false`.
+            /// If you are not using Amazon RDS Proxy, you should probably leave this set to `true` (the default).
+            public let requireBackendKeyData: Bool
```
fabianfett, line 209: can we make the properties mutable?

Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
```diff
@@ -0,0 +1,318 @@
...
+            ///   - requireBackendKeyData: See ``requireBackendKeyData``. Default is `true`.
+            public init(
+                connectTimeout: TimeAmount = .seconds(10),
+                tlsServerName: String? = nil,
+                requireBackendKeyData: Bool = true
+            ) {
+                self.connectTimeout = connectTimeout
+                self.tlsServerName = tlsServerName
+                self.requireBackendKeyData = requireBackendKeyData
+            }
```
fabianfett, line 227: can we have just an init() that sets the default values? If users want to change this they can use the properties. This makes adding more properties to this way simpler in the future.

Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
```diff
@@ -0,0 +1,318 @@
...
+        ///   - options: See ``Options-swift.struct``. Most users do not need to adjust any of these.
+        public init(
+            server: Server,
+            authentication: Authentication,
+            options: Options = .init()
+        ) {
+            self.server = server
+            self.authentication = authentication
+            self.options = options
+        }
```
fabianfett, line 259: In an ideal case this is just an init and users do all changes through setting properties. See TLSConfiguration for comparison.

fabianfett (commented):
Looks way better.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -170,22 +68,15 @@ public final class PostgresConnection: @unchecked Sendable {
...
-                )
-                try channel.pipeline.syncOperations.addHandler(sslHandler, position: .first)
-            }
-        }
+            let sslHandler = try NIOSSLClientHandler(
+                context: sslContext,
+                serverHostname: configuration.serverNameForSNI
+            )
+            try channel.pipeline.syncOperations.addHandler(sslHandler, position: .first)
+        } }
```
fabianfett, line 79: can we keep the impl here as is? I actually like that we use the internal enum here. This will make us think about it, if we add another case in the future. Also I find the create closure inside a closure extremely hard to read.

Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
```diff
@@ -0,0 +1,280 @@
...
+        case .configureChannel(let channel): self.connection = .bootstrapped(channel: channel)
+        }
+        self.username = config.username
+        self.password = config.password
+        self.database = config.database
+        self.tls = config.tls
+        self.options = config.options
+    }
+    
+    var serverNameForSNI: String? {
```
fabianfett, line 257: can we rename this to `serverNameForTLS`? Since it also used for certificate validation?

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -1,11 +1,11 @@
 import Atomics
 import NIOCore
+@_implementationOnly import NIOPosix
```
fabianfett, line 3: why did you add `@_implementationOnly`?

Tests/PostgresNIOTests/New/PostgresChannelHandlerTests.swift
```diff
@@ -36,9 +36,8 @@ class PostgresChannelHandlerTests: XCTestCase {
         XCTAssertNoThrow(try embedded.writeInbound(PostgresBackendMessage.readyForQuery(.idle)))
     }
     
-    func testEstablishSSLCallbackIsCalledIfSSLIsSupported() {
-        var config = self.testConnectionConfiguration()
-        XCTAssertNoThrow(config.tls = .require(try NIOSSLContext(configuration: .makeClientConfiguration())))
+    func testEstablishSSLCallbackIsCalledIfSSLIsSupported() throws {
+        let config = self.testConnectionConfiguration(tls: .require(try NIOSSLContext(configuration: .makeClientConfiguration())))
```
fabianfett, line 40: can we keep the nothrow behavior here?

Tests/PostgresNIOTests/New/PostgresChannelHandlerTests.swift
```diff
@@ -50,15 +49,15 @@ class PostgresChannelHandlerTests: XCTestCase {
         ])
         
         var maybeMessage: PostgresFrontendMessage?
-        XCTAssertNoThrow(embedded.connect(to: try .init(ipAddress: "0.0.0.0", port: 5432), promise: nil))
-        XCTAssertNoThrow(maybeMessage = try embedded.readOutbound(as: PostgresFrontendMessage.self))
+        embedded.connect(to: try .init(ipAddress: "0.0.0.0", port: 5432), promise: nil)
+        maybeMessage = try embedded.readOutbound(as: PostgresFrontendMessage.self)
```
fabianfett, line 53: keep nothrow behavior

fabianfett (commented):
LOVELY! Nits

fabianfett (approved):
Lovely! Thanks so much!

PR #337, opened 2023-04-14
Update CI in preparation for bumping to 5.6 min version

gwynne:


fabianfett (approved):

.github/workflows/test.yml
```diff
@@ -52,7 +56,8 @@ jobs:
             dbauth: md5
           - dbimage: postgres:11
             dbauth: trust
-    container: swift:5.7-jammy
+    container:
+      image: swift:5.8-jammy
```
0xTim, line 60: This is inconsistent with the other instances of `container` in the workflow but ¯\_(ツ)_/¯

0xTim (approved):

.github/workflows/test.yml
gwynne: That's 'cause I cherry-picked from #335, where this is changed to be able to specify a volume used for testing the UDS support. It didn't seem worth the extra effort when it'll be moot in another PR or two anyway.

PR #342, opened 2023-04-19
Make PSQLError public

fabianfett:
### Motivation

Users want to inspect what has gone wrong. For this they need access to the error type that we throw.

### Changes

- Rewrite `PSQLError` to ensure it is a type that we can evolve in public
  - For this to work we shouldn't use associated types on the code enum
- Merge `channel` and `connectionError` code into one
- Rename `PSQLDecodingError` to `PostgresMessageDecodingError` to ensure a clear distinction to `PostgresDecodingError`
- `PSQLError` is now a CoW type
- Attach the query, if we have one to the PSQLError to make debugging easier

### Result

- Better debuggable errors.

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -283,7 +283,7 @@ public final class PostgresConnection: @unchecked Sendable {
                 case is PSQLError:
                     throw error
                 default:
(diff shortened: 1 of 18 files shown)


gwynne (approved):
Looks fine, but one nit: If you missed any other uses of `#file` (I dunno if you did, I didn't look), might as well change the rest to `#fileID` too (or `#filePath` if used within XCTest stuff) so there isn't any inconsistency.

PR #351, opened 2023-04-27
Fixes crash in queries that timeout

fabianfett:
Fixes #347.

In an ideal case I add a unit test for this as well, but sadly this is rather involved.

cc @trasch.

Issue #347, opened 2023-04-21
Crash when using 'SET statement_timeout=x'


trasch:
### Describe the bug

I want to prevent rogue queries from hogging up my database connections and therefore use
`SET statement_timeout=10000;` 
to let the server kill long running queries.

This leads to a crash in `ConnectionStateMachine.swift#845 ` with the message `Tried to cancel stream without active query` whenever PostgreSQL kills a query.

This is the stacktrace:
```
#0  in ConnectionStateMachine.cancelQueryStream() at /postgres-nio/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift:845
#1  in PostgresChannelHandler.cancel(for:) at /postgres-nio/Sources/PostgresNIO/New/PostgresChannelHandler.swift:515
#2  in protocol witness for PSQLRowsDataSource.cancel(for:) in conformance PostgresChannelHandler ()
#3  in PSQLRowStream.cancel0() at /postgres-nio/Sources/PostgresNIO/New/PSQLRowStream.swift:144
#4  in PSQLRowStream.cancel() at /postgres-nio/Sources/PostgresNIO/New/PSQLRowStream.swift:132
#5  in PSQLRowStream.didTerminate() at /postgres-nio/Sources/PostgresNIO/New/PSQLRowStream.swift:425
#6  in protocol witness for NIOAsyncSequenceProducerDelegate.didTerminate() in conformance PSQLRowStream ()
#7  in NIOThrowingAsyncSequenceProducer.Storage.finish(_:) at /swift-nio/Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift:476
#8  in NIOThrowingAsyncSequenceProducer.Source.finish(_:) at /swift-nio/Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift:340
#9  in PSQLRowStream.receiveError(_:) at /postgres-nio/Sources/PostgresNIO/New/PSQLRowStream.swift:386
#10 in PSQLRowStream.receive(completion:) at /postgres-nio/Sources/PostgresNIO/New/PSQLRowStream.swift:337
#11 in PostgresChannelHandler.run(_:with:) at /postgres-nio/Sources/PostgresNIO/New/PostgresChannelHandler.swift:276
#12 in closure #1 in PostgresChannelHandler.channelRead(context:data:) at /postgres-nio/Sources/PostgresNIO/New/PostgresChannelHandler.swift:147
[...]

PR #352, opened 2023-04-27
Follow up to fixing crash in queries that timeout

gwynne:
This hopefully solves the failed force-unwrap crash in the fix for #347 provided by #351.

cc @fabianfett @trasch

### DIFF
--- a/Sources/PostgresNIO/New/PostgresChannelHandler.swift
+++ b/Sources/PostgresNIO/New/PostgresChannelHandler.swift
@@ -262,7 +262,7 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
             
         case .forwardStreamComplete(let buffer, let commandTag):
             guard let rowStream = self.rowStream else {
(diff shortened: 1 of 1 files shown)


trasch:
I can confirm that this fixes the remaining crash for me 👍

Thanks @gwynne and @fabianfett !

fabianfett:
@trasch Do you have a repro for the crash?

gwynne:
It fixes the proximate crash, but it's not really dealing with the actual underlying problem; I defer to @fabianfett for the real solution.

trasch:
>  Do you have a repro for the crash?

@fabianfett Not publicly available, and it uses internal databases that aren't available externally.

What I'm doing is running a few million PostGIS queries with a connection pool of size 50, so lots of parallel queries with short timeouts.
I will try to create a test case that mimics this, but I will need some time. (And I hope that I didn't simply botch the connection pool...)

PR #356, opened 2023-04-28
Add good PSQLError debugDescription

### LINKED ISSUE #354: Make the PSQLError description more useful
... by making it conform to `CustomStringConvertible` and `CustomDebugStringConvertible`.

Currently, `print(error as! PSQLError)` will print something like this:
`PSQLError(backing: PostgresNIO.PSQLError.(unknown context at $100aeb99c).Backing)`
which is not very helpful.

I would propose to add a description to PSQLError that uses some of the infos provided in `.code`, `.serverInfo` and other properties to convey a more helpfull message.

And I can tackle this myself, unless someone else wants to do it... 😄

fabianfett:
This shall be the start of our discussion.

Will eventually fix #354.

### DIFF
--- a/Sources/PostgresNIO/Deprecated/PostgresData+UInt.swift
+++ b/Sources/PostgresNIO/Deprecated/PostgresData+UInt.swift
@@ -1,6 +1,6 @@
 private func warn(
     _ old: Any.Type, mustBeConvertedTo new: Any.Type,
-    file: StaticString = #filePath, line: UInt = #line
+    file: StaticString = #file, line: UInt = #line
 ) {
     assertionFailure("""
     Integer conversion unsafe.


trasch:
My thoughts about the `serverInfo` part:
- Wrapping the `message` in quotes would make it look nicer. Probably also some of the other fields as `detail`, `hint` and so on (and also `backendMessage` and `query` on the top level).
- Not sure if `routine`, `file` and `line` from Postgres are really necessary. I personally have never gained any useful information from that, so IMO we only display it because we have it...
- Is `localizedSeverity` really necessary in a debug description that's intended only for developers? And it's somewhat redundant with `severity`...

Apart from the nitpicking: Thanks @fabianfett , this is already a big improvement!

fabianfett:
What do you think about going through debugDescription to prevent users from accidentally leaking?

trasch:
@fabianfett Thinking about it a bit longer, I'm not so convinced anymore about the security problem. I mean, there are of course developers who have no clue about security, and oversights happen to the best of us, but is it really the job of a library to prevent this?

I think that `description` should be a short error description, like in the Postgres logs:
`ERROR: canceling statement due to statement timeout`
and `debugDescription` should be the full monty, as it is now.

But maybe there are other users of this library that have yet another opinion? 😃

PR #358, opened 2023-05-02
Add PostgresDecodingError debugDescription

gwynne:
Follows up on #356


Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -476,3 +476,23 @@ extension PostgresDecodingError: CustomStringConvertible {
         "Database error"
     }
 }
+
+extension PostgresDecodingError: CustomDebugStringConvertible {
+    public var debugDescription: String {
```
fabianfett, line 481: Can we add a test for this one? I really want us to improve on this :)

fabianfett (approved):

PR #360, opened 2023-05-03
Add PSQLError debugDescription

fabianfett:
Still needs tests.

Also continues work from #356.


gwynne:
Closing in favor of continuing work in #372

PR #368, opened 2023-05-11
Decode `.bpchar` as `String`

gwynne:
`bpchar` is "blank-padded char", the low-level Postgres name for `character(N)` (the auto-padded form of `character varying`). `String`'s `PostgresCodable` conformance should thus recognize it as a valid representation.


michal-tomlein:
Any chance this could be released soon? We would like to get the CVE fix, but this broke our app. Thanks!

gwynne:
@michal-tomlein We just released 1.15.0, which includes this fix, thanks for the nudge! ❤️

PR #372, opened 2023-06-02
Add PSQLError debugDescription

### LINKED ISSUE #354: Make the PSQLError description more useful

gwynne:
Fixes #354, vapor/fluent-postgres-driver#213. Continues work from #356 and #360.

Requires minor version bump (adds new public API in the form of protocol conformances).


Tests/IntegrationTests/AsyncTests.swift
```diff
@@ -106,7 +106,7 @@ final class AsyncPostgresConnectionTests: XCTestCase {
             } catch {
                 guard let error = error as? PSQLError else { return XCTFail("Unexpected error type") }
 
-                print(error)
+                print(String(reflecting: error))
```
fabianfett, line 109: I know this is a leftover from me. Can we remove this dude?

fabianfett (commented):
Looks great. Now we need some tests!

Sources/PostgresNIO/New/PostgresQuery.swift
```diff
@@ -122,6 +136,16 @@ public struct PostgresBindings: Sendable, Hashable {
...
+        
+        @usableFromInline
+        var description: String {
+            "\(self.dataType)"
+        }
+        
+        @usableFromInline
+        var debugDescription: String {
+            "\(self.format == .text ? "text" : "bin")(\(self.dataType))"
+        }
```
fabianfett, line 148: I don't think we should add this conformance here. Instead we should unwrap directly in `PostgresBindings`

Sources/PostgresNIO/New/PostgresQuery.swift
```diff
@@ -179,3 +203,22 @@ public struct PostgresBindings: Sendable, Hashable {
...
+
+extension PostgresBindings: CustomDebugStringConvertible {
+    /// See ``Swift/CustomDebugStringConvertible/debugDescription``.
+    public var debugDescription: String {
+        """
+        PostgresBindings(\
+        metadata: [\(self.metadata.map { String(reflecting: $0) }.joined(separator: "; "))], \
+        bytes: \(self.bytes.readableBytes) [\(Array(self.bytes.readableBytesView).prefix(8).map(\.description).joined(separator: ","))\(self.bytes.readableBytes > 8 ?",..." : "")]\
+        )
+        """
```
fabianfett, line 222: I think this leaks an internal impl. detail that the user doesn't have to be aware of...

I like the array representation in `description`. Can we go for something like this:

```
[(type: uuid, format: bin, 16 bytes), (type: int4, format: bin, 4 bytes)]
```

Later we could present the actual values, if we wanted to. This could make debugging a lot easier!
```
[(type: uuid, format: bin, value: "45223d46-c8ca-4c90-9ad1-8a829db6a7ba"), (int4, bin, value: 12)]
```

fabianfett (commented):
Thanks for the progress on this. A few comments

Sources/PostgresNIO/New/PostgresQuery.swift
gwynne: Both times we'd have to actually parse out the buffer just to generate the description. That's a lot of work.

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -490,7 +583,11 @@ extension PostgresDecodingError: CustomStringConvertible {
         // This may seem very odd... But we are afraid that users might accidentally send the
         // unfiltered errors out to end-users. This may leak security relevant information. For this
         // reason we overwrite the error description by default to this generic "Database error"
-        "Database error"
+        """
+        PostgresDecodingError – This error description is generic on purpose to prevent leakage of
+        impl. details to upstream services. For debugging details use String(reflecting: error) on
+        PostgresDecodingError.
+        """
```
gwynne, line 590: I don't love a multiline message here, that's liable to make quite a muddle of peoples' logs.
```suggestion
        """
        PostgresDecodingError – Generic description to prevent accidental leakage of sensitive data. For details, use `String(reflecting: error)`.
        """
```
I know, it's longer on the one line, but it's barely 1.5x as long as one line of yours.

fabianfett (approved):

PR #375, opened 2023-06-21
Feature/typed queries

marius-se:
WIP cc @fabianfett


fabianfett:
This should use #390 once it has landed.

ptoffy:
I'll close this for now as it is stale but maybe we can revisit this in the future

PR #376, opened 2023-07-20
Use computed property to PostgresConnection.Configuration.TLS.disable for concurrency safe

sidepelican:
`PostgresConnection.Configuration.TLS.disable` is declared as `var` but it causes compiler warning on strict concurrency checking.

cocurrency error

This value seems to not to be modified in runtime.
So this should be changed to `let`.

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
@@ -13,7 +13,7 @@ extension PostgresConnection {
             // MARK: Initializers
             
             /// Do not try to create a TLS connection to the server.
-            public static let disable: Self = .init(base: .disable)
+            public static var disable: Self { .init(base: .disable) }
 
             /// Try to create a TLS connection to the server. If the server supports TLS, create a TLS connection.
             /// If the server does not support TLS, create an insecure connection.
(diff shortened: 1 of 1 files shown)


Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
```diff
@@ -13,7 +13,7 @@ extension PostgresConnection {
             // MARK: Initializers
             
             /// Do not try to create a TLS connection to the server.
-            public static var disable: Self = .init(base: .disable)
+            public static let disable: Self = .init(base: .disable)
```
gwynne, line 16: Good catch!, But rather than making it a `let`, which also has concurrency implications due to lazy initialization, let's make it a stored property:

```suggestion
            public static var disable: Self { .init(base: .disable) }
```

gwynne (dismissed):

MahdiBM:
This is a breaking change but it does seem to me that the property was supposed to be a computed property not stored.

I personally don't mind letting it slide although it's a breaking change, because it isn't supposed to be used like that anyway.

MahdiBM:
@gwynne beat me to it while i was typing on phone 😅

gwynne:
It is technically a breaking change, but in this case the likelihood of someone having an actual source-compatibility issue is very low, and would have required doing something that was fairly obviously unintended.

gwynne (approved):
Thanks for the fix!

fabianfett (approved):
I'm aware that this is technically a breaking change. However, if someone actually used the set on this property, they are likely in much more trouble. For this reason, I'm fine with breaking API in a non major release.

PR #377, opened 2023-07-27
Add support for decimal types other than Foundation.Decimal

ratranqu:
Currently, the only numerical option for an exact representation of the Postgres NUMERIC type is to use `Foundation.Decimal`. However, `Foundation.Decimal` is not fully implemented, and in particular is missing some useful conformances such as `LosslessStringConvertible` or `FloatingPoint`. The current implementation makes it complex to use other Decimal implementations to map to the NUMERIC type.

This PR aims to provide a new protocol `ExpressibleByFloatingPointString`, to which `Foundation.Decimal` can trivially conform, and at the same time allow to easily make other Decimal types conform to (*), and be used as mapping to the Postgres NUMERIC type.

```swift
extension Decimal: ExpressibleByPostgresFloatingPointString {
    public init?(floatingPointString: String) {
        self.init(string: floatingPointString)
    }
}
```

The change aims to make no changes to the API. It has been tested with the Test suite from postgres-nio and that of postgres-kit as well as from use in a higher level Vapor app.

--
(*) for example `BigDecimal` from https://github.com/mgriebling/BigDecimal.git

```swift
import PostgresNIO
import BigDecimal

extension BigDecimal: ExpressibleByPostgresFloatingPointString {
    public init?(floatingPointString: String) {
        self.init(floatingPointString)
    } 
}
```

### DIFF
--- a/Sources/PostgresNIO/Data/PostgresData+Decimal.swift
+++ b/Sources/PostgresNIO/Data/PostgresData+Decimal.swift
@@ -30,6 +30,6 @@ extension Decimal: PostgresDataConvertible {
     }
 
     public var postgresData: PostgresData? {
-        return .init(numeric: PostgresNumeric(decimal: self))
+        return .init(numeric: PostgresNumeric(decimalString: self.description))
     }
 }
--- a/Sources/PostgresNIO/Data/PostgresData+Numeric.swift
+++ b/Sources/PostgresNIO/Data/PostgresData+Numeric.swift
@@ -1,5 +1,4 @@
 import NIOCore
-import struct Foundation.Decimal
 
 public struct PostgresNumeric: CustomStringConvertible, CustomDebugStringConvertible, ExpressibleByStringLiteral {
     /// The number of digits after this metadata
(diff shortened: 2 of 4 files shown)


gwynne:
It feels to me like this could be accomplished without a new protocol by making `PostgresNumeric` (which is already public) directly `PostgresCodable`; non-Foundation decimal types would then be able to more easily conform to `PostgresCodable` themselves using it. But I may be missing something. @fabianfett, any thoughts?

ratranqu:
> It feels to me like this could be accomplished without a new protocol by making `PostgresNumeric` (which is already public) directly `PostgresCodable`; non-Foundation decimal types would then be able to more easily conform to `PostgresCodable` themselves using it. But I could easily be missing something. @fabianfett, any thoughts?

My gut feel is that as long as `PostgresNumeric` (an exact type) allows itself to be represented on the Swift side by `Float`s or `Double`s (non exact), it will be difficult to allow for that since `Float`/`Double` have their own `PostgresCodable` specificities?

Issue #374, opened 2023-06-07
Attempted re-use of already closed connection


GNMoseke:
### Describe the bug
Bug has been previously described [here](https://forums.swift.org/t/postgresnio-precondition-failure-how-can-an-error-occur-if-the-connection-is-already-closed/57075)

We are seeing a crasher from `PostgresNIO.ConnectionStateMachine.closeAndCleanup(_:)` where postgres-nio attempts to re-use an already closed connection. The failure happens about 80% of the time in a given unit test, and seems to stem from a quick sequence of Failed Insert -> Select -> Delete requests. 

### To Reproduce
The reproducer linked in the swift forum post above is a good place to start. I've also gathered logs from the (rare) success case and much more frequent case below:

#### Success Logs
```
**2023-06-05T10:27:20.488-06:00 [TRACE] StorageInterface PostgresChannelHandler.swift#L.103 ["psql_message": .error(PostgresNIO.PostgresBackendMessage.ErrorResponse(fields: [Code: "23503", Schema name: "public", Localized Severity: "ERROR", Line: "2465", Detail: "... violates foreign key constraint \"...\"", Constraint name: "...", Severity: "ERROR", File: "ri_triggers.c", Routine: "ri_ReportViolation", Table name: "..."])), "psql_connection_id": 3] Backend message received**

**2023-06-05T10:27:20.490-06:00 [TRACE] StorageInterface PostgresChannelHandler.swift#L.217 ["psql_connection_action": closeConnectionAndCleanup(PostgresNIO.ConnectionStateMachine.ConnectionAction.CleanUpContext(action: PostgresNIO.ConnectionStateMachine.ConnectionAction.CleanUpContext.Action.close, tasks: [PostgresNIO.PSQLTask.extendedQuery(PostgresNIO.ExtendedQueryContext)], error: PostgresNIO.PSQLError(backing: PostgresNIO.PSQLError.(unknown context at $129798b04).Backing), closePromise: nil)), "psql_connection_id": 3] Run action**

**2023-06-05T10:27:20.490-06:00 [DEBUG] StorageInterface PostgresChannelHandler.swift#L.480 ["psql_error": PSQLError(backing: PostgresNIO.PSQLError.(unknown context at $129798b04).Backing), "psql_connection_id": 3] Cleaning up and closing connection.**

**2023-06-05T10:27:20.490-06:00 [TRACE] StorageInterface EventLoopConnectionPool.swift#L.261  Releasing connection**

[...]

PR #379, opened 2023-08-03
Crash fix: Multiple bad messages could trigger reentrancy issue

fabianfett:
If we receive multiple unexpected messages from the backend we can run into a reentrency situation in which we still have unread messages in the incomming buffer after we have received `channelInactive`. This pr patches this crash.

Fixes the crash mentioned in #374.

### DIFF
--- a/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
+++ b/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
@@ -928,7 +928,7 @@ struct ConnectionStateMachine {
                  .forwardStreamComplete,
                  .wait,
                  .read:
-                preconditionFailure("Expecting only failure actions if an error happened")
+                preconditionFailure("Invalid state: \(self.state)")
             case .evaluateErrorAtConnectionLevel:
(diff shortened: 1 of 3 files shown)


Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -970,22 +970,19 @@ struct ConnectionStateMachine {
...
-        case .closing:
-            let cleanupContext = self.setErrorAndCreateCleanupContext(error)
-            return .closeConnectionAndCleanup(cleanupContext)
-        case .closed:
-            preconditionFailure("How can an error occur if the connection is already closed?")
+        case .error, .closing, .closed:
+            // We might run into this case because of reentrancy. For example: After we received an
+            // backend unexpected message, that we read of the wire, we bring this connection into
+            // the error state and will try to close the connection. The call to close() might be
+            // executed sync. This means that we received a `channelInactive` before we
```
gwynne, line 981: before we what...?

gwynne (approved):
LGTM aside from one comment nit

gwynne (approved):

PR #382, opened 2023-08-05
Upgrade CI

gwynne:
- Improves naming conventions a little
- Factors out code coverage into its own single job (results in overall speed improvement for CI)
- Adds workflow for analyzing the code with GitHub's CodeQL. Initial commit purposely leaves out the additional permission requirement for the token so it can't open any security complaints on the repo yet.


.github/workflows/test.yml
```diff
@@ -128,44 +124,56 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container: swift:jammy
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Run unit tests for coverage data
+        run: swift test --filter=^PostgresNIOTests --enable-code-coverage
+      - name: Upload coverage data
+        uses: vapor/swift-codecov-action@v0.2
+
```
fabianfett, line 172: why did you move code cov into its own job? why isn't this part of unit testing anymore?

gwynne: Because I did some digging into what actually results from providing the code coverage from every unit test variant. The results are always identical (except where there are Swift- or DB-version dependent semantics that are touched by unit tests, which is extremely rare right across all of Vapor). In short, it slows down each individual variant (both the build _and_ the execution, enough to be easily measurable in some cases) and provides nowhere near enough valuable data to be worth the trouble even in CI. Splitting it out to its own run adds another job, yes, but that job doesn't need to have TSan enabled; in most repos the end result is the overall time taken to run CI goes _down_, despite using the additional queue slot.

And on a minor note, Codecov's backend seems to appreciate not having to process 20 identical uploads for every single push to a PR; the upload failure rate drops from ~4% to literally zero (at least as observed thus far). On, and for good measure, the additional env vars that were being included with the uploads don't seem to be visible anywhere in Codecov's interface, so even where there _were_ differences between coverage on a per-variant basis, you couldn't tell which was which anyway.

fabianfett (approved):
Love it! Thanks!

PR #383, opened 2023-08-06
`close()` closes immediately; Add new `closeGracefully()`

### LINKED ISSUE #370: `PostgresConnection.close()` should just close the connection
### Current behavior

- `PostgresConnection.close()` waits for all the queries, that were previously scheduled on the connection to finish, before it closes the connection

### Expected behavior

- `PostgresConnection.close()` should just close the connection.
- We should add a `PostgresConnection.closeGracefully()` that keeps the current `close` behavior

fabianfett:
Fixes #370.

This PR changes the behavior of `close()`. `close()` now closes a connection immediately. We introduce a `closeGracefully()` that now has the same behavior as close before. Since we never documented the old close behavior and it is unfortunate that close does depend on the remote server to fulfill all queries, we consider it okay to change the behavior in this way. The current behavior is harmful since there is currently no way to force close a connection. Users might be in a position to force close a connection if a server stops responding. 

Alternative: close keeps graceful close behavior and we add a closeForcefully method instead

### DIFF
--- a/Sources/PostgresNIO/New/PSQLError.swift
+++ b/Sources/PostgresNIO/New/PSQLError.swift
@@ -54,6 +54,12 @@ public struct PSQLError: Error {
         public static let listenFailed = Self.init(.listenFailed)
         public static let unlistenFailed = Self.init(.unlistenFailed)
 
+        @available(*, deprecated, renamed: "clientClosesConnection")
+        public static let connectionQuiescing = Self.clientClosesConnection
+
+        @available(*, deprecated, message: "Use the more specific `serverClosedConnection` or `clientClosedConnection` instead")
+        public static let connectionClosed = Self.serverClosedConnection
+
         public var description: String {
             switch self.base {
             case .sslUnsupported:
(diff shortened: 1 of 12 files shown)


Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -45,13 +46,20 @@ public struct PSQLError: Error {
...
+        public static let serverClosedConnection = Self(.serverClosedConnection)
         public static let connectionError = Self(.connectionError)
         public static let uncleanShutdown = Self.init(.uncleanShutdown)
         public static let listenFailed = Self.init(.listenFailed)
         public static let unlistenFailed = Self.init(.unlistenFailed)
 
+        @available(*, deprecated, renamed: "clientClosesConnection")
+        public static let connectionQuiescing = Self.clientClosesConnection
+
+        @available(*, deprecated, message: "Use the more specific `serverClosedConnection` or `clientClosedConnection` instead")
```
fabianfett, line 60: 🚨 @gwynne

gwynne (approved):
I concur that the change in behavior is worthwhile.

PR #384, opened 2023-08-07
Cleanup PostgresBackendMessage

fabianfett:
We have a bunch of patterns in `PostgresBackendMessage` that are very unneeded. Let's remove those, because we think less code is better!


gwynne (approved):
I'm not sure I've ever approved more strongly of a PR ❤️

PR #386, opened 2023-08-07
PostgresBackendMessage.ID should be backed by UInt8 directly

fabianfett:
Remove code that can automatically be created by the compiler.


gwynne (approved):
Seriously, keep doing these! ❤️

PR #389, opened 2023-08-08
Use EventLoop provided by SwiftNIO's MultiThreadedEventLoopGroup.singleton

### LINKED ISSUE #388: Allow PostgresConnection creation without specifying an EventLoop
SwiftNIO has landed EventLoopGroup singletons in `2.58.0`. See https://github.com/apple/swift-nio/pull/2471.

We should add an additional PostgresConnection connect method, that internally uses the existing one, but uses the NIO singleton:

```swift
    static func connect(
        configuration: PostgresConnection.Configuration,
        id connectionID: PostgresConnection.ID,
        logger: Logger
    ) async throws -> PostgresConnection {
        try await Self.connect(
            on: MultiThreadedEventLoopGroup.singleton.any(),
            configuration: configuration,
            id: connectionID,
            logger: logger
        )
    }
}
```

- [ ] Add API
- [ ] Remove all `EventLoopGroup` usages in the README

tkrajacic:
Since SwiftNIO provides a global EventLoopGroup singleton, you can now create a PostgresConnection without explicitly providing an EventLoop. In such a case, an eventLoop from this singleton will be used.

```swift
// Uses the MultiThreadedEventLoopGroup.singleton.any() eventLoop
let connection = try await PostgresConnection.connect(
  configuration: config,
  id: 1,
  logger: logger
)
```

Fixes #388


fabianfett (commented):
Two small nits.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -378,6 +378,28 @@ extension PostgresConnection {
...
+    ///
+    /// This method uses an `EventLoop` from `MultiThreadedEventLoopGroup.singleton`, which is provided by SwiftNIO
+    ///
+    /// - Parameters:
+    ///   - configuration: A ``Configuration`` that shall be used for the connection
+    ///   - connectionID: An `Int` id, used for metadata logging
+    ///   - logger: A logger to log background events into
+    /// - Returns: An established  ``PostgresConnection`` asynchronously that can be used to run queries.
+    public static func connect(
+        configuration: PostgresConnection.Configuration,
```
weissi, line 391: We're still figuring out the patterns here so I don't know if this is a good suggestion but maybe a defaulted parameter `on eventLoop: EventLoop = MultiThreadedEventLoopGroup.singleton.any()` could work?

fabianfett: Not sure about this in the context of NIOTS... If we can import NIOTS, I would love to use the NIOTS EventLoop in the future for those cases. Defaulting inside the API might prove problematic then.

weissi: Once NIOTS is supported, I'd suggest to go down the route of my AHC PR: https://github.com/swift-server/async-http-client/pull/697/files#diff-914333008acc7684e7d98c700381a134ecdf49f095ce644e5898e857a16feabfR880

Then PostgresNIO could provide a `PostgresNIO.defaultEventLoopGroup` which would return either `MTELG.singleton` or `NIOTSELG.singleton`. And apart from the default value nothing needs to change.

If you don't want to "leak" that we're using MTELG.singleton today, then I'd suggest to add the `static var defaultEventLoopGroup` to PostgresNIO with this PR.

fabianfett: @weissi Lovely! @tkrajacic can you take ti from here?

fabianfett: @tkrajacic Just discussed this a bit further with @weissi:

```swift
     /// Returns the default `EventLoopGroup` singleton, automatically selecting the best for the platform.
     ///
     /// This will select the concrete `EventLoopGroup` depending which platform this is running on.
     public static var defaultEventLoopGroup: EventLoopGroup {
 #if canImport(Network)
         if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *) {
             return NIOTSEventLoopGroup.singleton
         } else {
             return MultiThreadedEventLoopGroup.singleton
         }
 #else
         return MultiThreadedEventLoopGroup.singleton
 #endif
     }
 }
```

And then we should default the existing connect method to `. defaultEventLoopGroup`, instead of creating a separate connect method.

tkrajacic: @fabianfett I have all that locally staged, but I was waiting until I could use the new NIOTS default eventloop group. So I am waiting for that to get released

fabianfett (approved):
Looks great! Thanks @tkrajacic

gwynne (approved):

Issue #385, opened 2023-08-07
async/await prepared statement API


fabianfett:
Before we can remove the existing prepared statement API, we need to introduce a new async/await one.

PR #390, opened 2023-08-09
async/await prepared statement API

mariosangiorgio:
Implementation of async/await API for prepared statements (See also https://github.com/vapor/postgres-nio/issues/385).

This PR adds a new `PreparedStatement` protocol to represent prepared statements and an `execute` function on `PostgresConnection` to prepare and execute statements.

To implement the features the PR also introduces a `PreparedStatementStateMachine` that keeps track of the state of a prepared statement at the connection level. This ensures that, for each connection, each statement is prepared once at time of first use and then subsequent uses are going to skip the preparation step and just execute it.

## Example usage

First define the struct to represent the prepared statement:
```swift
 struct ExamplePreparedStatement: PreparedStatement {
     static var sql = "SELECT pid, datname FROM pg_stat_activity WHERE state = $1"
     typealias Row = (Int, String)
     var state: String
     
    func makeBindings() -> PostgresBindings {
        var bindings = PostgresBindings()
        bindings.append(.init(string: self.state))
        return bindings
    }

    func decodeRow(_ row: PostgresNIO.PostgresRow) throws -> Row {
        try row.decode(Row.self)
    }
}
```
then, assuming you already have a `PostgresConnection` you can execute it:
```swift
let preparedStatement = ExamplePreparedStatement(state: "active")
let results = try await connection.execute(preparedStatement, logger: logger)
for (pid, database) in results {
    print("PID: \(pid), database: \(database)")
}
```

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -462,15 +462,15 @@ extension PostgresConnection {
     }
 
     /// Execute a prepared statement, taking care of the preparation when necessary
-    public func execute<P: PreparedStatement>(
-        _ preparedStatement: P,
+    public func execute<Statement: PostgresPreparedStatement>(
+        _ preparedStatement: Statement,
         logger: Logger
-    ) async throws -> AsyncThrowingMapSequence<PostgresRowSequence, P.Row>
+    ) async throws -> AsyncThrowingMapSequence<PostgresRowSequence, Statement.Row>
     {
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
         let task = HandlerTask.executePreparedStatement(.init(
-            name: String(reflecting: P.self),
-            sql: P.sql,
+            name: String(reflecting: Statement.self),
+            sql: Statement.sql,
             bindings: preparedStatement.makeBindings(),
             logger: logger,
             promise: promise
@@ -480,6 +480,27 @@ extension PostgresConnection {
             .map { $0.asyncSequence() }
             .get()
             .map { try preparedStatement.decodeRow($0) }
+    }
+
+    /// Execute a prepared statement, taking care of the preparation when necessary
+    public func execute<Statement: PostgresPreparedStatement>(
+        _ preparedStatement: Statement,
+        logger: Logger
+    ) async throws -> String
+    where Statement.Row == ()
+    {
+        let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
+        let task = HandlerTask.executePreparedStatement(.init(
+            name: String(reflecting: Statement.self),
+            sql: Statement.sql,
+            bindings: preparedStatement.makeBindings(),
+            logger: logger,
+            promise: promise
+        ))
+        self.channel.write(task, promise: nil)
+        return try await promise.futureResult
+            .map { $0.commandTag }
+            .get()
     }
 }
 
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -468,7 +468,7 @@ extension PostgresConnection {
         file: String = #fileID,
         line: Int = #line
     ) async throws -> AsyncThrowingMapSequence<PostgresRowSequence, Statement.Row> {
-        let bindings = preparedStatement.makeBindings()
+        let bindings = try preparedStatement.makeBindings()
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
         let task = HandlerTask.executePreparedStatement(.init(
             name: String(reflecting: Statement.self),
@@ -502,7 +502,7 @@ extension PostgresConnection {
         file: String = #fileID,
         line: Int = #line
     ) async throws -> String where Statement.Row == () {
-        let bindings = preparedStatement.makeBindings()
+        let bindings = try preparedStatement.makeBindings()
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
         let task = HandlerTask.executePreparedStatement(.init(
             name: String(reflecting: Statement.self),
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -462,12 +462,13 @@ extension PostgresConnection {
     }
 
     /// Execute a prepared statement, taking care of the preparation when necessary
-    public func execute<Statement: PostgresPreparedStatement>(
+    @inlinable
+    public func execute<Statement: PostgresPreparedStatement, Row>(
         _ preparedStatement: Statement,
         logger: Logger,
         file: String = #fileID,
         line: Int = #line
-    ) async throws -> AsyncThrowingMapSequence<PostgresRowSequence, Statement.Row> {
+    ) async throws -> AsyncThrowingMapSequence<PostgresRowSequence, Row> where Row == Statement.Row {
         let bindings = try preparedStatement.makeBindings()
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
         let task = HandlerTask.executePreparedStatement(.init(
@@ -496,6 +497,7 @@ extension PostgresConnection {
     }
 
     /// Execute a prepared statement, taking care of the preparation when necessary
+    @inlinable
     public func execute<Statement: PostgresPreparedStatement>(
         _ preparedStatement: Statement,
         logger: Logger,
@@ -525,7 +527,6 @@ extension PostgresConnection {
             )
             throw error // rethrow with more metadata
         }
-
     }
 }
 
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -464,9 +464,10 @@ extension PostgresConnection {
     /// Execute a prepared statement, taking care of the preparation when necessary
     public func execute<Statement: PostgresPreparedStatement>(
         _ preparedStatement: Statement,
-        logger: Logger
-    ) async throws -> AsyncThrowingMapSequence<PostgresRowSequence, Statement.Row>
-    {
+        logger: Logger,
+        file: String = #fileID,
+        line: Int = #line
+    ) async throws -> AsyncThrowingMapSequence<PostgresRowSequence, Statement.Row> {
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
         let task = HandlerTask.executePreparedStatement(.init(
             name: String(reflecting: Statement.self),
@@ -485,10 +486,10 @@ extension PostgresConnection {
     /// Execute a prepared statement, taking care of the preparation when necessary
     public func execute<Statement: PostgresPreparedStatement>(
         _ preparedStatement: Statement,
-        logger: Logger
-    ) async throws -> String
-    where Statement.Row == ()
-    {
+        logger: Logger,
+        file: String = #fileID,
+        line: Int = #line
+    ) async throws -> String where Statement.Row == () {
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
         let task = HandlerTask.executePreparedStatement(.init(
             name: String(reflecting: Statement.self),
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -462,7 +462,6 @@ extension PostgresConnection {
     }
 
     /// Execute a prepared statement, taking care of the preparation when necessary
-    @inlinable
     public func execute<Statement: PostgresPreparedStatement, Row>(
         _ preparedStatement: Statement,
         logger: Logger,
@@ -497,7 +496,6 @@ extension PostgresConnection {
     }
 
     /// Execute a prepared statement, taking care of the preparation when necessary
-    @inlinable
     public func execute<Statement: PostgresPreparedStatement>(
         _ preparedStatement: Statement,
         logger: Logger,
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -468,19 +468,31 @@ extension PostgresConnection {
         file: String = #fileID,
         line: Int = #line
     ) async throws -> AsyncThrowingMapSequence<PostgresRowSequence, Statement.Row> {
+        let bindings = preparedStatement.makeBindings()
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
         let task = HandlerTask.executePreparedStatement(.init(
             name: String(reflecting: Statement.self),
             sql: Statement.sql,
-            bindings: preparedStatement.makeBindings(),
+            bindings: bindings,
             logger: logger,
             promise: promise
         ))
         self.channel.write(task, promise: nil)
-        return try await promise.futureResult
-            .map { $0.asyncSequence() }
-            .get()
-            .map { try preparedStatement.decodeRow($0) }
+        do {
+            return try await promise.futureResult
+                .map { $0.asyncSequence() }
+                .get()
+                .map { try preparedStatement.decodeRow($0) }
+        } catch var error as PSQLError {
+            error.file = file
+            error.line = line
+            error.query = .init(
+                unsafeSQL: Statement.sql,
+                binds: bindings
+            )
+            throw error // rethrow with more metadata
+        }
+
     }
 
     /// Execute a prepared statement, taking care of the preparation when necessary
@@ -490,18 +502,30 @@ extension PostgresConnection {
         file: String = #fileID,
         line: Int = #line
     ) async throws -> String where Statement.Row == () {
+        let bindings = preparedStatement.makeBindings()
         let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
         let task = HandlerTask.executePreparedStatement(.init(
             name: String(reflecting: Statement.self),
             sql: Statement.sql,
-            bindings: preparedStatement.makeBindings(),
+            bindings: bindings,
             logger: logger,
             promise: promise
         ))
         self.channel.write(task, promise: nil)
-        return try await promise.futureResult
-            .map { $0.commandTag }
-            .get()
+        do {
+            return try await promise.futureResult
+                .map { $0.commandTag }
+                .get()
+        } catch var error as PSQLError {
+            error.file = file
+            error.line = line
+            error.query = .init(
+                unsafeSQL: Statement.sql,
+                binds: bindings
+            )
+            throw error // rethrow with more metadata
+        }
+
     }
 }
 
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -460,6 +460,27 @@ extension PostgresConnection {
             self.channel.write(task, promise: nil)
         }
     }
+
+    /// Execute a prepared statement, taking care of the preparation when necessary
+    public func execute<P: PreparedStatement>(
+        _ preparedStatement: P,
+        logger: Logger
+    ) async throws -> AsyncThrowingMapSequence<PostgresRowSequence, P.Row>
+    {
+        let promise = self.channel.eventLoop.makePromise(of: PSQLRowStream.self)
+        let task = HandlerTask.executePreparedStatement(.init(
+            name: String(reflecting: P.self),
+            sql: P.sql,
+            bindings: preparedStatement.makeBindings(),
+            logger: logger,
+            promise: promise
+        ))
+        self.channel.write(task, promise: nil)
+        return try await promise.futureResult
+            .map { $0.asyncSequence() }
+            .get()
+            .map { try preparedStatement.decodeRow($0) }
+    }
 }
 
 // MARK: EventLoopFuture interface
--- a/Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
+++ b/Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
@@ -49,7 +49,7 @@ struct PreparedStatementStateMachine {
         switch state {
         case .preparing(let statements):
             // When sending the bindings we are going to ask for binary data.
-            if var rowDescription {
+            if var rowDescription = rowDescription {
                 for i in 0..<rowDescription.columns.count {
                     rowDescription.columns[i].format = .binary
                 }
--- a/Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
+++ b/Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
@@ -6,21 +6,22 @@ struct PreparedStatementStateMachine {
         case prepared(RowDescription?)
         case error(PSQLError)
     }
-    
-    enum Action {
-        case prepareStatement
-        case waitForAlreadyInFlightPreparation
-        case executePendingStatements([PreparedStatementContext], RowDescription?)
-        case returnError([PreparedStatementContext], PSQLError)
-    }
-    
+
     var preparedStatements: [String: State]
     
     init() {
         self.preparedStatements = [:]
     }
-    
-    mutating func lookup(name: String, context: PreparedStatementContext) -> Action {
+
+    enum LookupAction {
+        case prepareStatement
+        case waitForAlreadyInFlightPreparation
+        case executeStatement(RowDescription?)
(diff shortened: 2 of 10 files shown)


Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
```diff
@@ -0,0 +1,68 @@
+import NIOCore
+
+struct PreparedStatementStateMachine {
+    enum State {
+        case preparing([PreparedStatementContext])
+        case prepared(RowDescription?)
+        case error(PSQLError)
```
mariosangiorgio, line 7: Here I keep track of `PSQLError` instead of a `PostgresBackendMessage.ErrorResponse` because that's what I get from the state machine that deals with statement preparation.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -235,6 +238,56 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+            case .prepareStatement:
+                let promise = self.eventLoop.makePromise(of: RowDescription?.self)
+                promise.futureResult.whenSuccess { rowDescription in
+                    self.prepareStatementComplete(
+                        name: preparedStatement.name,
+                        rowDescription: rowDescription,
+                        context: context
+                    )
+                }
+                promise.futureResult.whenFailure { error in
```
mariosangiorgio, line 255: With this I want to propagate failures happening during preparation to all the pending queries we have.
I am not 100% convinced by the way it looks (especially by the need to do `error as! PSQLError`) but I cannot find any better alternative.

Do you have any advice?

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -235,6 +238,56 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+                for statement in pendingStatements {
+                    let action = self.state.enqueue(task: .extendedQuery(.init(
+                        executeStatement: .init(
+                            name: statement.name,
+                            binds: statement.bindings,
+                            rowDescription: rowDescription),
+                        logger: statement.logger,
+                        promise: statement.promise
+                    )))
+                    self.run(action, with: context)
```
mariosangiorgio, line 282: Here (and in a couple other places) I need to execute all the pending statement executions.
To do that I `enqueue` and `run` many actions at the same time. Is this okay?

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -448,6 +448,27 @@ extension PostgresConnection {
             self.channel.write(task, promise: nil)
         }
     }
+
+    /// Execute a prepared statement, taking care of the preparation when necessary
+    public func execute<P: PreparedStatement>(
+        _ preparedStatement: P,
+        logger: Logger
+    ) async throws -> AsyncThrowingMapSequence<PostgresRowSequence, P.Row>
```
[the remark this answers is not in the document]
fabianfett: `some AsyncSequence where Element == P.Row`?

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -448,6 +448,27 @@ extension PostgresConnection {
             self.channel.write(task, promise: nil)
         }
     }
+
+    /// Execute a prepared statement, taking care of the preparation when necessary
+    public func execute<P: PreparedStatement>(
```
fabianfett, line 453: Lets rename `P` to `Statement`

Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
```diff
@@ -0,0 +1,68 @@
+import NIOCore
+
+struct PreparedStatementStateMachine {
+    enum State {
+        case preparing([PreparedStatementContext])
```
fabianfett, line 5: why is this an array?

Sources/PostgresNIO/Connection/PostgresConnection.swift
fabianfett, line 456: also let's overload this method with a method where `Statement.Row == ()` where we only return the command tag?

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -235,6 +238,56 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
                 listener.failed(CancellationError())
                 return
             }
+        case .executePreparedStatement(let preparedStatement):
+            switch self.preparedStatementState.lookup(
+                name: preparedStatement.name,
+                context: preparedStatement
+            ) {
```
fabianfett, line 245: can we assign a variable here and then switch over the variable?

Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
```diff
@@ -0,0 +1,68 @@
...
+        case prepared(RowDescription?)
+        case error(PSQLError)
+    }
+    
+    enum Action {
+        case prepareStatement
+        case waitForAlreadyInFlightPreparation
+        case executePendingStatements([PreparedStatementContext], RowDescription?)
+        case returnError([PreparedStatementContext], PSQLError)
+    }
```
fabianfett, line 15: returning the same Action for all mutations is what I did in the past, but what we generally frown upon by now. Can we get specific Actions for each mutation?

Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
```diff
@@ -0,0 +1,68 @@
...
+    
+    mutating func lookup(name: String, context: PreparedStatementContext) -> Action {
+        if let state = self.preparedStatements[name] {
+            switch state {
+            case .preparing(var statements):
+                statements.append(context)
+                self.preparedStatements[name] = .preparing(statements)
+                return .waitForAlreadyInFlightPreparation
+            case .prepared(let rowDescription):
+                return .executePendingStatements([context], rowDescription)
```
fabianfett, line 31: we shouldn't allocate here :P

Sources/PostgresNIO/New/PreparedStatement.swift
```diff
@@ -0,0 +1,40 @@
...
+///
+///     func decodeRow(_ row: PostgresNIO.PostgresRow) throws -> Row {
+///         try row.decode(Row.self)
+///     }
+/// }
+/// ```
+///
+/// Structs conforming to this protocol can then be used with `PostgresConnection.execute(_ preparedStatement:, logger:)`,
+/// which will take care of preparing the statement on the server side and executing it.
+public protocol PreparedStatement {
```
fabianfett, line 28: we should require that this is Sendable :)

fabianfett (commented):
First review pass

fabianfett:
Fixed #385

Sources/PostgresNIO/New/PreparedStatement.swift
fabianfett: Also we should prefix this one with `Postgres`. Reason: https://github.com/apple/swift-nio/blob/main/docs/public-api.md#1-no-global-namespace-additions-that-arent-prefixed

Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
mariosangiorgio: So that we can send the `PREPARE` command only once if there are are multiple attempts to execute the same prepared statement before the command reaches the server.

I thought we said this is something that can happen and we want to handle

Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    enum PreparationCompleteAction {
+        case executePendingStatements([PreparedStatementContext], RowDescription?)
+    }
+
+    mutating func preparationComplete(
+        name: String,
+        rowDescription: RowDescription?
+    ) -> PreparationCompleteAction {
```
fabianfett, line 49: would this work better as a simple struct?

Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
```diff
@@ -0,0 +1,77 @@
...
+            self.preparedStatements[name] = .prepared(nil)
+            return .executePendingStatements(statements, nil)
+        }
+    }
+
+    enum ErrorHappenedAction {
+        case returnError([PreparedStatementContext], PSQLError)
+    }
+    
+    mutating func errorHappened(name: String, error: PSQLError) -> ErrorHappenedAction {
```
fabianfett, line 70: is a simple struct easier here?

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -33,6 +34,7 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
         self.state = ConnectionStateMachine(requireBackendKeyData: configuration.options.requireBackendKeyData)
         self.eventLoop = eventLoop
         self.listenState = ListenStateMachine()
+        self.preparedStatementState = PreparedStatementStateMachine()
```
fabianfett, line 37: I wonder if we should just init those two directly in the declaration...

Sources/PostgresNIO/New/PostgresChannelHandler.swift
fabianfett: First I think we should use whenComplete here instead of whenSuccess and whenFailure. Reason for this is, we can save an allocation...

Second, we must check if the error we get is actually a PSQLError, if it is not, we need to wrap error with a PSQLError.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -233,6 +236,66 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+                    )))
+                    self.run(action, with: context)
+                }
+                return
+            case .returnError(let pendingStatements, let error):
+                for statement in pendingStatements {
+                    statement.promise.fail(error)
+                }
+                return
+            }
```
fabianfett, line 298: can we move this into its own method? we should likely do the same with all the other longer ones above.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -664,6 +727,49 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+        rowDescription: RowDescription?,
+        context: ChannelHandlerContext
+    ) {
+        let action = self.preparedStatementState.preparationComplete(
+            name: name,
+            rowDescription: rowDescription
+        )
+        guard case .executePendingStatements(let statements, let rowDescription) = action else {
+            preconditionFailure("Expected to have pending statements to execute")
+        }
```
fabianfett, line 741: we should reconsider the return type from the statemachine here :). No enum needed.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -664,6 +727,49 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+        error: PSQLError,
+        context: ChannelHandlerContext
+    ) {
+        let action = self.preparedStatementState.errorHappened(
+            name: name,
+            error: error
+        )
+        guard case .returnError(let statements, let error) = action else {
+            preconditionFailure("Expected to have pending statements to execute")
+        }
```
fabianfett, line 768: see above.

fabianfett (commented):
Looks better and better. For unit test examples have a look at `PostgresConnectionTests.testSimpleListen`

Sources/PostgresNIO/New/PreparedStatement.swift
```diff
@@ -0,0 +1,40 @@
...
+/// ```swift
+/// struct Example: PostgresPreparedStatement {
+///     static var sql = "SELECT pid, datname FROM pg_stat_activity WHERE state = $1"
+///     typealias Row = (Int, String)
+///
+///     var state: String
+///
+///     func makeBindings() -> PostgresBindings {
+///         var bindings = PostgresBindings()
+///         bindings.append(.init(string: self.state))
```
fabianfett, line 16:
```suggestion
///         bindings.append(self.state)
```
does this work as well?

mariosangiorgio: No, I get back an error: `Cannot convert value of type 'String' to expected argument type 'PostgresData'`

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -664,6 +688,92 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+                    name: preparedStatement.name,
+                    rowDescription: rowDescription,
+                    context: context
+                )
+            case .failure(let error):
+                let psqlError: PSQLError
+                if let error = error as? PSQLError {
+                    psqlError = error
+                } else {
+                    psqlError = .connectionError(underlying: error)
```
mariosangiorgio, line 709: @fabianfett we said to wrap the `Error` in a `PSQLError`.

All the cases I see here
```swift
        enum Base: Sendable, Hashable {
            case sslUnsupported
            case failedToAddSSLHandler
            case receivedUnencryptedDataAfterSSLRequest
            case server
            case messageDecodingFailure
            case unexpectedBackendMessage
            case unsupportedAuthMechanism
            case authMechanismRequiresPassword
            case saslError
            case invalidCommandTag

            case queryCancelled
            case tooManyParameters
            case clientClosesConnection
            case clientClosedConnection
            case serverClosedConnection
            case connectionError
            case uncleanShutdown

            case listenFailed
            case unlistenFailed
        }
```
are quite specific. Is there any case I should use to wrap a generic error? Inspecting error to gather more information could be tricky

Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
```diff
@@ -0,0 +1,83 @@
...
+    struct PreparationCompleteAction {
+        var statements: [PreparedStatementContext]
+        var rowDescription: RowDescription?
+    }
+
+    mutating func preparationComplete(
+        name: String,
+        rowDescription: RowDescription?
+    ) -> PreparationCompleteAction {
+        guard case .preparing(let statements) = self.preparedStatements[name] else {
```
fabianfett, line 46: can we switch over all cases here?

Sources/PostgresNIO/New/Connection State Machine/PreparedStatementStateMachine.swift
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    struct ErrorHappenedAction {
+        var statements: [PreparedStatementContext]
+        var error: PSQLError
+    }
+    
+    mutating func errorHappened(name: String, error: PSQLError) -> ErrorHappenedAction {
+        guard case .preparing(let statements) = self.preparedStatements[name] else {
```
fabianfett, line 74: can we switch over all cases here?

Sources/PostgresNIO/New/PreparedStatement.swift
fabianfett: Oh, we shouldn't create `PostgresData`. The problem here is that we better names for those methods, so that users can use them without Interpolation as well.

```swift
struct PostgresBindings {
    @inlinable
    public mutating func appendInterpolation<Value: PostgresEncodable>(_ value: Value) throws {
        try self.binds.append(value, context: .default)
    }

    @inlinable
    public mutating func appendInterpolation<Value: PostgresEncodable>(_ value: Optional<Value>) throws {
        switch value {
        case .none:
            self.binds.appendNull()
        case .some(let value):
            try self.binds.append(value, context: .default)
        }

    }
}
```

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -664,6 +688,92 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+
+    private func makeExecutPreparedStatementAction(
+        preparedStatement: PreparedStatementContext,
+        rowDescription: RowDescription?
+    ) -> PSQLTask {
+        return .extendedQuery(.init(
+            executeStatement: .init(
+                name: preparedStatement.name,
+                binds: preparedStatement.bindings,
+                rowDescription: rowDescription),
```
fabianfett, line 734:
```suggestion
                rowDescription: rowDescription
            ),
```

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -664,6 +688,92 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+        }
+        return .extendedQuery(.init(
+            name: preparedStatement.name,
+            query: preparedStatement.sql,
+            logger: preparedStatement.logger,
+            promise: promise
+        ))
+    }
+
+    private func makeExecutPreparedStatementAction(
```
fabianfett, line 726:
```suggestion
    private func makeExecutePreparedStatementTask(
```

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -233,6 +234,29 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+                psqlTask = self.makePrepareStatementAction(
+                    preparedStatement: preparedStatement,
+                    context: context
+                )
+            case .waitForAlreadyInFlightPreparation:
+                // The state machine already keeps track of this
+                // and will execute the statement as soon as it's prepared
+                return
+            case .executeStatement(let rowDescription):
+                psqlTask = self.makeExecutPreparedStatementAction(
```
fabianfett, line 252:
```suggestion
                psqlTask = self. makeExecutePreparedStatementTask(
```

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -233,6 +234,29 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
                 listener.failed(CancellationError())
                 return
             }
+        case .executePreparedStatement(let preparedStatement):
+            let action = self.preparedStatementState.lookup(
+                preparedStatement: preparedStatement
+            )
+            switch action {
+            case .prepareStatement:
+                psqlTask = self.makePrepareStatementAction(
```
fabianfett, line 243:
```suggestion
                psqlTask = self.makePrepareStatementTask(
```

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -664,6 +688,92 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
         }
     }
 
+    private func makePrepareStatementAction(
```
fabianfett, line 691:
```suggestion
    private func makePrepareStatementTask(
```

Sources/PostgresNIO/New/PreparedStatement.swift
```diff
@@ -0,0 +1,40 @@
+/// A prepared statement.
+///
+/// Structs conforming to this protocol will need to provide the SQL statement to
+/// send to the server and a way of creating bindings are decoding the result.
+///
+/// As an example, consider this struct:
+/// ```swift
+/// struct Example: PostgresPreparedStatement {
+///     static var sql = "SELECT pid, datname FROM pg_stat_activity WHERE state = $1"
```
fabianfett, line 9: Making sure our example is Sendable.
```suggestion
///     static let sql = "SELECT pid, datname FROM pg_stat_activity WHERE state = $1"
```

fabianfett (commented):
This looks so good. A few nits.

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -275,6 +275,100 @@ class PostgresConnectionTests: XCTestCase {
...
+            bindings.append(.init(string: self.state))
+            return bindings
+        }
+
+        func decodeRow(_ row: PostgresNIO.PostgresRow) throws -> Row {
+            try row.decode(Row.self)
+        }
+    }
+
+    func testPreparedStatement() async throws {
```
fabianfett, line 295: I know that these tests are somewhat painful to write but we should embrace them, as we can produce hard to reproduce scenarios here. 

For this reason, please lets try to create those scenarios in extra tests here:

- Sending the a prepared query twice, before the preparation step has succeeded. Is the order of the send statements still correct? Is the query just prepared once?
- Sending the a prepared, after the the query of the same type has already succeeded. Is the query executed right away?
- Failing a prepared statement in preparation are other executes failed in the same way right away?

Happy to help.

fabianfett (approved):
I'm extremely happy with this. Let's wait on the go from @gwynne and then let's merge it! Thanks @mariosangiorgio!

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -275,6 +275,288 @@ class PostgresConnectionTests: XCTestCase {
...
+                ])
+            )
+
+            let preparedRequest1 = try await channel.waitForPreparedRequest()
+            var buffer = preparedRequest1.bind.parameters[0]!
+            let parameter1 = buffer.readString(length: buffer.readableBytes)!
+            XCTAssertEqual(parameter1, "active")
+            try await channel.sendPreparedResponse(
+                dataRows: [
+                    DataRow(arrayLiteral: "test_database")
```
fabianfett, line 397:
```suggestion
                    ["test_database"]
```

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -275,6 +275,288 @@ class PostgresConnectionTests: XCTestCase {
...
+            )
+
+            let preparedRequest = try await channel.waitForPreparedRequest()
+            XCTAssertEqual(preparedRequest.bind.preparedStatementName, String(reflecting: TestPrepareStatement.self))
+            XCTAssertEqual(preparedRequest.bind.parameters.count, 1)
+            XCTAssertEqual(preparedRequest.bind.resultColumnFormats, [.binary])
+
+            try await channel.sendPreparedResponse(
+                dataRows: [
+                    DataRow(arrayLiteral: "test_database")
```
fabianfett, line 342:
```suggestion
                    ["test_database"]
```

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -275,6 +275,288 @@ class PostgresConnectionTests: XCTestCase {
...
+                XCTAssertEqual(rows, 1)
+            }
+
+            let preparedRequest2 = try await channel.waitForPreparedRequest()
+            buffer = preparedRequest2.bind.parameters[0]!
+            let parameter2 = buffer.readString(length: buffer.readableBytes)!
+            XCTAssertEqual(parameter2, "idle")
+            try await channel.sendPreparedResponse(
+                dataRows: [
+                    DataRow(arrayLiteral: "test_database")
```
fabianfett, line 421:
```suggestion
                    ["test_database"]
```

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -275,6 +275,288 @@ class PostgresConnectionTests: XCTestCase {
...
+            )
+
+            // Now both the tasks have their statements prepared.
+            // We should see both of their execute requests coming in, the order is nondeterministic
+            let preparedRequest1 = try await channel.waitForPreparedRequest()
+            var buffer = preparedRequest1.bind.parameters[0]!
+            let parameter1 = buffer.readString(length: buffer.readableBytes)!
+            try await channel.sendPreparedResponse(
+                dataRows: [
+                    DataRow(arrayLiteral: "test_database_\(parameter1)")
```
fabianfett, line 492:
```suggestion
                    ["test_database_\(parameter1)"]
```

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -275,6 +275,288 @@ class PostgresConnectionTests: XCTestCase {
...
+                ],
+                commandTag: TestPrepareStatement.sql
+            )
+            let preparedRequest2 = try await channel.waitForPreparedRequest()
+            buffer = preparedRequest2.bind.parameters[0]!
+            let parameter2 = buffer.readString(length: buffer.readableBytes)!
+            try await channel.sendPreparedResponse(
+                dataRows: [
+                    DataRow(arrayLiteral: "test_database_\(parameter2)")
```
fabianfett, line 501:
```suggestion
                    ["test_database_\(parameter2)"]
```

fabianfett (approved):
Super nits

gwynne (approved):

PR #391, opened 2023-08-09
Remove PrepareStatementStateMachine

fabianfett:
Preparing a statement is a substep of running an extended query. For this reason we should reuse the ExtendedQueryStateMachine as much as we can. This patch removes the `PrepareStatementStateMachine` and uses the `ExtendedQueryStateMachine`. As a result of this we can simplify our code in lots of other places.


Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -4,7 +4,7 @@ struct ExtendedQueryStateMachine {
     
     private enum State {
         case initialized(ExtendedQueryContext)
-        case parseDescribeBindExecuteSyncSent(ExtendedQueryContext)
+        case messagesSent(ExtendedQueryContext)
```
mariosangiorgio, line 7: Is the name `messageSent` intentionally generic? When I read it my first reaction was "What message? Do we have different kind of messages flying around?"

Sources/PostgresNIO/New/PSQLRowStream.swift
```diff
@@ -1,12 +1,23 @@
...
+}
+
 // Thread safety is guaranteed in the RowStream through dispatching onto the NIO EventLoop.
 final class PSQLRowStream: @unchecked Sendable {
     private typealias AsyncSequenceSource = NIOThrowingAsyncSequenceProducer<DataRow, Error, AdaptiveRowBuffer, PSQLRowStream>.Source
 
-    enum RowSource {
-        case stream(PSQLRowsDataSource)
+    enum Source {
+        case stream([RowDescription.Column], PSQLRowsDataSource)
```
mariosangiorgio, line 20: What's the difference between `RowDescription` and `[RowDescription.Column]`?

Sources/PostgresNIO/New/PSQLRowStream.swift
```diff
@@ -31,27 +42,28 @@ final class PSQLRowStream: @unchecked Sendable {
...
+    ) {
         let bufferState: BufferState
-        switch rowSource {
-        case .stream(let dataSource):
+        switch source {
+        case .stream(let rowDescription, let dataSource):
+            self.rowDescription = rowDescription
             bufferState = .streaming(buffer: .init(), dataSource: dataSource)
         case .noRows(.success(let commandTag)):
+            self.rowDescription = []
```
mariosangiorgio, line 56: In other parts of the codebase (e.g. in `PrepareStatementStateMachine`) we allowed row description to be `nil`. Is there any semantic difference between a `RowDescription?` with a `nil` value and an empty `[RowDescription.Column]`?

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
fabianfett: yes. it depends on what the internal QueryContext says. 

If we run an unnamed query, we sent `parse, describe, bind, execute & sync`.
If we prepare a statement, we sent `parse, describe & sync`.
If we execute a statement, we sent `bind, execute & sync`.

gwynne (approved):

PR #394, opened 2023-08-12
Remove generic usage from `avoidingStateMachineCoW()`  so it doesn't need to be force-inlined

gwynne:
The comment explaining `avoidingStateMachineCoW()` states:
```swift
/// Sadly, because it's generic and has a closure, we need to force it to be inlined
/// at all call sites, which is not ideal.
```
However, at all but one call site, it always returns the same type, so it doesn't really need to be generic and we can avoid the "not ideal" inlining requirement.


fabianfett:
I think this change is fine, BUT we should only merge it once we have perf tests, that validate that we don't regress performance here.

fabianfett (changes requested):
I think this change is fine, BUT we should only merge it once we have perf tests, that validate that we don't regress performance here.

Blocking now to prevent accidental merges.

PR #398, opened 2023-08-18
Fix a few inaccurate or confusing precondition failure messages

gwynne:
More for code clarity than anything else, really.


fabianfett (approved):
LGTM

PR #410, opened 2023-09-23
Add Interpolation Function for Collection types

MahdiBM:
Adds an interpolation function for collection types.

This has proved quite useful since manually interpolating an array is quite hard to do properly, to a `PostgresQuery`.

### DIFF
--- a/Sources/PostgresNIO/New/PostgresQuery.swift
+++ b/Sources/PostgresNIO/New/PostgresQuery.swift
@@ -99,7 +99,7 @@ extension PostgresQuery {
         @inlinable
         public mutating func appendInterpolation<C: Collection>(
             _ values: C
-        ) throws where C.Element: PostgresThrowingDynamicTypeEncodable {
+        ) throws where C.Element: PostgresArrayEncodable {
             guard !values.isEmpty else {
                 throw PostgresEncodableSequenceCannotBeEmpty()
             }
--- a/Sources/PostgresNIO/New/PostgresQuery.swift
+++ b/Sources/PostgresNIO/New/PostgresQuery.swift
@@ -92,7 +92,7 @@ extension PostgresQuery {
 
 
         @usableFromInline
-        struct PostgresEncodableSequenceCannotBeEmpty: Error {
+        struct InterpolatedCollectionCannotBeEmpty: Error {
             @usableFromInline init() { }
         }
 
@@ -101,7 +101,7 @@ extension PostgresQuery {
             _ values: C
         ) throws where C.Element: PostgresArrayEncodable {
             guard !values.isEmpty else {
-                throw PostgresEncodableSequenceCannotBeEmpty()
+                throw InterpolatedCollectionCannotBeEmpty()
             }
             let bindsSQL = try values.map { value in
                 try self.binds.append(value, context: .default)
--- a/Sources/PostgresNIO/New/PostgresQuery.swift
+++ b/Sources/PostgresNIO/New/PostgresQuery.swift
@@ -43,7 +43,6 @@ extension PostgresQuery {
             self.sql.append(contentsOf: literal)
         }
 
-        @_disfavoredOverload
         @inlinable
         public mutating func appendInterpolation<Value: PostgresThrowingDynamicTypeEncodable>(_ value: Value) throws {
             try self.binds.append(value, context: .default)
@@ -62,7 +61,6 @@ extension PostgresQuery {
             self.sql.append(contentsOf: "$\(self.binds.count)")
         }
 
-        @_disfavoredOverload
         @inlinable
         public mutating func appendInterpolation<Value: PostgresDynamicTypeEncodable>(_ value: Value) {
             self.binds.append(value, context: .default)
@@ -98,7 +96,7 @@ extension PostgresQuery {
 
         @inlinable
         public mutating func appendInterpolation<C: Collection>(
-            _ values: C
+            collection values: C
         ) throws where C.Element: PostgresArrayEncodable {
             guard !values.isEmpty else {
                 throw InterpolatedCollectionCannotBeEmpty()
--- a/Sources/PostgresNIO/New/PostgresQuery.swift
+++ b/Sources/PostgresNIO/New/PostgresQuery.swift
@@ -43,6 +43,7 @@ extension PostgresQuery {
             self.sql.append(contentsOf: literal)
         }
 
+        @_disfavoredOverload
         @inlinable
         public mutating func appendInterpolation<Value: PostgresThrowingDynamicTypeEncodable>(_ value: Value) throws {
             try self.binds.append(value, context: .default)
@@ -61,6 +62,7 @@ extension PostgresQuery {
             self.sql.append(contentsOf: "$\(self.binds.count)")
         }
 
+        @_disfavoredOverload
         @inlinable
         public mutating func appendInterpolation<Value: PostgresDynamicTypeEncodable>(_ value: Value) {
             self.binds.append(value, context: .default)
@@ -86,6 +88,26 @@ extension PostgresQuery {
         ) throws {
             try self.binds.append(value, context: context)
             self.sql.append(contentsOf: "$\(self.binds.count)")
+        }
+
+
+        @usableFromInline
+        struct PostgresEncodableSequenceCannotBeEmpty: Error {
+            @usableFromInline init() { }
+        }
+
+        @inlinable
+        public mutating func appendInterpolation<C: Collection>(
+            _ values: C
+        ) throws where C.Element: PostgresThrowingDynamicTypeEncodable {
+            guard !values.isEmpty else {
+                throw PostgresEncodableSequenceCannotBeEmpty()
+            }
+            let bindsSQL = try values.map { value in
+                try self.binds.append(value, context: .default)
+                return "$\(self.binds.count)"
+            }.joined(separator: ", ")
+            self.sql.append(contentsOf: "(\(bindsSQL))")
         }
 
         @inlinable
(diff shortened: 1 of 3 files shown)


Sources/PostgresNIO/New/PostgresQuery.swift
```diff
@@ -43,6 +43,7 @@ extension PostgresQuery {
             self.sql.append(contentsOf: literal)
         }
 
+        @_disfavoredOverload
```
gwynne, line 46: It's not satisfactory to me to need to use `@_disfavoredOverload` here, especially not twice. Is there another way to make the other overload more preferred? (Keep in mind that `@_disfavoredOverload` exists to work around constraint solver bugs, _not_ to let you change the language's overload resolution rules at a whim.)

Sources/PostgresNIO/New/PostgresQuery.swift
```diff
@@ -88,6 +90,26 @@ extension PostgresQuery {
...
+
+        @usableFromInline
+        struct PostgresEncodableSequenceCannotBeEmpty: Error {
+            @usableFromInline init() { }
+        }
+
+        @inlinable
+        public mutating func appendInterpolation<C: Collection>(
+            _ values: C
+        ) throws where C.Element: PostgresThrowingDynamicTypeEncodable {
```
gwynne, line 102: You might want to consider constraining to `PostgresArrayEncodable` depending on context

gwynne (changes requested):
I dunno, I feel slightly weird about this and couldn't say why. @fabianfett, thoughts?

Sources/PostgresNIO/New/PostgresQuery.swift
MahdiBM, line 46: I know this is not what you'd like to see, but i couldn't come up with a better solution.

Without this, as far as I could figure out, the new collection interpolate function will need a parameter name which will make it so users need to interpolate their values like `"\(parameterName: myValues)"`. 
The problem with this is that users can still do `"\(myValues)"` and it will compile, but it will fail when making the requests to Postgres.
With this new way though, the compiler will be able to figure that out and automatically interpolate the value as a sequence.

Sources/PostgresNIO/New/PostgresQuery.swift
```diff
@@ -88,6 +90,26 @@ extension PostgresQuery {
             self.sql.append(contentsOf: "$\(self.binds.count)")
         }
 
+
+        @usableFromInline
+        struct PostgresEncodableSequenceCannotBeEmpty: Error {
+            @usableFromInline init() { }
+        }
```
MahdiBM, line 97: Not sure where to define this error. Open to moving it to somewhere else.

Sources/PostgresNIO/New/PostgresQuery.swift
MahdiBM, line 102: We could also make this be more specific about the type just so there won't be accidental problems with someone not expecting their value to be inferred as a collection.
Maybe just have a function for Array and one for Set?

MahdiBM:
@gwynne sorry i was righting some comments, i knew you'll respond before my comments are done but couldn't do much 😅

Sources/PostgresNIO/New/PostgresQuery.swift
MahdiBM: See https://github.com/vapor/postgres-nio/pull/410#discussion_r1334975466

MahdiBM:
I do realize this is tricky and weird, but i also do think this will be a nice addition and hopefully worth the effort.

MahdiBM:
Screenshot 2023-09-23 at 3 42 33 PM

This test which has started to fail, shows a proper way how to do array interpolation.

@fabianfett @gwynne The approach of have this kind of interpolation with no casting looks better, but i honestly hadn't figured out that you could cast a bind to an array type and it would work, so not sure if this PR is still of enough value.

I'll remove the underscored attributes and use a parameter name for the collection interpolation instead, for now, until @fabianfett has the time to take a look.

Sources/PostgresNIO/New/PostgresQuery.swift
MahdiBM: Outdated, see https://github.com/vapor/postgres-nio/pull/410#issuecomment-1732296628

MahdiBM: Outdated, see https://github.com/vapor/postgres-nio/pull/410#issuecomment-1732296628

Sources/PostgresNIO/New/PostgresQuery.swift
MahdiBM: Outdated now that i changed the function to have a parameter name.

MahdiBM:
So, I'm testing a little bit and the PR seems to be more useful than I thought might be the case.

Basically, this is the same situation that made me create such function in a private project, and i can't find a way to do it without multiple different binds, which means it cannot be done with current `PostgresQuery` interpolation functions:
```sql
PREPARE my_query (bigint[]) AS SELECT my_column FROM my_table WHERE my_column in ($1);

Execute my_query( (1111, 2222) :: bigint[] );
```
```
ERROR: operator does not exist: bigint = bigint[] Hint: No operator matches the given name and argument types. You might need to add explicit type casts.
```
I've tested a lot of different prepared queries, non worked.

In contrast, what this PR adds: 
```sql
PREPARE my_query (bigint, bigint) AS SELECT my_column FROM my_table WHERE my_column in ($1, $2);

Execute my_query(1111, 2222);
```
This executes and returns results as expected. The example query would look like this in `PostgresQuery`:
```swift
"""
SELECT my_column FROM my_table WHERE my_column in \(collection: array_of_values)
"""
```

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -212,6 +212,31 @@ final class IntegrationTests: XCTestCase {
...
+        var conn: PostgresConnection?
+        XCTAssertNoThrow(conn = try PostgresConnection.test(on: eventLoop).wait())
+        defer { XCTAssertNoThrow(try conn?.close().wait()) }
+
+        var result: PostgresQueryResult?
+        let doubles: [Double] = [3.14, 42]
+        XCTAssertNoThrow(try conn?.query("CREATE TABLE foo (accuracy double precision)", logger: .psqlTest).wait())
+        XCTAssertNoThrow(try conn?.query("INSERT INTO foo VALUES (\(doubles[0]))", logger: .psqlTest).wait())
+        XCTAssertNoThrow(try conn?.query("INSERT INTO foo VALUES (\(doubles[1]))", logger: .psqlTest).wait())
+        XCTAssertNoThrow(result = try conn?.query("SELECT * FROM foo WHERE accuracy IN \(collection: doubles)", logger: .psqlTest).wait())
```
MahdiBM, line 229: Writing the test like the above function fails because postgres returns the values as the RECORD type.
Anyway, this test much better demonstrates what i actually wanted this new function to do.

Sources/PostgresNIO/New/PostgresQuery.swift
```diff
@@ -88,6 +88,26 @@ extension PostgresQuery {
...
+            collection values: C
+        ) throws where C.Element: PostgresArrayEncodable {
+            guard !values.isEmpty else {
+                throw InterpolatedCollectionCannotBeEmpty()
+            }
+            let bindsSQL = try values.map { value in
+                try self.binds.append(value, context: .default)
+                return "$\(self.binds.count)"
+            }.joined(separator: ", ")
+            self.sql.append(contentsOf: "(\(bindsSQL))")
```
fabianfett, line 108: This looks wrong to me... Why do we need to `for` loop this ourselves?

Does this sound better?

```suggestion
            try self.binds.append(value, context: .default)
            self.sql.append(contentsOf: "$\(self.binds.count)")
```

The problem currently is that it always go through the json path.

MahdiBM: > This looks wrong to me

Why?!

> Why do we need to for loop this ourselves?

That's the point, to produce something like 

```
SELECT my_column FROM my_table WHERE my_column in ($1, $2, $3, $4)
```

> Does this sound better?

Sound-wise yes, but i don't think it'll work with a _collection_. That's what the PR is trying to resolve.

> The problem currently is that it always go through the json path.

What goes through the JSON path? 🤔

MahdiBM: A query like
```sql
SELECT my_column FROM my_table WHERE my_column in $1
```
won't work. At least i couldn't get something like that to work.

fabianfett (changes requested):
@MahdiBM I think the problem here is a misunderstanding of what an array encodable means. An array encodable is not a short hand to write an *in* statement. 

Instead, Postgres supports having an array of something as a column type. And in those cases the binding to a single placeholder is correct. The postgres docs about this are here: https://www.postgresql.org/docs/current/arrays.html

Please note we currently only support one dimensional arrays.

Currently PostgresNIO, supports single bindings of an array, since the json overload will kick in, if you were to write this PostgresQuery:

SELECT \([“hello”, “world”]);

PostgresQuery would send this over the wire as a JSON array with two strings, instead of sending it as a Postgres TEXT array.

MahdiBM:
After asking some SQL folks, here's the conclusion:

I wanted to do a query like:
```sql
SELECT * FROM t WHERE c in $1
```
This is not possible to do properly without the additions that this PR introduces. 
Even with this PR, we need to use multiple binds instead of just 1.

It appears the correct way to do that using a single bind is to instead use `any`:
```sql
SELECT * FROM t WHERE c = any($1)
```

This means that if you think this addition is not of enough value, it's fine with me and I won't even need to leave this code as an extension in my code. I can instead use this alternative way I found with existing `PostgresQuery` stuff.

Issue #256, opened 2022-03-11
Provide a ConnectionPool


fabianfett:
We should provide a ConnectionPool out of the box.

Jerry-Carter:
Absolutely agreed.

Jerry-Carter:
I have a very simple connection pool implementation which is an actor with dependencies on NIO, PostgresNIO, and Logging.  Having the application pass in the PostgresConnection.Configuration would further simplify it.  Typical usage is to initialize the actor, and then Checkout() and Release() database connections as required in the application.

Happy to donate it, but I would expect better code to exist elsewhere in the Vapor code base.

CyrilusK:
<img width="1440" alt="Снимок экрана 2022-05-15 в 19 02 22" src="https://user-images.githubusercontent.com/47518991/168483944-79c658f9-1ffb-43f3-9d1c-cd023917d927.png">
Can you help with this?

trasch:
I can also offer a connection pool for those interested: https://github.com/Outdooractive/PostgresConnectionPool

majortom64:
For security purposes it would be great to support ephemeral database user credentials, generated by something like HashiCorp’s Vault. These credentials timeout and then need to be regenerated. When they change, the Connection Pool would need to automatically close or automatically handle getting the new credentials and updating the connection.

fabianfett:
Work can be followed in the [ConnectionPool milestone](https://github.com/vapor/postgres-nio/milestone/6).

fabianfett:
@majortom64 your request is tracked as #431.

fabianfett:
We've released a PostgresClient implementation behind an SPI flag in the `1.19.0` release:
https://github.com/vapor/postgres-nio/releases/tag/1.19.0

We would love to learn, if this solves your use-cases.

lovetodream:
I just implemented the `_ConnectionPoolModule` in the [Oracle driver](https://github.com/lovetodream/oracle-nio) I'm working on. The API is very easy to understand imo. It actually surprised me how little I had to do to get everything working. Thanks for the great work @fabianfett!

PR #412, opened 2023-10-12
Add target `ConnectionPoolModule`

fabianfett:
We want to land a new ConnectionPool into PostgresNIO in the comming weeks. Since this pool is abstract, let's create a target and product for it. The target and product are both underscored, to signal that we don't make any API stability guarantees.


Jerry-Carter:
This change caught my attention.  I'm curious, how does this relate to [Issue 256](https://github.com/vapor/postgres-nio/issues/256)?  Specifically, does this mean that one or more implementations of a PostgreSQL connection pool will be included with postgres-nio in future?  Likewise, will there continue to be support for developers writing their own custom connection pools as is required today?

Thanks.

fabianfett:
> This change caught my attention. I'm curious, how does this relate to #256? 
This PR means work on #256 has started.

> Specifically, does this mean that one or more implementations of a PostgreSQL connection pool will be included with postgres-nio in future? 
Yes we intend to add one implementation directly to PostgresNIO.

> Likewise, will there continue to be support for developers writing their own custom connection pools as is required today?

We don't intend to break any adopter. However I'd be curious to learn about your use-case... Maybe the new pool is able to fulfill cater to your needs directly?

gwynne (approved):

PR #423, opened 2023-10-18
Fix `PostgresDecodable` inference for `RawRepresentable` enums

MahdiBM:
The function in extension responsible for making `PostgresDecodable` work with `RawRepresentable`, was not marked as public.

I added an integration test to make sure this now functions as expected.

I also removed the `@testable` in the `IntegrationsTests` since it wasn't needed and it would also manipulate Codable behaviors. If there is an internal extension for Codable, @testable makes it so the tests will use that while the `public` will not be able to use that, which will either result in unexpected compile errors for users, or different behavior in test vs what users see.


gwynne:
Looks fine to me, but I defer to @fabianfett in case this wasn't accidental.

fabianfett (approved):
Great catch! Thanks for the patch!

PR #439, opened 2023-11-12
Test cancel connection request

fabianfett:
- Increase test coverage

### DIFF
--- a/Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
+++ b/Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
@@ -368,6 +368,65 @@ final class ConnectionPoolTests: XCTestCase {
         }
     }
 
-}
+    func testCancelConnectionRequestWorks() async throws {
+        let clock = MockClock()
+        let factory = MockConnectionFactory<MockClock>()
+        let keepAliveDuration = Duration.seconds(30)
+        let keepAlive = MockPingPongBehavior(keepAliveFrequency: keepAliveDuration, connectionType: MockConnection.self)
 
+        var mutableConfig = ConnectionPoolConfiguration()
+        mutableConfig.minimumConnectionCount = 0
+        mutableConfig.maximumConnectionSoftLimit = 4
+        mutableConfig.maximumConnectionHardLimit = 4
+        mutableConfig.idleTimeout = .seconds(10)
+        let config = mutableConfig
+
+        let pool = ConnectionPool(
+            configuration: config,
+            idGenerator: ConnectionIDGenerator(),
+            requestType: ConnectionRequest<MockConnection>.self,
+            keepAliveBehavior: keepAlive,
+            observabilityDelegate: NoOpConnectionPoolMetrics(connectionIDType: MockConnection.ID.self),
+            clock: clock
+        ) {
+            try await factory.makeConnection(id: $0, for: $1)
+        }
+
+        try await withThrowingTaskGroup(of: Void.self) { taskGroup in
+            taskGroup.addTask {
+                await pool.run()
+            }
+
+            let leaseTask = Task {
+                do {
+                    _ = try await pool.leaseConnection()
+                    XCTFail("Did not expect call to succeed")
+                } catch {
+                    XCTAssertEqual((error as? ConnectionPoolError), .requestCancelled)
+                }
+            }
+
+            let connectionAttemptWaiter = Waiter(of: Void.self)
+
+            taskGroup.addTask {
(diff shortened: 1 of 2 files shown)


Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
```diff
@@ -368,6 +368,65 @@ final class ConnectionPoolTests: XCTestCase {
...
+            taskGroup.addTask {
+                await pool.run()
+            }
+
+            let leaseTask = Task {
+                do {
+                    _ = try await pool.leaseConnection()
+                    XCTFail("Did not expect call to succeed")
+                } catch {
+                    XCTAssertEqual((error as? ConnectionPoolError), .requestCancelled)
```
gwynne: Haven't the XCTAssert...Async utilities (either my version or NIO's (mine's better, of course)) shown up by now, instead of these awkward inexpressive-of-the-actual-intent constructs? (I mean, `swift-testing`'s gonna utterly blast this sorta thing outta the water anyway, of course...) Here, have the full set: https://gist.github.com/gwynne/2fbbc57eaad942368ea87dd0d2e6628e

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
```diff
@@ -368,6 +368,65 @@ final class ConnectionPoolTests: XCTestCase {
...
+                }
+            }
+
+            let connectionAttemptWaiter = Waiter(of: Void.self)
+
+            taskGroup.addTask {
+                try await factory.nextConnectAttempt { connectionID in
+                    connectionAttemptWaiter.yield(value: ())
```
gwynne, line 408: Does the runtime actually guarantee that at least one waiter (if any exist) will be awoke before the cancellation error is thrown? (Asking more for my own clarification than out of any particular concern for the test's functionality.)

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
```diff
@@ -368,6 +368,65 @@ final class ConnectionPoolTests: XCTestCase {
...
+                }
+            }
+
+            try await connectionAttemptWaiter.result
+            leaseTask.cancel()
+
+            await leaseTask.result
+
+            taskGroup.cancelAll()
+            print("cancelled")
```
gwynne, line 425: Stray `print()` in a test.

Tests/ConnectionPoolModuleTests/Utils/Waiter.swift
```diff
@@ -0,0 +1,105 @@
...
+                }
+
+                cont?.resume(throwing: CancellationError())
+            }
+        }
+    }
+
+    func yield(value: Result) {
+        let continuations = self.stateBox.withLockedValue { state in
+            guard state.result == nil else { return [].lazy.map(\.1) }
```
gwynne: `return [].lazy.map(\.1)`? I'll bite - what does this do? 😳

gwynne (changes requested):
Only actual change needed is get rid of that stray `print()`, though wouldn't mind thoughts on the other comments as well.

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
gwynne: Not sure this comment still applies after your revisions, but I'll leave it here just in case.

Tests/ConnectionPoolModuleTests/Utils/Waiter.swift
gwynne: Okay, the code's changed, but the updated code still does essentially the same thing, so the question stands 🙂

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
fabianfett: The connection pool shall not have any weird dependencies. So we need to dance here.

Tests/ConnectionPoolModuleTests/Utils/Waiter.swift
fabianfett: map to the required return type `LazyMapSequence<LazySeqence<Array<bla>>>`

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
fabianfett: Nope. Works very much like a ELF. If it is already succeeded will just run through.

gwynne (approved):

Issue #443, opened 2023-12-11
Connection Pool Crash: Precondition failure when ping and close happen at the same time


lovetodream:
### Describe the bug

When running a connection pool using PostgresClient, the pool crashes with a precondition failure ("All connections that have been created should say goodbye exactly once!") on certain cases. 

This seems to happen if the keep alive is triggered at the same time as the close on a connection. 

### To Reproduce

1. In the `PostgresClient.Configuration`, set the `connectionIdleTimeout` to a multiple of `keepAliveBehavior?.frequency`.
2. Repeat running a query shortly after `connectionIdleTimeout` until it crashes. (The crash doesn't happen every time)

I've created a minimal reproduction sample: https://github.com/lovetodream/connection-pool-crash 
Note: the crash might not occur on every run, but for me it did (mostly around the ~5th connection)

Here is a log dump, note the second ping pong on the crashing connection:
```
2023-12-11T13:55:20+0100 debug connection-pool-crash : psql_connection_id=0 [PostgresNIO] Creating new connection
2023-12-11T13:55:30+0100 debug connection-pool-crash : psql_connection_id=0 [PostgresNIO] run ping pong
2023-12-11T13:55:41+0100 debug connection-pool-crash : psql_connection_id=0 [PostgresNIO] Close connection
2023-12-11T13:55:41+0100 debug connection-pool-crash : psql_connection_id=0 psql_error=PSQLError – Generic description to prevent accidental leakage of sensitive data. For debugging details, use `String(reflecting: error)`. [PostgresNIO] Cleaning up and closing connection.
2023-12-11T13:55:41+0100 debug connection-pool-crash : psql_connection_id=0 psql_error=ioOnClosedChannel [PostgresNIO] Channel error caught.
2023-12-11T13:55:41+0100 debug connection-pool-crash : psql_connection_id=0 [PostgresNIO] Connection closed
2023-12-11T13:55:43+0100 debug connection-pool-crash : psql_connection_id=1 [PostgresNIO] Creating new connection
2023-12-11T13:55:54+0100 debug connection-pool-crash : psql_connection_id=1 [PostgresNIO] run ping pong
2023-12-11T13:56:04+0100 debug connection-pool-crash : psql_connection_id=1 [PostgresNIO] Close connection
[...]

PR #444, opened 2023-12-11
Fix crash in PoolStateMachine+ConnectionGroup when closing connection while keepAlive is running

lovetodream:
Fixes: #443

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -158,9 +158,10 @@ jobs:
         run: |
           export PATH="$(brew --prefix)/opt/${POSTGRES_FORMULA}/bin:$PATH" PGDATA=/tmp/vapor-postgres-test
           brew install --overwrite "${POSTGRES_FORMULA}"
+          brew link --overwrite --force "${POSTGRES_FORMULA}"
           initdb --locale=C --auth-host "${POSTGRES_AUTH_METHOD}" -U "${POSTGRES_USER}" --pwfile=<(echo "${POSTGRES_PASSWORD}")
           pg_ctl start --wait
-        timeout-minutes: 2
+        timeout-minutes: 15
       - name: Checkout code
         uses: actions/checkout@v4
       - name: Run all tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
           - swift:5.8-jammy
           - swift:5.9-jammy
           - swiftlang/swift:nightly-5.10-jammy
-          #- swiftlang/swift:nightly-main-jammy
+          - swiftlang/swift:nightly-main-jammy
         include:
           - swift-image: swift:5.9-jammy
             code-coverage: true
@@ -133,7 +133,7 @@ jobs:
       matrix:
         postgres-formula:
           # Only test one version on macOS, let Linux do the rest
-          - postgresql@15
+          - postgresql@16
         postgres-auth:
           # Only test one auth method on macOS, Linux tests will cover the others
           - scram-sha-256
@@ -157,7 +157,7 @@ jobs:
       - name: Install Postgres, setup DB and auth, and wait for server start
         run: |
           export PATH="$(brew --prefix)/opt/${POSTGRES_FORMULA}/bin:$PATH" PGDATA=/tmp/vapor-postgres-test
-          (brew unlink postgresql || true) && brew install "${POSTGRES_FORMULA}" && brew link --force "${POSTGRES_FORMULA}"
+          brew install --overwrite "${POSTGRES_FORMULA}"
           initdb --locale=C --auth-host "${POSTGRES_AUTH_METHOD}" -U "${POSTGRES_USER}" --pwfile=<(echo "${POSTGRES_PASSWORD}")
           pg_ctl start --wait
         timeout-minutes: 2
@@ -183,7 +183,7 @@ jobs:
 
   gh-codeql:
     runs-on: ubuntu-latest
-    container: swift:5.8-jammy # CodeQL currently broken with 5.9
+    container: swift:5.9-jammy
     permissions: { actions: write, contents: read, security-events: write }
     steps:
       - name: Check out code
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -157,7 +157,13 @@ jobs:
       - name: Install Postgres, setup DB and auth, and wait for server start
         run: |
           export PATH="$(brew --prefix)/opt/${POSTGRES_FORMULA}/bin:$PATH" PGDATA=/tmp/vapor-postgres-test
-          (brew unlink postgresql || true) && brew install "${POSTGRES_FORMULA}" && brew link --force "${POSTGRES_FORMULA}"
+         # ** BEGIN ** Work around bug in both Homebrew and GHA
+          (brew upgrade python@3.11 || true) && (brew link --force --overwrite python@3.11 || true)
+          (brew upgrade python@3.12 || true) && (brew link --force --overwrite python@3.12 || true)
+          brew upgrade
+          # ** END ** Work around bug in both Homebrew and GHA
+          brew install --overwrite "${POSTGRES_FORMULA}"
+          brew link --overwrite --force "${POSTGRES_FORMULA}"
           initdb --locale=C --auth-host "${POSTGRES_AUTH_METHOD}" -U "${POSTGRES_USER}" --pwfile=<(echo "${POSTGRES_PASSWORD}")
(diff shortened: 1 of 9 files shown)


Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -449,6 +449,21 @@ extension PoolStateMachine {
             return (index, context)
         }
 
+        /// Returns `true` if the connection should be explicitly closed, `false` if nothing needs to be done.
+        @inlinable
+        mutating func keepAliveFailed(_ connectionID: Connection.ID) -> Bool {
```
fabianfett, line 454: Can we return an enum for this? KeepAliveFailedAction with two cases: close and none? Makes this more explicit

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -449,6 +449,21 @@ extension PoolStateMachine {
...
+            // We don't have to inform the ConnectionStates about any of this, because the connection will close
+            // immediately after this or is closed already
+            switch self.connections.first(where: { $0.id == connectionID })?.state {
```
fabianfett, line 457: please don‘t access the connection‘s state here. the state is only visible here, to allow inlining. Instead please forward a call keepAliveFailed into the ConnectionStateMachine. Also make sure to change the state to closing, if we need to explicitly close the connection.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -547,8 +562,9 @@ extension PoolStateMachine {
 
             if closedAction.wasRunningKeepAlive {
                 self.stats.runningKeepAlive -= 1
+            } else {
```
fabianfett, line 565: this change looks wrong, do you mind explaining, why you changed this one?

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -374,6 +374,14 @@ struct PoolStateMachine<
         return self.handleAvailableConnection(index: index, availableContext: context)
     }
 
+    @inlinable
+    mutating func connectionKeepAliveFailed(_ connection: Connection) -> Action {
+        if self.connections.keepAliveFailed(connection.id) {
+            return self.connectionClosed(connection)
```
fabianfett, line 380: That doesn‘t look right, if we the keep alive failed, we can‘t declare the connection dead right away, we need to create a close action. Also if we close and have waiting requests, we might need to create a new connection right away.

Tests/ConnectionPoolModuleTests/Mocks/MockConnection.swift
```diff
@@ -97,6 +97,17 @@ final class MockConnection: PooledConnection, Sendable {
...
         }
     }
+
+    var isClosing: Bool {
+        self.lock.withLockedValue { state in
+            switch state {
+            case .closing:
+                return true
+            case .running, .closed:
+                return true
```
fabianfett, line 107: this looks wrong ;)

fabianfett (commented):
Great progress! However we need to go a bit deeper at some points.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
lovetodream: Yeah, that's wrong. I did this initially because a test I've added force closed the connection without going the proper route. This caused a arythmic overflow due to leasedStreams being 0 at that point in time. I've fixed the test and reverted the change.

Tests/ConnectionPoolModuleTests/Mocks/MockConnection.swift
lovetodream: Of course it is :), I've removed it as it's no longer needed

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
lovetodream: The connection state is receiving the failure now and changes the state accordingly.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
lovetodream: This is now returning an optional CloseAction. It should be explicit enough now, let me know if not.

Sources/ConnectionPoolModule/PoolStateMachine.swift
lovetodream: The connection is now properly closed by reusing the existing closed states and actions. I've added a test to ensure a new connection is established if required.

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
```diff
@@ -300,6 +300,92 @@ final class ConnectionPoolTests: XCTestCase {
...
+
+            let deadline3 = await clock.nextTimerScheduled()
+            XCTAssertEqual(deadline3, clock.now.advanced(by: keepAliveDuration))
+            print(deadline3)
+
+            clock.advance(to: clock.now.advanced(by: keepAliveDuration))
+
+            // the following keep alive should not cause a crash
+            _ = try? await keepAlive.nextKeepAlive { keepAliveConnection in
+                defer { print("failing keep alive has run") }
```
fabianfett, line 375: wanna ensure this by using an atomic?

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
```diff
@@ -300,6 +300,92 @@ final class ConnectionPoolTests: XCTestCase {
...
+                defer { print("keep alive 1 has run") }
+                XCTAssertTrue(keepAliveConnection === lease1Connection)
+                return true
+            }
+
+            // keep alive 2
+
+            let deadline3 = await clock.nextTimerScheduled()
+            XCTAssertEqual(deadline3, clock.now.advanced(by: keepAliveDuration))
+            print(deadline3)
```
fabianfett, line 369: wanna remove the `print`s here?

fabianfett (commented):
Superb PR. Thanks so much for looking into this!

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
```diff
@@ -354,29 +354,26 @@ final class ConnectionPoolTests: XCTestCase {
...
             XCTAssertEqual(deadline3, clock.now.advanced(by: keepAliveDuration))
-            print(deadline3)
-
             clock.advance(to: clock.now.advanced(by: keepAliveDuration))
 
+            let failingKeepAliveDidRun = ManagedAtomic(false)
             // the following keep alive should not cause a crash
             _ = try? await keepAlive.nextKeepAlive { keepAliveConnection in
-                defer { print("failing keep alive has run") }
+                defer { failingKeepAliveDidRun.store(true, ordering: .relaxed) }
```
fabianfett, line 371:
```suggestion
                defer { 
                    XCTAssertFalse(failingKeepAliveDidRun.store(true, ordering: .relaxed).original)
                }
```

fabianfett (approved):

.github/workflows/test.yml
```diff
@@ -157,10 +157,14 @@ jobs:
       - name: Install Postgres, setup DB and auth, and wait for server start
         run: |
           export PATH="$(brew --prefix)/opt/${POSTGRES_FORMULA}/bin:$PATH" PGDATA=/tmp/vapor-postgres-test
-          (brew unlink postgresql || true) && brew install "${POSTGRES_FORMULA}" && brew link --force "${POSTGRES_FORMULA}"
+          # ** BEGIN ** Work around bug in both Homebrew and GHA
+          (brew upgrade python@3.12 || true) && (brew link --force --overwrite python@3.12 || true) && brew upgrade
+          # ** END ** Work around bug in both Homebrew and GHA
+          brew install --overwrite "${POSTGRES_FORMULA}"
+          brew link --overwrite --force "${POSTGRES_FORMULA}"
```
fabianfett, line 164: Let's revert those changes for now
```suggestion
          (brew unlink postgresql || true) && brew install "${POSTGRES_FORMULA}" && brew link --force "${POSTGRES_FORMULA}"
```

fabianfett:
Thanks @lovetodream. Great catch and great fix!

PR #450, opened 2024-01-28
Add Sendable conformance to PostgresEncodingContext

gwynne:
Like it says.

Technically should probably be semver-minor, but doesn't really seem worth it.


fabianfett (approved):
LGTM

PR #451, opened 2024-01-30
Fix mishandling of SASL attribute parsing

gwynne:
When parsing SASL attribute strings, we were incorrectly allowing them to be split on `=` characters more than once per attribute. This broke for attribute values ending with `=`, most particularly Base64-like values, which appear often in - for example - the nonce strings sent by Supabase PostgreSQL severs.

### DIFF
--- a/Sources/PostgresNIO/Utilities/SASLAuthentication+SCRAM-SHA256.swift
+++ b/Sources/PostgresNIO/Utilities/SASLAuthentication+SCRAM-SHA256.swift
@@ -209,14 +209,13 @@ fileprivate struct SCRAMMessageParser {
     }
     
     static func parse(raw: [UInt8], isGS2Header: Bool = false) -> [SCRAMAttribute]? {
-        
         // There are two ways to implement this parse:
         //  1. All-at-once: Split on comma, split each on equals, validate
         //     each results in a valid attribute.
         //  2. Sequential: State machine lookahead parse.
         // The former is simpler. The latter provides better validation.
-        let likelyAttributeSets = raw.split(separator: .comma, maxSplits: isGS2Header ? 3 : Int.max, omittingEmptySubsequences: false)
-        let likelyAttributePairs = likelyAttributeSets.map { $0.split(separator: .equals, maxSplits: 2, omittingEmptySubsequences: false) }
+        let likelyAttributeSets = raw.split(separator: .comma, maxSplits: isGS2Header ? 2 : Int.max, omittingEmptySubsequences: false)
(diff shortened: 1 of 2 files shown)


Sources/PostgresNIO/Utilities/SASLAuthentication+SCRAM-SHA256.swift
```diff
@@ -209,14 +209,13 @@ fileprivate struct SCRAMMessageParser {
...
     static func parse(raw: [UInt8], isGS2Header: Bool = false) -> [SCRAMAttribute]? {
-        
         // There are two ways to implement this parse:
         //  1. All-at-once: Split on comma, split each on equals, validate
         //     each results in a valid attribute.
         //  2. Sequential: State machine lookahead parse.
         // The former is simpler. The latter provides better validation.
-        let likelyAttributeSets = raw.split(separator: .comma, maxSplits: isGS2Header ? 3 : Int.max, omittingEmptySubsequences: false)
-        let likelyAttributePairs = likelyAttributeSets.map { $0.split(separator: .equals, maxSplits: 2, omittingEmptySubsequences: false) }
+        let likelyAttributeSets = raw.split(separator: .comma, maxSplits: isGS2Header ? 2 : Int.max, omittingEmptySubsequences: false)
```
fabianfett, line 217: why did we change this one?

Sources/PostgresNIO/Utilities/SASLAuthentication+SCRAM-SHA256.swift
```diff
@@ -209,14 +209,13 @@ fileprivate struct SCRAMMessageParser {
...
+        let likelyAttributePairs = likelyAttributeSets.map { $0.split(separator: .equals, maxSplits: 1, omittingEmptySubsequences: false) }
```
fabianfett, line 218: can we add a comment with an example here?

fabianfett (approved):
Two tiny nits! Extremely good work! Thanks so much! Love the increase in test coverage as well!

Sources/PostgresNIO/Utilities/SASLAuthentication+SCRAM-SHA256.swift
gwynne: Because it was also wrong.

Sources/PostgresNIO/Utilities/SASLAuthentication+SCRAM-SHA256.swift
gwynne: What example? I forever and a day ago misremembered how `maxSplits` works. It just happened to be one of the times where the problem didn't actually manifest for a long time.

fabianfett (approved):

PR #452, opened 2024-01-31
Be resilient about a read after connection closed; fixes #449

### LINKED ISSUE #449: Fatal error: How can we receive a read, if the connection is closed
### Describe the bug

Getting a fatal error bug when connected to neon postgresQL.

### To Reproduce

Neon (https://neon.tech) is a fully managed postgresQL service. The following simple code produces this fatal error and crashes when connected to Neon, it doesn't crash when connected to a locally hosted db. Neon uses some new connection options which apparently Postgres-nio doesn't handle well.

The code:
```
let config = PostgresConnection.Configuration(
    host: "<your neon host>",
    port: 5432,
    username: "<username>",
    password: "<password>",
    database: "neondb",
    tls: .require(try NIOSSLContext(configuration: TLSConfiguration.makeClientConfiguration()))
  )
  
  let connection = try await PostgresConnection.connect(
    configuration: config,
    id: 1,
    logger: Logger(label: "log")
  )
  
  try await connection.close() // <- here it crashes with above Fatal error: How can we receive a read, if the connection is closed
````

### Expected behavior

The connection should close cleanly.

### Environment

• So far was only able re reproduce when connected to Neon.
- Any OS, Mac or Linux.
- You can register on neon for free and receive a free account to test this on.

fabianfett:


gwynne (approved):
❤️❤️❤️

PR #454, opened 2024-02-21
Fix warnings

fabianfett:
Fix warnings:

1. Use non deprecated `NIOThrowingAsyncSequenceProducer` factory method
2. Remove `UInt8`'s conformance to `ExpressibleByUnicodeScalarLiteral`
3. Fix @retroactive conformance issues in test by fully specifying all types


fabianfett:
Breaking changes:
```
3 breaking changes detected in PostgresNIO:
  💔 API breakage: extension UInt8 has removed conformance to ExpressibleByUnicodeScalarLiteral
  💔 API breakage: constructor UInt8.init(unicodeScalarLiteral:) has been removed
  💔 API breakage: typealias UInt8.UnicodeScalarLiteralType has been removed
```

While those are actual breaking changes, it was 100% wrong that this is part of our public API and noone should have ever relied on this! @gwynne are you fine with deciding that this doesn't require a major version?

gwynne (approved):

gwynne:
@fabianfett The API breakage checker is more trouble than it's worth most of the time, the false positive rate is astronomical.

fabianfett:
@gwynne this _is not_ a false positive! We are really removing public API here.

PR #465, opened 2024-03-08
Fix `reverseChunked(by:)` Method Implementation

jiahan-wu:
# Summary

This pull request is being made to resolve an inconsistency found in the reverseChunked(by:) method located within the Collection extension. The intended functionality of the method is to divide a collection into chunks of a given size, where, if the collection is not evenly divisible, the first chunk will be smaller. Instead of this expected behavior, the current implementation produces incorrect results, with the first chunk not being the smallest when required and subsequent chunks varying incorrectly in size.

# Example

Here are concrete examples that demonstrate the problem with the reverseChunked(by:) method's current implementation:

Example 1:
```swift
let collection = "1234567898765"
let chunks = collection.reverseChunked(by: 4)

// Expected Chunks: ["1", "2345", "6789", "8765"]
// Actual Chunks:   ["1", "2345", "67898", "765"]
```

Example 2:
```swift
let collection = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17]
let chunks = collection.reverseChunked(by: 4)

// Expected Chunks: [[1], [2, 3, 4, 5], [6, 7, 8, 9], [10, 11, 12, 13], [14, 15, 16, 17]]
// Actual Chunks:   [[1], [2, 3, 4, 5], [6, 7, 8, 9, 10], [11, 12, 13, 14, 15, 16, 17], []]
```

# Impact of Changes

The changes introduced by this pull request will resolve the crashing issue in the [PostgresNumeric(decimalString:)](https://github.com/vapor/postgres-nio/blob/dc9caf85c1b9c8e6df32884b81147af3ef01e0c5/Sources/PostgresNIO/Data/PostgresData%2BNumeric.swift#L56) initializer. The crash is currently caused by the improper handling of string to numeric conversion when the string is split into chunks using the reverseChunked(by:) method.

Screenshot 2024-03-08 at 10 10 48 PM


fabianfett (approved):

gwynne (approved):
At some point we should really just replace those methods altogether with an import of `Algorithms` and use of `.chunks(ofCount:).reversed()`, but for now this is fine.

fabianfett:
@jiahan-wu Thanks for your contribution! Great fix!

PR #466, opened 2024-03-18
Fixes `LISTEN` to quote channel name

NeedleInAJayStack:
This adjusts the SQL `LISTEN` statement called by `PostgresConnection.addListener` to quote the channel name.

For example, previously `addListener("abc")` would run `LISTEN abc;` and it will now run `LISTEN "abc";`

As far as I can tell, this only matters when the channel name is a reserved Postgres keyword, like `default`.

### DIFF
--- a/Sources/PostgresNIO/New/PostgresChannelHandler.swift
+++ b/Sources/PostgresNIO/New/PostgresChannelHandler.swift
@@ -594,7 +594,7 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
     private func makeStartListeningQuery(channel: String, context: ChannelHandlerContext) -> PSQLTask {
         let promise = context.eventLoop.makePromise(of: PSQLRowStream.self)
         let query = ExtendedQueryContext(
-            query: PostgresQuery(unsafeSQL: "LISTEN \(channel);"),
+            query: PostgresQuery(unsafeSQL: #"LISTEN "\#(channel)";"#),
             logger: self.logger,
             promise: promise
         )
@@ -642,7 +642,7 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
     private func makeUnlistenQuery(channel: String, context: ChannelHandlerContext) -> PSQLTask {
         let promise = context.eventLoop.makePromise(of: PSQLRowStream.self)
         let query = ExtendedQueryContext(
-            query: PostgresQuery(unsafeSQL: "UNLISTEN \(channel);"),
+            query: PostgresQuery(unsafeSQL: #"UNLISTEN "\#(channel)";"#),
             logger: self.logger,
             promise: promise
         )
--- a/Tests/IntegrationTests/AsyncTests.swift
+++ b/Tests/IntegrationTests/AsyncTests.swift
@@ -225,25 +225,32 @@ final class AsyncPostgresConnectionTests: XCTestCase {
     }
 
     func testListenAndNotify() async throws {
+        let channelNames = [
+            "foo",
+            "default"
+        ]
+        
         let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
         defer { XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully()) }
         let eventLoop = eventLoopGroup.next()
 
-        try await self.withTestConnection(on: eventLoop) { connection in
-            let stream = try await connection.listen("foo")
-            var iterator = stream.makeAsyncIterator()
+        for channelName in channelNames {
+            try await self.withTestConnection(on: eventLoop) { connection in
+                let stream = try await connection.listen(channelName)
+                var iterator = stream.makeAsyncIterator()
 
-            try await self.withTestConnection(on: eventLoop) { other in
-                try await other.query(#"NOTIFY foo, 'bar';"#, logger: .psqlTest)
+                try await self.withTestConnection(on: eventLoop) { other in
+                    try await other.query(#"NOTIFY "\#(unescaped: channelName)", 'bar';"#, logger: .psqlTest)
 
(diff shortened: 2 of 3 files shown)


NeedleInAJayStack:
This behavior appears to have been introduced in v1.18.0. Interestingly, by watching SQL logs this change it appears that version also changed `addListener` to call `LISTEN` automatically, so some of the notification tests are calling `LISTEN` twice.

Tests/IntegrationTests/PostgresNIOTests.swift
```diff
@@ -227,6 +227,23 @@ final class PostgresNIOTests: XCTestCase {
...
+        conn?.addListener(channel: "default") { context, notification in
+            receivedNotifications.append(notification)
+        }
+        XCTAssertNoThrow(_ = try conn?.simpleQuery("LISTEN \"default\"").wait())
+        XCTAssertNoThrow(_ = try conn?.simpleQuery("NOTIFY \"default\"").wait())
+        XCTAssertNoThrow(_ = try conn?.simpleQuery("SELECT 1").wait())
+        XCTAssertEqual(receivedNotifications.count, 1)
+        XCTAssertEqual(receivedNotifications.first?.channel, "default")
+        XCTAssertEqual(receivedNotifications.first?.payload, "")
+    }
```
fabianfett, line 246: How does this test interact with the code that you changed? Maybe we want to change `AsyncPostgresConnectionTests.testListenAndNotify` to something like this:

```swift
    func testListenAndNotify() async throws {
        let channelNames = [
            "foo",
            "default"
        ]

        defer { XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully()) }
        let eventLoop = eventLoopGroup.next()

        for channelName in channelNames {
            try await self.withTestConnection(on: eventLoop) { connection in
                let stream = try await connection.listen(channelName)
                var iterator = stream.makeAsyncIterator()

                try await self.withTestConnection(on: eventLoop) { other in
                    try await other.query(#"NOTIFY "\#(unescaped: channelName)", 'bar';"#, logger: .psqlTest)

                    try await other.query(#"NOTIFY "\#(unescaped: channelName)", 'foo';"#, logger: .psqlTest)
                }

                let first = try await iterator.next()
                XCTAssertEqual(first?.payload, "bar")

                let second = try await iterator.next()
                XCTAssertEqual(second?.payload, "foo")
            }
        }
    }
```

fabianfett (commented):
Nice catch! See comments!

fabianfett:
> This behavior appears to have been introduced in v1.18.0. Interestingly, by watching SQL logs it appears that version also changed `addListener` to call `LISTEN` automatically, so some of the notification tests are calling `LISTEN` twice.

Does this mean you actually used the old `LISTEN` API?

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -594,7 +594,7 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
     private func makeStartListeningQuery(channel: String, context: ChannelHandlerContext) -> PSQLTask {
         let promise = context.eventLoop.makePromise(of: PSQLRowStream.self)
         let query = ExtendedQueryContext(
-            query: PostgresQuery(unsafeSQL: "LISTEN \(channel);"),
+            query: PostgresQuery(unsafeSQL: "LISTEN \"\(channel)\";"),
```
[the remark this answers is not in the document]
NeedleInAJayStack: Done, thanks!

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -642,7 +642,7 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
     private func makeUnlistenQuery(channel: String, context: ChannelHandlerContext) -> PSQLTask {
         let promise = context.eventLoop.makePromise(of: PSQLRowStream.self)
         let query = ExtendedQueryContext(
-            query: PostgresQuery(unsafeSQL: "UNLISTEN \(channel);"),
+            query: PostgresQuery(unsafeSQL: "UNLISTEN \"\(channel)\";"),
```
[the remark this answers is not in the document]
NeedleInAJayStack: Good call. Done

Tests/IntegrationTests/PostgresNIOTests.swift
NeedleInAJayStack: My change was related to the quoting that occurs in the `LISTEN` SQL query caused by `addListener`. So before the changes, line 237 of this new test would fail on the PostgreSQL side when running `LISTEN default;` with the error:

> ERROR:  syntax error at or near "default" at character 8
> STATEMENT:  LISTEN default;

NeedleInAJayStack: I'm totally happy to combine it into `testListenAndNotify` though if you prefer it that way.

NeedleInAJayStack:
> Does this mean you actually used the old LISTEN API?

Yeah, we were using the old LISTEN API. However, my comment was more related to [these tests](https://github.com/vapor/postgres-nio/blob/6f0fc054babeed13850f9014e03ced7a1d714868/Tests/IntegrationTests/PostgresNIOTests.swift#L119), and specifically that the lines with `conn?.simpleQuery("LISTEN example").wait()` are technically no longer needed since `addListener` now calls `LISTEN example;`. When watching the SQL logs I could see two `LISTEN example` queries in a row.

Tests/IntegrationTests/PostgresNIOTests.swift
NeedleInAJayStack: Cool, I've made your recommended changes to `testListenAndNotify` and removed the `testNotificationWithKeywordChannelName` test

PR #471, opened 2024-05-01
Make `TLS.disable` a let instead of a var

FranzBusch:
This currently emits a Sendable warning since a global var isn't Sendable safe.


fabianfett (approved):
Great catch!

gwynne (approved):
I blame @fabianfett for making me leave a bunch of other config things `var` instead of `let` 🤣

PR #474, opened 2024-05-09
Fix Connection Pool crash when requests are waiting while timer is triggered

### LINKED ISSUE #472: ConnectionPool crash on keepalive with non-empty request queue
### Describe the issue

The connection pool crashes when a lot of requests occur in a burst

### Vapor version

n/a

macOS 14.4.1

### Swift version

Swift Package Manager - Swift 5.10.0-dev

### Steps to reproduce

Run the following script:
```swift
let config = PostgresClient.Configuration(...)
var logger = Logger(label: "connection-pool-crash")
logger.logLevel = .debug

let client = PostgresClient(
    configuration: config,
    backgroundLogger: logger
)
let task = Task {
    await client.run()
}

func runOnce() async throws {
    let logger = logger
    try await withThrowingDiscardingTaskGroup { group in
        for _ in 0..<20 {
            group.addTask {
                let hello = try await client.withConnection { db in
                    try await db.query("SELECT 'hello'", logger: logger)
                }.collect().first?.decode(String.self)
                print(hello)

                // wait for the connection pool to do ping pong and close
                try await Task.sleep(for: .seconds(idleTimeout + 1))
            }
        }
    }
}

for i in 0..<50 {
    try await runOnce()
    print("Burst done: \(i + 1)/50")
}

task.cancel()
```

### Outcome

A crash happens (typically on the start of the second burst, after running the script for approx. 1.5 min), the crash occurs in `PoolStateMachine.connectionKeepAliceTimerTriggered(_:)` due to a non-empty request queue. My assumption is, that there's a race condition when the timer triggers the ping and the next batch of requests is "inserted" into the pool. 

### Additional notes

_No response_

lovetodream:
When connections are requested from the pool in bursts, the pool might have requests in the queue while a keep alive or idle timer is triggered. This should resolve the issue, closes #472.

Currently, I've only implemented a test case to reproduce the behaviour. (The test case is a bit verbose atm, but I wanted to include the full picture for now)
One way to fix it would be to check the request queue in `PoolStateMachine.connectionKeepAliveTimerTriggered(_:)` and `PoolStateMachine.connectionIdleTimerTriggered(_:)` instead of a precondition. This can be archived by calling `PoolStateMachine.handleAvailableConnection(index:availableContext:)` if the queue is not empty.
Maybe this should be handled in `ConnectionPool` instead?

Let me know what you think @fabianfett


ptoffy:
Hey @lovetodream I've tried this out locally now with the latest main version and Swift Testing and the test seems to pass. Can you confirm this isn't an issue anymore and that we can close this?

lovetodream:
I can confirm this is no longer an issue. Thanks for taking care of it! I'm closing it.

PR #478, opened 2024-05-29
workaround DiscardingTaskGroup non-conformance with nightly compilers

weissi:
Some nightly Swift compilers complain that `DiscardingTaskGroup` doesn't conform to `TaskGroupProtocol`. This patch works around it.


gwynne:
@weissi It seems according to the CI that this not only breaks 5.8+, it doesn't actually solve the error (despite the delightfully ironic "candidate exactly matches" message in the error spew 😆)

weissi:
> @weissi It seems according to the CI that this not only breaks 5.8+, it doesn't actually solve the error (despite the delightfully ironic "candidate exactly matches" message in the error spew 😆)

argh, seems to depend on the compiler & SDK version :P.

weissi:
@gwynne better now

fabianfett (approved):

Issue #458, opened 2024-02-21
PostgresNIO/ListenStateMachine.swift:182: Fatal error: Invalid state: initialized


CrownedPhoenix:
### Description

ListenStateMachine reaches an invalid state during concurrent operation of `addListener` on a pool that is shutting down.
I'm not sure if this behavior is expected given the obvious concurrent execution described in the reproducer below.

At any rate, a fatalError here seems out of place unless there's something I'm misunderstanding about the ListenStateMachine.

### To Reproduce

Simply run the following:
```swift
import NIO
import AsyncKit
import PostgresNIO
import PostgresKit

let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)

let configuration = SQLPostgresConfiguration(
    hostname: "localhost",
    port: 5432,
    username: "postgres",
    password: "postgres",
    database: "test",
    tls: .disable
)
for _ in 0..<1000 {
    let pool = EventLoopGroupConnectionPool(
        source: PostgresConnectionSource(sqlConfiguration: configuration),
        on:  group
    )

    Task {
        for _ in 0..<100 {
            _ = pool.withConnection({ conn in
                let channel = "default"
                _ = conn.addListener(channel: channel) { _, notifications in
                    assert(notifications.channel == channel, "Notification channel mismatch")
                }

                return group.future()
            })
        }
    }

    let promise: EventLoopPromise<Void> = group.next().makePromise()

    pool.shutdownGracefully({ _ in
        promise.succeed(())

    })
    try promise.futureResult.wait()
}
```
with the following dependencies:
```swift
        .package(url: "https://github.com/apple/swift-nio.git",  from: "2.54.0"),
        .package(url: "https://github.com/vapor/async-kit.git", exact: "1.17.0"),
        .package(url: "https://github.com/vapor/postgres-nio.git", exact: "1.20.0"),
        .package(url: "https://github.com/vapor/postgres-kit.git", exact: "2.11.3"),
```
and the following for the database setup:
```
docker run \
    --detach \
    --name timescaledb \
    --publish 5432:5432 \
    --env POSTGRES_HOST_AUTH_METHOD=trust \
    --env TIMESCALEDB_TELEMETRY=off \
    timescale/timescaledb:latest-pg16;

psql -h localhost -U postgres -w -c 'create database test'
```
the result will be:
```
PostgresNIO/ListenStateMachine.swift:182: Fatal error: Invalid state: initialized
```

### Expected behavior

[...]

PR #487, opened 2024-06-19
[Fix] Query Hangs if Connection is Closed

MahdiBM:
See the test for how to trigger this.

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -426,7 +426,11 @@ extension PostgresConnection {
             promise: promise
         )
 
-        self.channel.write(HandlerTask.extendedQuery(context), promise: nil)
+        let writePromise = self.channel.eventLoop.makePromise(of: Void.self)
+        self.channel.write(HandlerTask.extendedQuery(context), promise: writePromise)
+        writePromise.futureResult.whenFailure { error in
+            promise.fail(error)
+        }
 
         do {
             return try await promise.futureResult.map({ $0.asyncSequence() }).get()
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -222,7 +222,7 @@ public final class PostgresConnection: @unchecked Sendable {
             promise: promise
         )
 
-        self.write(HandlerTask.extendedQuery(context), cascadingFailureTo: promise)
+        self.write(.extendedQuery(context), cascadingFailureTo: promise)
 
         return promise.futureResult
     }
@@ -239,7 +239,7 @@ public final class PostgresConnection: @unchecked Sendable {
             promise: promise
         )
 
-        self.write(HandlerTask.extendedQuery(context), cascadingFailureTo: promise)
+        self.write(.extendedQuery(context), cascadingFailureTo: promise)
 
         return promise.futureResult.map { rowDescription in
             PSQLPreparedStatement(name: name, query: query, connection: self, rowDescription: rowDescription)
@@ -256,7 +256,7 @@ public final class PostgresConnection: @unchecked Sendable {
             logger: logger,
             promise: promise)
 
-        self.write(HandlerTask.extendedQuery(context), cascadingFailureTo: promise)
+        self.write(.extendedQuery(context), cascadingFailureTo: promise)
 
         return promise.futureResult
     }
@@ -265,7 +265,7 @@ public final class PostgresConnection: @unchecked Sendable {
         let promise = self.channel.eventLoop.makePromise(of: Void.self)
         let context = CloseCommandContext(target: target, logger: logger, promise: promise)
 
-        self.write(HandlerTask.closeCommand(context), cascadingFailureTo: promise)
+        self.write(.closeCommand(context), cascadingFailureTo: promise)
 
         return promise.futureResult
     }
@@ -429,7 +429,7 @@ extension PostgresConnection {
             promise: promise
         )
 
-        self.write(HandlerTask.extendedQuery(context), cascadingFailureTo: promise)
+        self.write(.extendedQuery(context), cascadingFailureTo: promise)
 
         do {
             return try await promise.futureResult.map({ $0.asyncSequence() }).get()
@@ -542,7 +542,7 @@ extension PostgresConnection {
         }
     }
 
-    private func write<T, Value>(_ any: T, cascadingFailureTo promise: EventLoopPromise<Value>) {
+    private func write<T>(_ task: HandlerTask, cascadingFailureTo promise: EventLoopPromise<T>) {
         let writePromise = self.channel.eventLoop.makePromise(of: Void.self)
         self.channel.write(any, promise: writePromise)
         writePromise.futureResult.cascadeFailure(to: promise)
@@ -730,7 +730,7 @@ extension PostgresConnection {
             closure: notificationHandler
         )
 
-        self.write(HandlerTask.startListening(listener), cascadingFailureTo: listenContext.promise)
+        self.write(.startListening(listener), cascadingFailureTo: listenContext.promise)
 
         listenContext.future.whenComplete { _ in
             let task = HandlerTask.cancelListening(channel, id)
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -461,6 +461,7 @@ extension PostgresConnection {
                 let writePromise = self.channel.eventLoop.makePromise(of: Void.self)
                 self.channel.write(task, promise: writePromise)
                 writePromise.futureResult.whenFailure { error in
+                    listener.failed(error)
                     continuation.resume(throwing: error)
                 }
             }
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -428,9 +428,7 @@ extension PostgresConnection {
 
         let writePromise = self.channel.eventLoop.makePromise(of: Void.self)
         self.channel.write(HandlerTask.extendedQuery(context), promise: writePromise)
-        writePromise.futureResult.whenFailure { error in
-            promise.fail(error)
-        }
+        writePromise.futureResult.cascadeFailure(to: promise)
 
         do {
             return try await promise.futureResult.map({ $0.asyncSequence() }).get()
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -468,7 +468,11 @@ extension PostgresConnection {
 
                 let task = HandlerTask.startListening(listener)
 
-                self.channel.write(task, promise: nil)
+                let writePromise = self.channel.eventLoop.makePromise(of: Void.self)
+                self.channel.write(task, promise: writePromise)
+                writePromise.futureResult.whenFailure { error in
+                    continuation.resume(throwing: error)
+                }
             }
         } onCancel: {
             let task = HandlerTask.cancelListening(channel, id)
@@ -536,7 +540,7 @@ extension PostgresConnection {
         let writePromise = self.channel.eventLoop.makePromise(of: Void.self)
         self.channel.write(task, promise: writePromise)
         writePromise.futureResult.cascadeFailure(to: promise)
-        
+
         do {
             return try await promise.futureResult
                 .map { $0.commandTag }
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -222,7 +222,9 @@ public final class PostgresConnection: @unchecked Sendable {
             promise: promise
         )
 
-        self.channel.write(HandlerTask.extendedQuery(context), promise: nil)
+        let writePromise = self.channel.eventLoop.makePromise(of: Void.self)
+        self.channel.write(HandlerTask.extendedQuery(context), promise: writePromise)
+        writePromise.futureResult.cascadeFailure(to: promise)
 
         return promise.futureResult
     }
@@ -258,7 +260,10 @@ public final class PostgresConnection: @unchecked Sendable {
             logger: logger,
             promise: promise)
 
-        self.channel.write(HandlerTask.extendedQuery(context), promise: nil)
+        let writePromise = self.channel.eventLoop.makePromise(of: Void.self)
+        self.channel.write(HandlerTask.extendedQuery(context), promise: writePromise)
+        writePromise.futureResult.cascadeFailure(to: promise)
+
         return promise.futureResult
     }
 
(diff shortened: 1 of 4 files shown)


MahdiBM:
Cancelled the integration tests CI manually, since they wouldn't have finished.

Side note @gwynne: probably good to have a time limit in CIs.

gwynne:
> Side note @gwynne: probably good to have a time limit in CIs.

The default timeout is 1 hour, IIRC.

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -360,4 +360,28 @@ final class IntegrationTests: XCTestCase {
...
+        SELECT * FROM non_existent_table
+        """
+
+        _ = try await withThrowingTaskGroup(
+            of: Void.self
+        ) { taskGroup in
+            for _ in (0 ..< 1_000) {
+                taskGroup.addTask {
+                    let conn = try await PostgresConnection.test(
+                        on: NIOSingletons.posixEventLoopGroup.next()
```
MahdiBM, line 374: @fabianfett not sure if you're fine with me just using the singleton ELG. It appears PostgresNIO already requires a high-enough SwiftNIO version.

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -360,4 +360,28 @@ final class IntegrationTests: XCTestCase {
         }
     }
     
+    func testConnectionClosureMidQueryDoesNotHang() async throws {
```
MahdiBM, line 363: Not sure if you're fine with me using async/await here. I could not quickly reproduce this with ELFs (the postgres connection would go out of scope and crash on deinit) but i think it should be possible.

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -360,4 +360,22 @@ final class IntegrationTests: XCTestCase {
...
+        _ = await withThrowingTaskGroup(of: Void.self) { taskGroup in
+            for _ in (0 ..< 1_000) {
```
fabianfett, line 365: why do we need to iterate a 1000 times here?

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -360,4 +360,22 @@ final class IntegrationTests: XCTestCase {
...
+                taskGroup.addTask {
+                    let conn = try await PostgresConnection.test(
+                        on: NIOSingletons.posixEventLoopGroup.next()
+                    ).get()
+
+                    async let close: () = conn.closeGracefully()
+                    async let query = conn.query("SELECT 1", logger: .psqlTest)
```
fabianfett, line 372: In which case do we run into a problem? Reading the code in `PostgresConnection` I assume we always run into the issue if we try to execute a query after the connection was closed?

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -222,7 +222,9 @@ public final class PostgresConnection: @unchecked Sendable {
             promise: promise
         )
 
-        self.channel.write(HandlerTask.extendedQuery(context), promise: nil)
+        let writePromise = self.channel.eventLoop.makePromise(of: Void.self)
+        self.channel.write(HandlerTask.extendedQuery(context), promise: writePromise)
+        writePromise.futureResult.cascadeFailure(to: promise)
```
fabianfett, line 227: Great catch! Can we please add for each of the following cases a `do {} catch {}` block in the following test:

```swift
    func testWeFailRequestsIfConnectionWasClosedBefore() async throws {
        let (connection, channel) = try await self.makeTestConnectionWithAsyncTestingChannel()

        try await connection.closeGracefully()

        XCTAssertEqual(channel.isActive, false)

        // 1. fail query
        do {
            _ = try await connection.query("SELECT version;", logger: self.logger)
            XCTFail("Expected to fail")
        } catch {
            print("\(error)")
        }

        // 2. fail listen
        // TODO!
    }
```

Add the test to `PostgresConnectionTests`

fabianfett (approved):
Great catch! We really need to make sure we cover all cases by adding unit tests for each of them!

fabianfett (changes requested):
Sorry used wrong button before :)

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -239,7 +241,10 @@ public final class PostgresConnection: @unchecked Sendable {
...
```
fabianfett, line 246: Does it make sense to move those three lines into a private function?

Tests/IntegrationTests/PSQLIntegrationTests.swift
MahdiBM: Not quite after. If the channel is already closed you get the "on IO Closed" error.

Tests/IntegrationTests/PSQLIntegrationTests.swift
MahdiBM: Hmm not quite sure. I'll check to see if it makes an actual difference compared to like only 100.
The problem is that the test is a bit flaky.
The problem often does get triggered by this test, but once in like 4/5 runs it doesn't get triggered.

fabianfett:
the test I added as an example hangs 100% of time for me.

Tests/IntegrationTests/PSQLIntegrationTests.swift
fabianfett: The point that I want to make here, is that your test case does not seem to be deterministic. The example one I wrote below is deterministic. Let's focus on real unit tests where we mock out the db!

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -530,6 +541,12 @@ extension PostgresConnection {
             throw error // rethrow with more metadata
         }
     }
+
+    private func write<T, Value>(_ any: T, cascadingFailureTo promise: EventLoopPromise<Value>) {
```
fabianfett, line 545: we don't need T here. Just make it `HandlerTask`

MahdiBM:
> the test I added as an example hangs 100% of time for me.
Hmm you're right. I thought i tested that properly and it turned out like i said, but apparently not.
I did try to remove the like you did with just closing the channel before hand, but i thought that didn't work out to get the hang.

Sources/PostgresNIO/Connection/PostgresConnection.swift
MahdiBM: done

Sources/PostgresNIO/Connection/PostgresConnection.swift
MahdiBM: done

Tests/IntegrationTests/PSQLIntegrationTests.swift
MahdiBM: I was wrong as mentioned below.
Changed the test.

Tests/IntegrationTests/PSQLIntegrationTests.swift
MahdiBM: I was wrong as mentioned below.

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -638,6 +638,63 @@ class PostgresConnectionTests: XCTestCase {
...
+            taskGroup.addTask {
+                try await Task.sleep(for: .seconds(3))
+
+                try await connection.close().get()
+                XCTAssertEqual(channel.isActive, false)
+            }
+
+            try await taskGroup.waitForAll()
+        }
+    }
```
MahdiBM, line 696: @fabianfett This is related to #458?
I noticed this same issue before they file an issue for when i was trying to put a `ServiceDiscovery` together using PostgresNIO for `DistributedActor`s, but didn't get the chance/mood to file and issue or it before they did.

Sources/PostgresNIO/New/Connection State Machine/ListenStateMachine.swift
```diff
@@ -224,12 +224,12 @@ extension ListenStateMachine {
         mutating func fail(_ error: Error) -> FailAction {
             switch self.state {
             case .initialized:
-                fatalError("Invalid state: \(self.state)")
+                return .none
```
MahdiBM, line 227: See https://github.com/vapor/postgres-nio/pull/487/files#r1647790890.
This is needed otherwise things will crash.

MahdiBM:
ahh hmm CI is failing with `Fatal error: SWIFT TASK CONTINUATION MISUSE: listen(_:) tried to resume its continuation more than once, throwing ioOnClosedChannel!` need to investigate.

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -638,6 +638,63 @@ class PostgresConnectionTests: XCTestCase {
...
+                    }
+                } catch let error as PSQLError {
+                    XCTAssertEqual(error.code, .listenFailed)
+                }
+            }
+
+            taskGroup.addTask {
+                try await Task.sleep(for: .seconds(3))
+
+                try await connection.close().get()
```
MahdiBM, line 690: Using `.close().get()` instead of `.closeGracefully()` here because `.closeGracefully()` will wait it out till infinitely and beyond since it's "graceful" and the listener is still there.

Sources/PostgresNIO/Connection/PostgresConnection.swift
MahdiBM: done.

MahdiBM:
I reckon I didn't add a unit test for "each" modified function. Should I?

fabianfett:
> I reckon I didn't add a unit test for "each" modified function. Should I?

Yes, please!

Sources/PostgresNIO/New/Connection State Machine/ListenStateMachine.swift
fabianfett: can we get this fix into another pr with a dedicated unit test case and explanation why this can happen?

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -638,6 +638,63 @@ class PostgresConnectionTests: XCTestCase {
...
+
+        do {
+            _ = try await connection.listen("test_channel")
+            XCTFail("Expected to fail")
+        } catch let error as ChannelError {
+            XCTAssertEqual(error, .ioOnClosedChannel)
+        }
+    }
+
+    func testListenFailsIfConnectionIsClosedMidway() async throws {
```
fabianfett, line 671: What connection state do you want to test here? In which connection state does this happen? Can we mock this explicitly and not by accident?

MahdiBM: The test is consistent, it only waits 3 seconds to make sure the listener has started listening.

It tests that if there is a listen in process, and then the connection is closed, the listener sequence needs to fail.

Sources/PostgresNIO/New/Connection State Machine/ListenStateMachine.swift
MahdiBM: hmmm sure. I'll remove the related test/changes from this PR.

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
fabianfett: ah! in that case we should wait for the client messages on the channel instead of using a timer.

See here how it is done:
https://github.com/vapor/postgres-nio/blob/7b621c16f6a0a8a0af8badd56b6f980457a1b7c5/Tests/PostgresNIOTests/New/PostgresConnectionTests.swift#L96

MahdiBM: Thanks for the clues, I did this in the new test.
We no longer need to change that `fatalError("Invalid state: \(self.state)")`, it appears that was triggered as a result of my bad test.

@fabianfett i think i no longer need to make another PR out of this?

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -224,6 +224,63 @@ class PostgresConnectionTests: XCTestCase {
...
+            try await channel.writeInbound(PostgresBackendMessage.parseComplete)
+            try await channel.writeInbound(PostgresBackendMessage.parameterDescription(.init(dataTypes: [])))
+            try await channel.writeInbound(PostgresBackendMessage.noData)
+            try await channel.writeInbound(PostgresBackendMessage.bindComplete)
+            try await channel.writeInbound(PostgresBackendMessage.commandComplete("LISTEN"))
+            try await channel.writeInbound(PostgresBackendMessage.readyForQuery(.idle))
+
+            try await channel.writeInbound(PostgresBackendMessage.notification(.init(backendPID: 12, channel: "foo", payload: "wooohooo")))
+
+            try await connection.close().get()
```
MahdiBM, line 271: again:

Using .close().get() instead of .closeGracefully() here because .closeGracefully() will wait it out till infinitely and beyond since it's "graceful" and the listener is still listening.

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
MahdiBM: @fabianfett I think this PR's changes are good, but I noticed this results in a fatal error regardless of my changes:
```swift
    func testAddListenerFailsIfConnectionIsClosed() async throws {
        let (connection, channel) = try await self.makeTestConnectionWithAsyncTestingChannel()

        connection.addListener(channel: "example") { context, notification in
            XCTFail("Did not expect to receive")
        }

        try await connection.close()

        XCTAssertEqual(channel.isActive, false)
    }
```
<kbd>
Screenshot 2024-06-23 at 1 16 11 AM
</kbd>

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -224,6 +224,63 @@ class PostgresConnectionTests: XCTestCase {
...
+            try await channel.writeInbound(PostgresBackendMessage.parseComplete)
+            try await channel.writeInbound(PostgresBackendMessage.parameterDescription(.init(dataTypes: [])))
+            try await channel.writeInbound(PostgresBackendMessage.noData)
+            try await channel.writeInbound(PostgresBackendMessage.bindComplete)
+            try await channel.writeInbound(PostgresBackendMessage.commandComplete("LISTEN"))
+            try await channel.writeInbound(PostgresBackendMessage.readyForQuery(.idle))
+
+            try await channel.writeInbound(PostgresBackendMessage.notification(.init(backendPID: 12, channel: "foo", payload: "wooohooo")))
+
+            try await connection.close()
```
MahdiBM, line 271: Using `.close()` instead of `.closeGracefully()` here because .closeGracefully() will wait it out till infinitely and beyond, apparently because it's "graceful" and the listener is still listening.

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
MahdiBM: revamped this test so this discussion is no longer valid. Though https://github.com/vapor/postgres-nio/pull/487#discussion_r1649799205 is related to the crash.

fabianfett (approved):
Great patch! Thanks for pushing through!

PR #497, opened 2024-08-03
Fix leakage of promises

### LINKED ISSUE #496: Memory leak in version 1.22.0
### Describe the issue

Memory usage spiked when we deployed our application with postgres-nio 1.22.0.

### Vapor version

4.102.1

Linux

### Swift version

Swift 5.10

### Steps to reproduce

Run a Swift backend application that uses postgres-nio and compare the memory usage when on 1.21.5 and 1.22.0.

### Outcome

Memory usage steadily increases after starting an application that uses postgres-nio 1.22.0.

### Additional notes

We observed this on an instance that runs recurring jobs but does not receive web traffic, so memory usage tends to be stable over time.

In the attached screenshot, the first release to include postgres-nio 1.22.0 was on July 26. The sudden drops in memory usage indicate when new releases went live, restarting the app. The graph stabilizes on July 30 when we reverted postgres-nio to 1.21.5.

postgres-nio-memory

Edit: I'm not sure if the memory usage spikes in 1.21.6 or 1.22.0.

MahdiBM:
resolves #496


MahdiBM:
Not sure how exactly add a test for this.
something like:
```swift
        let options = XCTMeasureOptions.default
        options.iterationCount = 100
        measure(metrics: [XCTMemoryMetric()], options: options) {
            do {
                let conn = try PostgresConnection.test(on: self.eventLoop).wait()
                _ = try conn.query(
                    "SELECT current_setting('application_name')",
                    logger: .psqlNoOpLogger
                ).wait()
                try conn.close().wait()
            } catch {
                XCTFail(String(reflecting: error))
            }
        }
```
could work but it's too integrated into Xcode.

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -544,12 +544,12 @@ struct ConnectionStateMachine {
     }
     
     mutating func enqueue(task: PSQLTask) -> ConnectionAction {
-        let psqlErrror: PSQLError
+        let psqlError: PSQLError
```
MahdiBM, line 547: a bunch of types, not sure if i should move them to another PR or it's fine.

Tests/PostgresNIOTests/New/Data/Date+PSQLCodableTests.swift
```diff
@@ -14,7 +14,7 @@ class Date_PSQLCodableTests: XCTestCase {
 
         var result: Date?
         XCTAssertNoThrow(result = try Date(from: &buffer, type: .timestamptz, format: .binary, context: .default))
-        XCTAssertEqual(value, result)
+        XCTAssertEqual(value.timeIntervalSince1970, result?.timeIntervalSince1970 ?? 0, accuracy: 0.1)
```
MahdiBM, line 17: was failing in nightly unit tests.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -212,23 +212,27 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
                 psqlTask = self.makeStartListeningQuery(channel: channel, context: context)
 
             case .none:
+                promise?.succeed(Void())
                 return
 
             case .succeedListenStart(let listener):
+                precondition(promise == nil, "This Promise will leak.")
```
MahdiBM, line 219: WDYT about these preconditions?
I tried to maximize the amount of preconditions, so for all the no-precondition paths there is a test that would otherwise trigger the precondition if it was actually there.

MahdiBM: It also appears these "listen"-related paths are not very well tested. Need to be careful with the preconditions.

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
fabianfett: Please create another PR for this.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
fabianfett: In this case we are already listening to channel in question. Just succeed the promise here as above.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -212,23 +212,27 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
                 psqlTask = self.makeStartListeningQuery(channel: channel, context: context)
```
fabianfett, line 212: we leak the promise here.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -212,23 +212,27 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
             case .succeedListenStart(let listener):
+                precondition(promise == nil, "This Promise will leak.")
                 listener.startListeningSucceeded(handler: self)
                 return
             }
 
         case .cancelListening(let channel, let id):
             switch self.listenState.cancelNotificationListener(channel: channel, id: id) {
             case .none:
+                precondition(promise == nil, "This Promise will leak.")
```
fabianfett, line 227: succeed the promise here.

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
MahdiBM: i'll remove the typo fixes then.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
MahdiBM: Not really.
2 things can happen in this function:
1- There are some `return`s. Before all those returns i've added some code to do something with the promise.
2- Otherwise we reach the end of the function where we pass the promise to the `run` function to take care of it.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -279,11 +285,11 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
         switch event {
         case PSQLOutgoingEvent.authenticate(let authContext):
             let action = self.state.provideAuthenticationContext(authContext)
-            self.run(action, with: context)
+            self.run(action, with: context, promise: promise)
 
         case PSQLOutgoingEvent.gracefulShutdown:
             let action = self.state.gracefulClose(promise)
-            self.run(action, with: context)
+            self.run(action, with: context, promise: nil)
```
MahdiBM, line 292: The action carries the promise, so the `run` is called without the promise.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -317,53 +323,74 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
             self.encoder.saslResponse(bytes)
-            context.writeAndFlush(self.wrapOutboundOut(self.encoder.flushBuffer()), promise: nil)
+            context.writeAndFlush(self.wrapOutboundOut(self.encoder.flushBuffer()), promise: runPromise)
         case .closeConnectionAndCleanup(let cleanupContext):
             self.closeConnectionAndCleanup(cleanupContext, context: context)
+            if let promise = cleanupContext.closePromise {
+                promise.futureResult.cascade(to: runPromise)
+            } else {
+                runPromise?.succeed(())
+            }
```
MahdiBM, line 361: If the cleanup context has a promise in it, cascade the result of that to `runPromise`.
Other wise succeed `runPromise`.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -317,53 +323,74 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
         case .fireChannelInactive:
             context.fireChannelInactive()
+            runPromise?.succeed(())
```
MahdiBM, line 364: There wasn't much in `fireChannelInactive()` so I can properly forward the promise to.
This looks fine to me though, to be clear.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -317,53 +323,73 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+            )
         case .sendBindExecuteSync(let executeStatement):
-            self.sendBindExecuteAndSyncMessage(executeStatement: executeStatement, context: context)
+            self.sendBindExecuteAndSyncMessage(executeStatement: executeStatement, context: context, promise: runPromise)
         case .sendParseDescribeBindExecuteSync(let query):
-            self.sendParseDescribeBindExecuteAndSyncMessage(query: query, context: context)
+            self.sendParseDescribeBindExecuteAndSyncMessage(query: query, context: context, promise: runPromise)
         case .succeedQuery(let promise, with: let result):
-            self.succeedQuery(promise, result: result, context: context)
+            self.succeedQuery(rowsPromise: promise, result: result, context: context, promise: runPromise)
```
MahdiBM, line 378: `succeedQuery()` currently doesn't do much complicated and I could just succeed `runPromise` here, but I decided to pass `runPromise` to it for implying that the function is responsible of it.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -317,53 +323,73 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
             }
         
         case .forwardRows(let rows):
             self.rowStream!.receive(rows)
-            
+            runPromise?.succeed(())
+
         case .forwardStreamComplete(let buffer, let commandTag):
             guard let rowStream = self.rowStream else {
+                runPromise?.succeed(())
```
MahdiBM, line 392: Should be good?

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -407,51 +437,62 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
                 context.writeAndFlush(self.wrapOutboundOut(self.encoder.flushBuffer()), promise: nil)
             }
             context.close(mode: .all, promise: promise)
+            if let promise {
+                promise.futureResult.cascade(to: runPromise)
+            } else {
+                runPromise?.succeed(())
+            }
```
MahdiBM, line 444: If the `closeConnection` has a promise in it, cascade the result of that to `runPromise`.
Other wise succeed `runPromise`.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -407,51 +437,62 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+            runPromise?.succeed(())
         case .failClose(let closeContext, with: let error, let cleanupContext):
             closeContext.promise.fail(error)
+            runPromise?.fail(error)
             if let cleanupContext = cleanupContext {
                 self.closeConnectionAndCleanup(cleanupContext, context: context)
             }
         case .forwardNotificationToListeners(let notification):
             self.forwardNotificationToListeners(notification, context: context)
+            runPromise?.succeed(())
```
MahdiBM, line 467: This also should be good to just succeed the promise here?
Should I handle it somehow else?
Should I pass the promise and succeed it in the function?

MahdiBM:
@fabianfett another round of review please 🙂

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -317,53 +323,73 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
 
     // MARK: Channel handler actions
     
-    private func run(_ action: ConnectionStateMachine.ConnectionAction, with context: ChannelHandlerContext) {
+    private func run(
+        _ action: ConnectionStateMachine.ConnectionAction,
+        with context: ChannelHandlerContext,
+        promise runPromise: EventLoopPromise<Void>?
```
fabianfett, line 329: this should not be part an extra argument here. Instead we should change: `ConnectionStateMachine.ConnectionAction`'s cases:

```
        case sendParseDescribeBindExecuteSync(PostgresQuery)
        case sendBindExecuteSync(PSQLExecuteStatement)
```

to also contain an optional writePromise.

fabianfett (changes requested):

Sources/PostgresNIO/New/Connection State Machine/AuthenticationStateMachine.swift
```diff
@@ -54,10 +54,10 @@ struct AuthenticationStateMachine {
                     return self.setAndFireError(PSQLError(code: .authMechanismRequiresPassword))
                 }
                 self.state = .passwordAuthenticationSent
-                return .sendPassword(.md5(salt: salt), self.authContext)
+                return .sendPassword(.md5(salt: salt), self.authContext, promise: nil)
             case .plaintext:
                 self.state = .passwordAuthenticationSent
-                return .sendPassword(.cleartext, authContext)
+                return .sendPassword(.cleartext, authContext, promise: nil)
```
MahdiBM, line 60: called from channel read, I double checked and it seems to be fine to pass `nil` as the actual write does have a promise which it fulfills.

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -583,14 +586,16 @@ struct ConnectionStateMachine {
...
+            writePromise?.fail(psqlErrror) /// Use `cleanupContext` or not?
             switch queryContext.query {
             case .executeStatement(_, let promise), .unnamed(_, let promise):
                 return .failQuery(promise, with: psqlErrror, cleanupContext: nil)
             case .prepareStatement(_, _, _, let promise):
                 return .failPreparedStatementCreation(promise, with: psqlErrror, cleanupContext: nil)
             }
-        case .closeCommand(let closeContext):
+        case .closeCommand(let closeContext, let writePromise):
+            writePromise?.fail(psqlErrror) /// Use `cleanupContext` or not?
```
MahdiBM, line 598: comment

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -226,15 +229,15 @@ struct ConnectionStateMachine {
         }
     }
     
-    mutating func sslSupportedReceived(unprocessedBytes: Int) -> ConnectionAction {
+    mutating func sslSupportedReceived(unprocessedBytes: Int, promise: EventLoopPromise<Void>?) -> ConnectionAction {
```
MahdiBM, line 232: This function is only used from channel-reads so currently it'll never have a non-nil promise.
I'm still properly accepting and passing the possible-promise around.

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -583,14 +586,16 @@ struct ConnectionStateMachine {
         }
 
         switch task {
-        case .extendedQuery(let queryContext):
+        case .extendedQuery(let queryContext, let writePromise):
```
MahdiBM, line 589: comment

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -650,32 +691,34 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
-                let task = self.makeStartListeningQuery(channel: channel, context: context)
+                let task = self.makeStartListeningQuery(channel: channel, context: context, writePromise: writePromise)
                 let action = self.state.enqueue(task: task)
                 self.run(action, with: context)
             }
 
         case .failure(let error):
             let action = self.state.errorHappened(.unlistenError(underlying: error))
             self.run(action, with: context)
+            writePromise?.fail(error) /// Should I pass the promise to the action? seemed troublesome
```
MahdiBM, line 721: comment

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -757,16 +808,18 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+                        name: preparedStatement.name,
+                        binds: preparedStatement.bindings,
+                        rowDescription: action.rowDescription
+                    ),
+                    logger: preparedStatement.logger,
+                    promise: preparedStatement.promise
                 ),
-                logger: preparedStatement.logger,
-                promise: preparedStatement.promise
+                writePromise: nil // Ignore
```
MahdiBM, line 821: ignore?

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -64,8 +64,8 @@ struct ConnectionStateMachine {
         
         case read
         case wait
-        case sendSSLRequest
-        case establishSSLConnection
```
fabianfett, line 68: when would this ever be non nil?

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -77,16 +77,16 @@ struct ConnectionStateMachine {
...
         
         // Auth Actions
-        case sendStartupMessage(AuthContext)
-        case sendPasswordMessage(PasswordAuthencationMode, AuthContext)
-        case sendSaslInitialResponse(name: String, initialResponse: [UInt8])
-        case sendSaslResponse([UInt8])
+        case sendStartupMessage(AuthContext, promise: EventLoopPromise<Void>?)
+        case sendPasswordMessage(PasswordAuthencationMode, AuthContext, promise: EventLoopPromise<Void>?)
+        case sendSaslInitialResponse(name: String, initialResponse: [UInt8], promise: EventLoopPromise<Void>?)
+        case sendSaslResponse([UInt8], promise: EventLoopPromise<Void>?)
```
fabianfett, line 83: when would those ever be non nil?

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -164,8 +164,11 @@ struct ConnectionStateMachine {
         }
     }
     
-    mutating func provideAuthenticationContext(_ authContext: AuthContext) -> ConnectionAction {
-        self.startAuthentication(authContext)
+    mutating func provideAuthenticationContext(
```
fabianfett, line 167: when would the promise ever be non nil?

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -583,14 +583,16 @@ struct ConnectionStateMachine {
         }
 
         switch task {
-        case .extendedQuery(let queryContext):
+        case .extendedQuery(let queryContext, let writePromise):
+            writePromise?.fail(psqlErrror) /// Use `cleanupContext` or not?
```
MahdiBM, line 587: comment

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -76,13 +76,14 @@ struct ExtendedQueryStateMachine {
                 case .none:
                     state = .noDataMessageReceived(queryContext)
                 }
-                return .sendBindExecuteSync(prepared)
+                return .sendBindExecuteSync(prepared, promise: promise)
             }
 
+            /// Not my code, but this is ignoring the last argument which is a promise? is that fine?
```
MahdiBM, line 82: comment

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -602,23 +631,28 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
         switch result {
         case .success:
             switch self.listenState.startListeningSucceeded(channel: channel) {
             case .activateListeners(let listeners):
                 for list in listeners {
-                    list.startListeningSucceeded(handler: self)
+                    list.startListeningSucceeded(handler: self, writePromise: nil)
                 }
-
+                writePromise?.succeed(()) /// Should we instead do smth like "whenAllSucceed"?
```
MahdiBM, line 652: comment

PR #504, opened 2024-08-22
Expose query `metadata` in `PostgresRowSequence`

MahdiBM:
Currently the functions that return `PostgresRowSequence` have no way of reporting back the `metadata` of the query.
This PR exposes the `metadata` through the `collectWithMetadata()` method for now.

### DIFF
--- a/Sources/PostgresNIO/New/PSQLRowStream.swift
+++ b/Sources/PostgresNIO/New/PSQLRowStream.swift
@@ -44,7 +44,7 @@ final class PSQLRowStream: @unchecked Sendable {
     }
     
     internal let rowDescription: [RowDescription.Column]
-    private let lookupTable: [String: Int]
+    internal let lookupTable: [String: Int]
     private var downstreamState: DownstreamState
     
     init(
@@ -114,7 +114,7 @@ final class PSQLRowStream: @unchecked Sendable {
             self.downstreamState = .consumed(.failure(error))
         }
         
-        return PostgresRowSequence(producer.sequence, lookupTable: self.lookupTable, columns: self.rowDescription)
+        return PostgresRowSequence(producer.sequence, stream: self)
     }
     
     func demand() {
--- a/Sources/PostgresNIO/New/PostgresRowSequence.swift
+++ b/Sources/PostgresNIO/New/PostgresRowSequence.swift
@@ -9,14 +9,18 @@ public struct PostgresRowSequence: AsyncSequence, Sendable {
 
     typealias BackingSequence = NIOThrowingAsyncSequenceProducer<DataRow, Error, AdaptiveRowBuffer, PSQLRowStream>
 
-    let backing: BackingSequence
-    let lookupTable: [String: Int]
-    let columns: [RowDescription.Column]
+    private let backing: BackingSequence
+    private let stream: PSQLRowStream
+    var lookupTable: [String: Int] {
+        self.stream.lookupTable
+    }
+    var columns: [RowDescription.Column] {
+        self.stream.rowDescription
+    }
 
-    init(_ backing: BackingSequence, lookupTable: [String: Int], columns: [RowDescription.Column]) {
+    init(_ backing: BackingSequence, stream: PSQLRowStream) {
         self.backing = backing
-        self.lookupTable = lookupTable
-        self.columns = columns
+        self.stream = stream
     }
 
     public func makeAsyncIterator() -> AsyncIterator {
@@ -66,6 +70,12 @@ extension PostgresRowSequence {
             result.append(row)
         }
         return result
+    }
+
+    public func collectWithMetadata() async throws -> (metadata: PostgresQueryMetadata?, rows: [PostgresRow]) {
+        let rows = try await self.collect()
+        let metadata = PostgresQueryMetadata(string: self.stream.commandTag)
+        return (metadata, rows)
     }
 }
 
--- a/Tests/IntegrationTests/AsyncTests.swift
+++ b/Tests/IntegrationTests/AsyncTests.swift
@@ -46,6 +46,31 @@ final class AsyncPostgresConnectionTests: XCTestCase {
         }
     }
 
+    func testSelect10kRowsAndCollect() async throws {
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer { XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully()) }
+        let eventLoop = eventLoopGroup.next()
+
+        let start = 1
+        let end = 10000
+
+        try await withTestConnection(on: eventLoop) { connection in
+            let rows = try await connection.query("SELECT generate_series(\(start), \(end));", logger: .psqlTest)
+            let (metadata, elements) = try await rows.collectWithMetadata()
+            var counter = 0
+            for row in elements {
+                let element = try row.decode(Int.self)
+                XCTAssertEqual(element, counter + 1)
+                counter += 1
+            }
+            XCTAssertEqual(metadata?.command, "SELECT")
+            XCTAssertEqual(metadata?.oid, nil)
+            XCTAssertEqual(metadata?.rows, 10000)
+
+            XCTAssertEqual(counter, end)
+        }
+    }
+
     func testSelectActiveConnection() async throws {
         let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
         defer { XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully()) }
@@ -207,7 +232,7 @@ final class AsyncPostgresConnectionTests: XCTestCase {
 
         try await withTestConnection(on: eventLoop) { connection in
             // Max binds limit is UInt16.max which is 65535 which is 3 * 5 * 17 * 257
-            // Max columns limit is 1664, so we will only make 5 * 257 columns which is less
+            // Max columns limit appears to be ~1600, so we will only make 5 * 257 columns which is less
             // Then we will insert 3 * 17 rows
             // In the insertion, there will be a total of 3 * 17 * 5 * 257 == UInt16.max bindings
             // If the test is successful, it means Postgres supports UInt16.max bindings
@@ -241,13 +266,9 @@ final class AsyncPostgresConnectionTests: XCTestCase {
                 unsafeSQL: "INSERT INTO table1 VALUES \(insertionValues)",
                 binds: binds
             )
-            try await connection.query(insertionQuery, logger: .psqlTest)
-
-            let countQuery = PostgresQuery(unsafeSQL: "SELECT COUNT(*) FROM table1")
-            let countRows = try await connection.query(countQuery, logger: .psqlTest)
-            var countIterator = countRows.makeAsyncIterator()
-            let insertedRowsCount = try await countIterator.next()?.decode(Int.self, context: .default)
-            XCTAssertEqual(rowsCount, insertedRowsCount)
+            let result = try await connection.query(insertionQuery, logger: .psqlTest)
+            let metadata = try await result.collectWithMetadata().metadata
+            XCTAssertEqual(metadata?.rows, rowsCount)
 
             let dropQuery = PostgresQuery(unsafeSQL: "DROP TABLE table1")
             try await connection.query(dropQuery, logger: .psqlTest)


Tests/IntegrationTests/AsyncTests.swift
```diff
@@ -241,13 +266,9 @@ final class AsyncPostgresConnectionTests: XCTestCase {
...
-            try await connection.query(insertionQuery, logger: .psqlTest)
-
-            let countQuery = PostgresQuery(unsafeSQL: "SELECT COUNT(*) FROM table1")
-            let countRows = try await connection.query(countQuery, logger: .psqlTest)
-            var countIterator = countRows.makeAsyncIterator()
-            let insertedRowsCount = try await countIterator.next()?.decode(Int.self, context: .default)
-            XCTAssertEqual(rowsCount, insertedRowsCount)
+            let result = try await connection.query(insertionQuery, logger: .psqlTest)
+            let metadata = try await result.collectWithMetadata().metadata
+            XCTAssertEqual(metadata?.rows, rowsCount)
```
MahdiBM, line 271: The metadata contains the count of the inserted rows so no need to do another query.

Sources/PostgresNIO/New/PostgresRowSequence.swift
```diff
@@ -67,6 +71,12 @@ extension PostgresRowSequence {
         }
         return result
     }
+
+    public func collectWithMetadata() async throws -> (metadata: PostgresQueryMetadata?, rows: [PostgresRow]) {
```
MahdiBM, line 75: I'm open to better names.

Sources/PostgresNIO/New/PostgresRowSequence.swift
```diff
@@ -60,13 +64,48 @@ extension PostgresRowSequence {
...
+    /// for try await row in myPostgresSequence {
+    ///     /// Process each row
+    /// }
+    /// ```
+    ///
+    /// - Parameter onRow: Processes each row.
+    /// - Returns: The query metadata.
+    public func consume(
+        onRow: @Sendable (PostgresRow) throws -> ()
+    ) async throws -> PostgresQueryMetadata {
```
MahdiBM, line 100: I'm open to better names.

Tests/IntegrationTests/AsyncTests.swift
```diff
@@ -241,13 +294,9 @@ final class AsyncPostgresConnectionTests: XCTestCase {
...
-            try await connection.query(insertionQuery, logger: .psqlTest)
-
-            let countQuery = PostgresQuery(unsafeSQL: "SELECT COUNT(*) FROM table1")
-            let countRows = try await connection.query(countQuery, logger: .psqlTest)
-            var countIterator = countRows.makeAsyncIterator()
-            let insertedRowsCount = try await countIterator.next()?.decode(Int.self, context: .default)
-            XCTAssertEqual(rowsCount, insertedRowsCount)
+            let result = try await connection.query(insertionQuery, logger: .psqlTest)
+            let metadata = try await result.collectWithMetadata().metadata
+            XCTAssertEqual(metadata.rows, rowsCount)
```
MahdiBM, line 299: The metadata contains the count of the inserted rows so no need to do another query.

Sources/PostgresNIO/New/PostgresRowSequence.swift
MahdiBM: Also the closure is intentionally not `async` since it's generally not a good idea to be doing `async` things on each row.
It's still possible to accumulate the rows and do whatever you want after.

MahdiBM:
@fabianfett I've thought a bit more about this and possibly we might want to have another type that wraps `PostgresRowSequence` in the next major version at least? so we can expose more stuff in the response. Not only the metadata, but maybe even the connection properties? Or who knows what.
I don't have an immediate use-case for connection properties but it looks like some lost info. Maybe someone finds a value in them although they can also be individually queried.

Sources/PostgresNIO/New/PSQLRowStream.swift
```diff
@@ -114,7 +114,7 @@ final class PSQLRowStream: @unchecked Sendable {
             self.downstreamState = .consumed(.failure(error))
         }
         
-        return PostgresRowSequence(producer.sequence, lookupTable: self.lookupTable, columns: self.rowDescription)
+        return PostgresRowSequence(producer.sequence, rowStream: self)
```
fabianfett, line 117: This looks very much like a retain cycle!

bridger:
I'd love this addition! I have queries that should update exactly 1 row. I'd like to check that the update was successful, but I can't check without the metadata. (These queries currently don't return the row itself.)

fabianfett (changes requested):
I'm not a fan of the proposed API changes: They either get everything into memory, or they force users to consume the rows synchronously. I consider both bad options.

However I do have an alternative in mind:

```swift
    // use this for queries where you don't expect any rows back.
    @discardableResult
    func execute(_ query: PostgresQuery) async throws -> PostgresQueryMetadata

    // use this for queries where you want to consume the rows.
    // we can use the `consume` scope to better ensure structured concurrency when consuming the rows.
    func query<Result>(
        _ query: PostgresQuery, 
        _ consume: (PostgresRowSequence) async throws -> Result
    ) async throws -> (Result, PostgresQueryMetadata)
}
```

@MahdiBM Wdyt?

MahdiBM:
> I'm not a fan of the proposed API changes: They either get everything into memory, or they force users to consume the rows synchronously. I consider both bad options.

I honestly don't think this is a problem. We're not _really_ taking any options away.

> However I do have an alternative in mind:

To be honest this looks better, at least at the first glance. So I'd be happy to move this way.

duncangroenewald:
@MahdiBM what is the expected timeframe for this change to be included in a release ?  Just wondering whether I should wait or fall back to using Connection.query in the interim ?  I don't mind using a pre-release branch if there is one since we are in development still.

MahdiBM:
@duncangroenewald IIRC you found a solution for now using that PostgresDatabase function that has a `onMetadata`?
Anyway, I'm not sure about when this'll be released. I finished the PR now, @fabianfett will need to review and approve when he has time, then we'll see.

MahdiBM:
CodeCov report is not accurate about those funcs in the first 2 listed files not having tests.
For some reason even Xcode can't correctly find the new funcs' symbols.

duncangroenewald:
@MahdiBM  - OK well if there is any chance of creating a beta branch with this added let me know as I would rather use that than have to go back and simplify things later on.

MahdiBM:
@duncangroenewald I took a look and this branch is up to date with main, so you _could_ temporarily depend on this branch (repo: https://github.com/MahdiBM/postgres-nio, branch: mmbm-row-seq-expose-metadata) if you want.
Or did you mean a "beta release"?

duncangroenewald:
thanks that branch of yours should be fine - just don't deleted it :)

I'll let you know how it goes.

Sources/PostgresNIO/New/PSQLRowStream.swift
```diff
@@ -276,7 +276,70 @@ final class PSQLRowStream: @unchecked Sendable {
             return self.eventLoop.makeFailedFuture(error)
         }
     }
-    
+
+    // MARK: Drain on EventLoop
+
+    func drain() -> EventLoopFuture<Void> {
+        if self.eventLoop.inEventLoop {
```
fabianfett, line 283: This is already implemented with `cancel`.

Sources/PostgresNIO/New/PSQLRowStream.swift
```diff
@@ -276,7 +276,70 @@ final class PSQLRowStream: @unchecked Sendable {
...
+            return self.drain0()
+        } else {
+            return self.eventLoop.flatSubmit {
+                self.drain0()
+            }
+        }
+    }
+
+    private func drain0() -> EventLoopFuture<Void> {
```
fabianfett, line 292: This is already implemented with `cancel0`.

.gitignore
```diff
@@ -1,5 +1,6 @@
 .DS_Store
 /.build
+/.index-build
```
fabianfett, line 3: why have you added this one?

fabianfett (commented):
Quick review

Sources/PostgresNIO/New/PSQLRowStream.swift
MahdiBM: cancel is not enough:

1- it doesn't wait until the query is over so metadata is available.
If you try to get the metadata too soon, the PostgresNIO code will crash.

2- it doesn't cover all cases that might happen. For example when downstream is waiting for consumer.
That might happen when the user hasn't even tried to consume the rows.

Sources/PostgresNIO/New/PSQLRowStream.swift
MahdiBM: See the other comment.

.gitignore
MahdiBM: This is like a .build directory but for sourcekit-lsp.
In the sourcekit-lsp bundled with swift 6.1 it'll be moved into .build dir itself, but that's not out.

MahdiBM: you need to have Swift 6.0.x and have experimental-indexing enabled (e.g. in the VSCode extension or using the sourcekit-lsp config file).

Tests/IntegrationTests/AsyncTests.swift
```diff
@@ -241,13 +333,8 @@ final class AsyncPostgresConnectionTests: XCTestCase {
...
             )
-            try await connection.query(insertionQuery, logger: .psqlTest)
-
-            let countQuery = PostgresQuery(unsafeSQL: "SELECT COUNT(*) FROM table1")
-            let countRows = try await connection.query(countQuery, logger: .psqlTest)
-            var countIterator = countRows.makeAsyncIterator()
-            let insertedRowsCount = try await countIterator.next()?.decode(Int.self, context: .default)
-            XCTAssertEqual(rowsCount, insertedRowsCount)
+            let metadata = try await connection.execute(insertionQuery, logger: .psqlTest)
+            XCTAssertEqual(metadata.rows, rowsCount)
```
MahdiBM, line 337: This is a test i originally wrote myself in another PR. Moved it to use query metadata since that's what should have been done anyway.

Tests/PostgresNIOTests/New/PSQLRowStreamTests.swift
```diff
@@ -19,20 +19,29 @@ final class PSQLRowStreamTests: XCTestCase {
         
         XCTAssertEqual(try stream.all().wait(), [])
         XCTAssertEqual(stream.commandTag, "INSERT 0 1")
+
+        XCTAssertNoThrow(try stream.drain().wait())
```
MahdiBM, line 23: all these tests are covering real-world use-cases.

we hand the postgres-sequence over to users. If they have fully consumed the rows, a `drain()` call should simply just do nothing. Not crash or ....

Tests/PostgresNIOTests/New/PSQLRowStreamTests.swift
```diff
@@ -19,20 +19,29 @@ final class PSQLRowStreamTests: XCTestCase {
...
+        let expectedError = PSQLError.serverClosedConnection(underlying: nil)
+
         XCTAssertThrowsError(try stream.all().wait()) {
-            XCTAssertEqual($0 as? PSQLError, .serverClosedConnection(underlying: nil))
+            XCTAssertEqual($0 as? PSQLError, expectedError)
+        }
+
+        // Drain should work
+        XCTAssertThrowsError(try stream.drain().wait()) {
+            XCTAssertEqual($0 as? PSQLError, expectedError)
```
MahdiBM, line 41: in case of error, drain will just throw the same error.
Though realistically if there is an error users will not reach this far in the code.

Tests/PostgresNIOTests/New/PSQLRowStreamTests.swift
```diff
@@ -231,6 +240,196 @@ final class PSQLRowStreamTests: XCTestCase {
...
+            [ByteBuffer(string: "1")]
+        ])
+
+        XCTAssertEqual(dataSource.hitDemand, 0, "Before we have a consumer demand is not signaled")
+
+        // attach consumers
+        let rowSequence = stream.asyncSequence()
+        XCTAssertEqual(dataSource.hitDemand, 0)
+        let drainFuture = stream.drain()
+        XCTAssertEqual(dataSource.hitDemand, 1)
```
MahdiBM, line 406: This might happen when someone is using the new `query` function. an async-seq is attached but user might simply ignore it.

Tests/PostgresNIOTests/New/PSQLRowStreamTests.swift
```diff
@@ -231,6 +240,196 @@ final class PSQLRowStreamTests: XCTestCase {
...
+            [ByteBuffer(string: "1")]
+        ])
+
+        XCTAssertEqual(dataSource.hitDemand, 0, "Before we have a consumer demand is not signaled")
+
+        // attach consumers
+        let allFuture = stream.all()
+        XCTAssertEqual(dataSource.hitDemand, 1)
+        let drainFuture = stream.drain()
+        XCTAssertEqual(dataSource.hitDemand, 2)
```
MahdiBM, line 359: I reckon this case won't actually happen at all with the current query functions and technically should only happen if users are misusing something like the postgres-seq by escaping it and trying to consume it weirdly or in a Task. It hits the `case .waitingForAll` code in `drain0()`.

Should i remove it or just let it be incase?

.gitignore
MahdiBM: SwiftNIO already has it as well:
https://github.com/apple/swift-nio/blob/main/.gitignore#L3

Tests/PostgresNIOTests/New/PSQLRowStreamTests.swift
MahdiBM: I'd personally put a different `preconditionFailure` in `case .waitingForAll ` and try to let users know what they did wrong + delete this test 🤔.

MahdiBM: Or maybe we should keep that code for the sake of completeness in `drain0()` and having it not be like "only supposed to be used before you want to get the metadata"

MahdiBM:
@fabianfett can you have another look at this please? 🙂

PR #505, opened 2024-08-25
Implement `SimpleQuery`

MahdiBM:
### Checklist
* [x] Implement basic SimpleQuery handling.
* [x] Tests.
* [ ] Decoding rows from a SimpleQuery response.
* [ ] Multiple statements in a single `SingleQuery`.


Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -198,7 +213,7 @@ struct ExtendedQueryStateMachine {
         }
 
         switch queryContext.query {
-        case .unnamed, .executeStatement:
+        case .unnamed, .executeStatement, .simpleQuery:
             return .wait
```
MahdiBM, line 217: After a `RowDescription`, just wait.

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -258,20 +276,40 @@ struct ExtendedQueryStateMachine {
...
+                    return self.setAndFireError(.unexpectedBackendMessage(.dataRow(dataRow)))
+                }
+
+                return self.avoidingStateMachineCoW { state -> Action in
+                    var demandStateMachine = RowStreamStateMachine()
+                    demandStateMachine.receivedRow(dataRow)
+                    state = .streaming(columns, demandStateMachine)
+                    let result = QueryResult(value: .rowDescription(columns), logger: queryContext.logger)
+                    return .succeedQuery(eventLoopPromise, with: result)
+                }
```
MahdiBM, line 294: If a `DataRow` is received and `state` is `RowDescription`, it means we have a query that does return rows.
In that case we need to succeed the query here.

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -292,10 +330,36 @@ struct ExtendedQueryStateMachine {
...
+            case .simpleQuery(_, let eventLoopGroup):
+                return self.avoidingStateMachineCoW { state -> Action in
+                    state = .commandComplete(commandTag: commandTag)
+                    let result = QueryResult(value: .noRows(.tag(commandTag)), logger: context.logger)
+                    return .succeedQuery(eventLoopGroup, with: result)
+                }
+
+            case .unnamed, .executeStatement, .prepareStatement:
+                return self.setAndFireError(.unexpectedBackendMessage(.commandComplete(commandTag)))
+            }
```
MahdiBM, line 348: If a `CommandComplete` is received and `state` is `messagesSent`, It means we had a query that wasn't querying any rows.

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -292,10 +330,36 @@ struct ExtendedQueryStateMachine {
...
-            
+
+        case .messagesSent(let context):
+            switch context.query {
+            case .simpleQuery(_, let eventLoopGroup):
+                return self.avoidingStateMachineCoW { state -> Action in
+                    state = .commandComplete(commandTag: commandTag)
+                    let result = QueryResult(value: .noRows(.tag(commandTag)), logger: context.logger)
+                    return .succeedQuery(eventLoopGroup, with: result)
+                }
```
MahdiBM, line 344: If a CommandComplete is received and state is messagesSent, It means we had a query that wasn't querying any rows.

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -292,10 +330,36 @@ struct ExtendedQueryStateMachine {
...
+            }
+
+        case .rowDescriptionReceived(let context, _):
+            switch context.query {
+            case .simpleQuery(_, let eventLoopPromise):
+                return self.avoidingStateMachineCoW { state -> Action in
+                    state = .commandComplete(commandTag: commandTag)
+                    let result = QueryResult(value: .noRows(.tag(commandTag)), logger: context.logger)
+                    return .succeedQuery(eventLoopPromise, with: result)
+                }
```
MahdiBM, line 357: If a `CommandComplete` is received and state is `RowDescription`, It means we had a query that was in fact querying some rows, but it just happened that there were no rows matching the query.

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -323,20 +385,32 @@ struct ExtendedQueryStateMachine {
...
-
-        case .prepareStatement(_, _, _, _):
+        case .messagesSent(let queryContext):
+            switch queryContext.query {
+            case .simpleQuery(_, let eventLoopPromise):
+                return self.avoidingStateMachineCoW { state -> Action in
+                    state = .emptyQueryResponseReceived
+                    let result = QueryResult(value: .noRows(.emptyResponse), logger: queryContext.logger)
+                    return .succeedQuery(eventLoopPromise, with: result)
+                }
```
MahdiBM, line 409: If a `EmptyQuery` is received and state is `messageSent`, It means we had a query that was empty, like only contained some comments.

Sources/PostgresNIO/New/PSQLTask.swift
```diff
@@ -31,16 +33,18 @@ enum PSQLTask {
...
 }
 
+// FIXME: Either rename all these `ExtendedQuery`s to just like `Query` or pull out `simpleQuery`
 final class ExtendedQueryContext {
     enum Query {
         case unnamed(PostgresQuery, EventLoopPromise<PSQLRowStream>)
         case executeStatement(PSQLExecuteStatement, EventLoopPromise<PSQLRowStream>)
         case prepareStatement(name: String, query: String, bindingDataTypes: [PostgresDataType], EventLoopPromise<RowDescription?>)
+        case simpleQuery(String, EventLoopPromise<PSQLRowStream>)
     }
```
MahdiBM, line 43: `simpleQuery`  being here doesn't quite make sense.

As i see, our options are:
* Rename all these `ExtendedQuery` types to just `Query`. (still doesn't make great sense)
* Create a new type and all, and possibly a new state machine for `SimpleQuery`.

@fabianfett WDYT?

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -438,6 +438,47 @@ extension PostgresConnection {
...
+    ///
+    /// - Parameters:
+    ///   - query: The simple query to run
+    ///   - logger: The `Logger` to log into for the query
+    ///   - file: The file, the query was started in. Used for better error reporting.
+    ///   - line: The line, the query was started in. Used for better error reporting.
+    /// - Returns: A ``PostgresRowSequence`` containing the rows the server sent as the query result.
+    ///            The sequence  be discarded.
+    @discardableResult
+    public func __simpleQuery(
```
MahdiBM, line 453: I'm not sure what will happen to this PR but the name is underscored for now so we can merge this PR as-is and I can then create follow-up PRs.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -438,6 +438,47 @@ extension PostgresConnection {
...
```
MahdiBM, line 453: I'm not sure what will happen to this PR but the name is underscored for now so we can merge this PR as-is and I can then create follow-up PRs.

fabianfett:
I wonder if a better way is to build a dedicated `SimpleQueryStateMachine`? Wdyt?

MahdiBM:
@fabianfett i moved the stuff to that, mid-PR.
There is a `SimpleQueryStateMachine` file, though in GitHub CI you might need to chase it and push that "Load diff" button.

MahdiBM:
FWIW we're already using this function on prod and it's been working as it should.
Of course we're not querying rows with multiple queries because that's not implemented.
The tests should also be pretty good 🙂.

PR #510, opened 2024-09-25
Fix cross-compilation to the static Linux SDK

leinelissen:
When cross-compiling a Vapor app with the Swift Static Linux SDK, postgres-nio will throw `error: no such module 'Glibc'`. This fix imports `Musl` instead of `Glibc` when it is available to fix cross-compilation.

Note: given other errors with the Static Linux SDK and Swift toolchain, you'll need to run at least the `swift-6.0-DEVELOPMENT-SNAPSHOT-2024-09-17-a` Swift development toolchain and corresponding Linux SDK. Also, there's an unreleased fix in swift-nio, so it's recommended to run at least the `1b33db2dea6a64d5b619b9e888175133c6d7f410` revision of the package when cross-compiling.

### DIFF
--- a/Sources/ConnectionPoolModule/PoolStateMachine.swift
+++ b/Sources/ConnectionPoolModule/PoolStateMachine.swift
@@ -1,7 +1,9 @@
 #if canImport(Darwin)
 import Darwin
-#else
+#elseif canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
 #endif
 
 @usableFromInline


0xTim (approved):
Thanks!

@gwynne we really need to work static Linux SDK into the CI at some point

gwynne (approved):

fabianfett (approved):

ffried:
Any chance this could make it into a release soon(-ish)? We're trying to enable multi-platform docker builds for our server side applications and this is one blocker (that can currently be worked around by manually adding `postgres-nio` as dependency with a fixed commit hash to the current `HEAD` of `main`).

PR #515, opened 2024-10-09
Drop Swift 5.8 support and update CI

gwynne:


.github/workflows/test.yml
```diff
@@ -171,8 +170,12 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
       - name: Run all tests
-        run: swift test
-
+        run: swift test --enable-code-coverage
+      - name: Submit code coverage
+        uses: vapor/swift-codecov-action@v0.3
+        with:
+          codecov_token: ${{ secrets.CODECOV_TOKEN }}
```
fabianfett, line 177: why change this?

gwynne: Because Codecov intelligently merges multiple reports for the same commit and will provide coverage info that takes into account any OS-specific logic.

fabianfett: I think there were reasons why we didn't run this on macOS.

gwynne: It's my impression that those reasons no longer apply, though I'm fine with taking it back out if you feel strongly about it.

fabianfett: macOS CI also runs integration tests, which is why codecov is through the roof. codecov shall only apply to unit tests. Please remove.

gwynne: Done

fabianfett (approved):

Issue #512, opened 2024-10-04
Transaction support


adam-fowler:
Does it make sense to add support for transactions in similar manner to this
```swift
extension PostgresClient {
    func withTransaction<Value>(logger: Logger, _ process: (PostgresConnection) async throws -> Value) async throws -> Value {
        try await withConnection { connection in
            try await connection.query("BEGIN;", logger: logger)
            do {
                let value = try await process(connection)
                try await connection.query("COMMIT;", logger: logger)
                return value
            } catch {
                try await connection.query("ROLLBACK;", logger: logger)
                throw error
            }
        }
    }
}
```

robertjpayne:
This was implemented in #519 and #538 but one very important omission I see is that the API doesn't allow specifying an isolation level. While the Postgres default isolation level is really safe, serializable isolation level can be preferred for some applications.

Serializable isolation can require some extra ability to allow retries since serializable errors are not fatal and retries can normally proceed just fine.

fabianfett:
@robertjpayne  I'd be happy to accept a PR that adds this :)

wibed:
is transaction support given in the postgres client sense?

because id assume at the moment  the pool rotating a backend per-transaction and reclaiming it the instant each transaction ends, so clients multiplex at transaction granularity, is currently not implemented.

PR #519, opened 2024-10-11
feat(transaction): Adding withTransaction

thoven87:
First pass at feature #512 

What's the best way to go about writing unit test for the withTransaction func?


Tests/IntegrationTests/PostgresClientTests.swift
```diff
@@ -42,6 +42,41 @@ final class PostgresClientTests: XCTestCase {
...
+        
+        await withThrowingTaskGroup(of: Void.self) { taskGroup in
+            taskGroup.addTask {
+                await client.run()
+            }
+            
+            let iterations = 1000
+            
+            for _ in 0..<iterations {
+                taskGroup.addTask {
```
fabianfett, line 66: I think we should create a throwaway table here and insert some data. then we should check if a commit commits and if a rollback rolls back.

Sources/PostgresNIO/Pool/PostgresClient.swift
```diff
@@ -303,6 +303,28 @@ public final class PostgresClient: Sendable, ServiceLifecycle.Service {
...
+    /// A transaction should end with a call to COMMIT or ROLLBACK
+    /// COMMIT is called upon successful completion and ROLLBACK is called should any steps fail
+    ///
+    /// - Parameter closure: A closure that uses the passed `PostgresConnection`. The closure **must not** capture
+    ///                      the provided `PostgresConnection`.
+    /// - Returns: The closure's return value.
+    public func withTransaction<Result>(logger: Logger, _ process: (PostgresConnection) async throws -> Result) async throws -> Result {
+        try await withConnection { connection in
+            do {
+                try await connection.query("BEGIN;", logger: logger)
```
adam-fowler, line 318: Should the BEGIN be outside the do catch. No point rolling back a begin that never happened.

Tests/IntegrationTests/PostgresClientTests.swift
thoven87: > I think we should create a throwaway table here and insert some data. then we should check if a commit commits and if a rollback rolls back.

I added the tests in the latest commit. Thanks!

Tests/IntegrationTests/PostgresClientTests.swift
```diff
@@ -42,6 +42,92 @@ final class PostgresClientTests: XCTestCase {
...
+    func testTransaction() async throws {
+        var mlogger = Logger(label: "test")
+        mlogger.logLevel = .debug
+        let logger = mlogger
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 8)
+        self.addTeardownBlock {
+            try await eventLoopGroup.shutdownGracefully()
+        }
+        
+        let tableName = "test_client_trasactions"
```
fabianfett, line 55:
```suggestion
        let tableName = "test_client_transactions"
```

Tests/IntegrationTests/PostgresClientTests.swift
```diff
@@ -42,6 +42,92 @@ final class PostgresClientTests: XCTestCase {
...
+                /// Test roll back
+                taskGroup.addTask {
+                    let _ = try await client.withTransaction(logger: logger) { transaction in
+                        try await transaction.query(
+                            """
+                            INSERT INTO "\(unescaped: tableName)" (uuid) VALUES (\(iterations));
+                            """,
+                            logger: logger
+                        )
+                    }
```
fabianfett, line 109: You don't check the error here at all. There is also nothing here to rollback.

fabianfett (commented):
Added some more comments.

Tests/IntegrationTests/PostgresClientTests.swift
thoven87: Tests have been updated. I apologize about skipping this part.

Sources/PostgresNIO/Pool/PostgresClient.swift
```diff
@@ -307,6 +307,28 @@ public final class PostgresClient: Sendable, ServiceLifecycle.Service {
...
+    
+    /// Lease a connection for the provided `closure`'s lifetime.
+    /// A transation starts with call to withConnection
+    /// A transaction should end with a call to COMMIT or ROLLBACK
+    /// COMMIT is called upon successful completion and ROLLBACK is called should any steps fail
+    ///
+    /// - Parameter closure: A closure that uses the passed `PostgresConnection`. The closure **must not** capture
+    ///                      the provided `PostgresConnection`.
+    /// - Returns: The closure's return value.
+    public func withTransaction<Result>(_ process: (PostgresConnection) async throws -> Result) async throws -> Result {
```
fabianfett, line 319: I wonder if we should capture calling file and line here. Then we could attach that info to the error that is thrown. We would wrap the thrown error in a `PostgresTransactionError`. We could also attach the Rollback error, if that happens. cc @gwynne WDYT?

gwynne: Makes sense to me, yeah.

fabianfett:
Thanks @thoven87! Let's merge this and I'll do the error handling in a follow up!

fabianfett (approved):

PR #526, opened 2024-12-08
Allow TLS enabled connections when providing an established channel

tkrajacic:
When connecting through an established channel (e.g. an SSH tunnel) it can still be useful to allow postgres to connect using TLS. This PR exposes the tls parameter on the connection method for using an established channel.

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
@@ -193,8 +193,8 @@ extension PostgresConnection {
         ///   - channel: The `NIOCore/Channel` to use. The channel must already be active and connected to an
         ///     endpoint (i.e. `NIOCore/Channel/isActive` must be `true`).
         ///   - tls: The TLS mode to use. Defaults to ``TLS-swift.struct/disable``.
-        public init(establishedChannel channel: Channel, username: String, password: String?, database: String?) {
-            self.init(endpointInfo: .configureChannel(channel), tls: .disable, username: username, password: password, database: database)
+        public init(establishedChannel channel: Channel, tls: PostgresConnection.Configuration.TLS = .disable, username: String, password: String?, database: String?) {
+            self.init(endpointInfo: .configureChannel(channel), tls: tls, username: username, password: password, database: database)
         }
 
         // MARK: - Implementation details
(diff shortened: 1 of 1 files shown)


Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
```diff
@@ -193,8 +193,8 @@ extension PostgresConnection {
         ///   - channel: The `NIOCore/Channel` to use. The channel must already be active and connected to an
         ///     endpoint (i.e. `NIOCore/Channel/isActive` must be `true`).
         ///   - tls: The TLS mode to use. Defaults to ``TLS-swift.struct/disable``.
-        public init(establishedChannel channel: Channel, username: String, password: String?, database: String?) {
-            self.init(endpointInfo: .configureChannel(channel), tls: .disable, username: username, password: password, database: database)
+        public init(establishedChannel channel: Channel, tls: PostgresConnection.Configuration.TLS = .disable, username: String, password: String?, database: String?) {
```
fabianfett, line 196: This is a breaking API change. Would you mind adding an additional initializer? This initializer should then call the new one that you just created.

Do you feel comfortable to add a unit test using a `NIOAsyncTestingChannel`? An example can be found here: 
https://github.com/vapor/postgres-nio/blob/f2a6394a2e7157d547727b975fc0328b92f89fb1/Tests/PostgresNIOTests/New/PostgresConnectionTests.swift#L41C1-L82C1

fabianfett (commented):
Great addition! Thanks for opening a PR. Please see the comment! If you need help please reach out.

Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
tkrajacic: I probably won't have time to do the test right now, as I need to fix the wrong channel handler order first for TLS 🤭

fabianfett: > I need to fix the wrong channel handler order first for TLS

Not sure I follow here.

tkrajacic: See #527

fabianfett (approved):
In an ideal case we would add at least one unit test for this...

PR #527, opened 2024-12-08
Correctly place the SSL channel handler in front of the PostgresChannelHandler

tkrajacic:
Ensure that `NIOSSLClientHandler` is added immediately before `PostgresChannelHandler` instead of at the front of the entire pipeline. 
This ensures that TLS works correctly when working with preexisting channels, such as tunneling connections with `NIOSSH`. See also #526 .

Thanks to @Joannis and @gwynne for finding the problem and pointing me to the right place to fix it.

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -64,14 +64,14 @@ public final class PostgresConnection: @unchecked Sendable {
         
         switch configuration.tls.base {
         case .prefer(let context), .require(let context):
-            configureSSLCallback = { channel, postgreChannelHandler in
+            configureSSLCallback = { channel, postgresChannelHandler in
                 channel.eventLoop.assertInEventLoop()
 
                 let sslHandler = try NIOSSLClientHandler(
                     context: context,
                     serverHostname: configuration.serverNameForTLS
                 )
-                try channel.pipeline.syncOperations.addHandler(sslHandler, position: .before(postgreChannelHandler))
+                try channel.pipeline.syncOperations.addHandler(sslHandler, position: .before(postgresChannelHandler))
             }
         case .disable:
             configureSSLCallback = nil
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -60,18 +60,18 @@ public final class PostgresConnection: @unchecked Sendable {
     func start(configuration: InternalConfiguration) -> EventLoopFuture<Void> {
         // 1. configure handlers
 
-        let configureSSLCallback: ((Channel) throws -> ())?
+        let configureSSLCallback: ((Channel, PostgresChannelHandler) throws -> ())?
         
         switch configuration.tls.base {
         case .prefer(let context), .require(let context):
-            configureSSLCallback = { channel in
+            configureSSLCallback = { channel, postgreChannelHandler in
                 channel.eventLoop.assertInEventLoop()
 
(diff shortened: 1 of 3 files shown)


Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -60,18 +60,18 @@ public final class PostgresConnection: @unchecked Sendable {
     func start(configuration: InternalConfiguration) -> EventLoopFuture<Void> {
         // 1. configure handlers
 
-        let configureSSLCallback: ((Channel) throws -> ())?
+        let configureSSLCallback: ((Channel, PostgresChannelHandler) throws -> ())?
         
         switch configuration.tls.base {
         case .prefer(let context), .require(let context):
-            configureSSLCallback = { channel in
+            configureSSLCallback = { channel, postgreChannelHandler in
```
fabianfett, line 67:
```suggestion
            configureSSLCallback = { channel, postgresChannelHandler in
```

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -60,18 +60,18 @@ public final class PostgresConnection: @unchecked Sendable {
...
                 channel.eventLoop.assertInEventLoop()
 
                 let sslHandler = try NIOSSLClientHandler(
                     context: context,
                     serverHostname: configuration.serverNameForTLS
                 )
-                try channel.pipeline.syncOperations.addHandler(sslHandler, position: .first)
+                try channel.pipeline.syncOperations.addHandler(sslHandler, position: .before(postgreChannelHandler))
```
fabianfett, line 74:
```suggestion
                try channel.pipeline.syncOperations.addHandler(sslHandler, position: .before(postgresChannelHandler))
```

Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
```diff
@@ -192,9 +192,22 @@ extension PostgresConnection {
...
+        ///
+        /// This is provided for calling code which wants to manage the underlying connection transport on its
+        /// own, such as when tunneling a connection through SSH.
+        ///
+        /// - Parameters:
+        ///   - channel: The `NIOCore/Channel` to use. The channel must already be active and connected to an
+        ///     endpoint (i.e. `NIOCore/Channel/isActive` must be `true`).
         public init(establishedChannel channel: Channel, username: String, password: String?, database: String?) {
-            self.init(endpointInfo: .configureChannel(channel), tls: .disable, username: username, password: password, database: database)
+            self.init(establishedChannel: channel, tls: .disable, username: username, password: password, database: database)
```
fabianfett, line 210: Let's revert this here and make this change in #526, which we can land after #527.

fabianfett (commented):
Great patch! See comments. Pretty sure you will need to fix some tests.

Sources/PostgresNIO/Connection/PostgresConnection+Configuration.swift
tkrajacic: This PR is useless without #526 as far as I understand

fabianfett: This PR enables #526 to work correctly. Therefore we should land it first and then land #526. But we should not add the changes that we want to make in #526 here.

tkrajacic: I needed a working branch locally so I had to base one on the other branch 🙃

fabianfett (approved):
Great patch! Thanks!

PR #541, opened 2025-03-11
Update README.md

Joannis:
The readme has a broken example snippet for running queries which is _extremely_ confusing if you try to use it.
The result of this omission results in trying to run a query which leads users to use a `String` overload rather than the `PostgresQuery` overload which is not only SQL injection vulnerable but also returns an `EventLoopFuture<PostgresResult>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ await withTaskGroup(of: Void.self) { taskGroup in
 Once a client is running, queries can be sent to the server. This is straightforward:
 
 ```swift
-let rows = try await client.query("SELECT id, username, birthday FROM users")
+let rows = try await client.query("SELECT id, username, birthday FROM users", logger: Logger(label: "my-example-logger"))
 ```
 
 The query will return a [`PostgresRowSequence`], which is an AsyncSequence of [`PostgresRow`]s. 


gwynne (approved):

README.md
```diff
@@ -107,7 +107,7 @@ await withTaskGroup(of: Void.self) { taskGroup in
 Once a client is running, queries can be sent to the server. This is straightforward:
 
 ```swift
-let rows = try await client.query("SELECT id, username, birthday FROM users")
+let rows = try await client.query("SELECT id, username, birthday FROM users", logger: Logger(label: "my-example-logger"))
```
fabianfett, line 110: I don't think this is true:
```
    @discardableResult
    public func query(
        _ query: PostgresQuery,
        logger: Logger? = nil,
        file: String = #fileID,
        line: Int = #line
    ) async throws -> PostgresRowSequence {
```
@Joannis can you validate that?

fabianfett: @Joannis PING!

Joannis: I must've pulled in an old version maybe? You're right

fabianfett: I think you talk about PostgresConnection, which implements PostgresDatabase.

PR #552, opened 2025-05-12
Add `@inlinable` to `ConnectionPool.run()`

fabianfett:
Previously the `ConnectionPool.run()` method wasn't marked as `@inlinable`, because of this we missed an opportunity to specialize the code that is run as part of the events.


gwynne (approved):
I assume this comes with meaning performance improvements.

PR #553, opened 2025-05-14
Make SASL faster

fabianfett:


gwynne (approved):
Any hint of how much faster?

PR #566, opened 2025-06-24
WIP: Implement `COPY … FROM STDIN`

### LINKED ISSUE #290: Support Copy In Mode
For fast bulk inserts we should support COPY IN mode.

ahoppen:
This is still WIP. To do items include:
- [ ] Test the various error cases, I have mostly focused on the success case so far
- [ ] Test the backpressure support
- [ ] Change the public API to accept the table + columns to copy into as well as options so that we can build the `COPY` query instead of letting the user write it
- [ ] Add an API that allows binary transfer of data

Fixes #290

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -694,6 +694,100 @@ extension PostgresConnection {
     }
 }
 
+// MARK: Copy from
+
+fileprivate extension EventLoop {
+    /// If we are on the given event loop, execute `task` immediately. Otherwise schedule it for execution.
+    func executeImmediatelyOrSchedule(_ task: @Sendable @escaping () -> Void) {
+        if inEventLoop {
+            return task()
+        }
+        return execute(task)
+    }
+}
+
+/// A handle to send 
+public struct PostgresCopyFromWriter: Sendable {
+    private let channelHandler: NIOLoopBound<PostgresChannelHandler>
+    private let context: NIOLoopBound<ChannelHandlerContext>
+    private let eventLoop: any EventLoop
+
+    struct NotWritableError: Error, CustomStringConvertible {
+        var description = "No data must be written to `PostgresCopyFromWriter` after it has sent a CopyDone or CopyFail message, ie. after the closure producing the copy data has finished"
+    }
+
+    init(handler: PostgresChannelHandler, context: ChannelHandlerContext, eventLoop: any EventLoop) {
+        self.channelHandler = NIOLoopBound(handler, eventLoop: eventLoop)
+        self.context = NIOLoopBound(context, eventLoop: eventLoop)
+        self.eventLoop = eventLoop
+    }
+
+    /// Send data for a `COPY ... FROM STDIN` operation to the backend.
+    public func write(_ byteBuffer: ByteBuffer) async throws {
+        await withCheckedContinuation { (continuation: CheckedContinuation<Void, Never>) in 
+            eventLoop.executeImmediatelyOrSchedule {
+                self.channelHandler.value.copyData(byteBuffer, context: self.context.value, readyForMoreWriteContinuation: continuation)
+            }
+        }
+    }
+
+    /// Finalize the data transfer, putting the state machine out of the copy mode and sending a `CopyDone` message to 
+    /// the backend.
+    func done() async throws {
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, any Error>) in 
+            eventLoop.executeImmediatelyOrSchedule {
+                self.channelHandler.value.sendCopyDone(continuation: continuation, context: self.context.value)
+            }
+        }
+    }
+
+    /// Finalize the data transfer, putting the state machine out of the copy mode and sending a `CopyFail` message to 
+    /// the backend.
+    func failed(error: any Error) async throws {
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, any Error>) in 
+            eventLoop.executeImmediatelyOrSchedule {
+                self.channelHandler.value.sendCopyFailed(message: "\(error)", continuation: continuation, context: self.context.value)
+            }
+        }
+    }
+}
+
+extension PostgresConnection {
+    // TODO: Instead of an arbitrary query, make this a structured data structure.
+    // TODO: Write doc comment
+    public func copyFrom(
+        _ query: PostgresQuery,
+        writeData: @escaping @Sendable (PostgresCopyFromWriter) async throws -> Void,
+        logger: Logger,
+        file: String = #fileID,
+        line: Int = #line
+    )  async throws {
+        var logger = logger
+        logger[postgresMetadataKey: .connectionID] = "\(self.id)"
+        guard query.binds.count <= Int(UInt16.max) else {
+            throw PSQLError(code: .tooManyParameters, query: query)
+        }
+
+        let writer = try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<PostgresCopyFromWriter, any Error>) in
+            let context = ExtendedQueryContext(
+                copyFromQuery: query,
+                triggerCopy: continuation,
+                logger: logger
+            )
+            self.channel.write(HandlerTask.extendedQuery(context), promise: nil)
+        }
+
+        do {
+            try await writeData(writer)
+        } catch {
+            try await writer.failed(error: error)
+            throw error
+        }
+        try await writer.done()
+    }
+
+}
+
 // MARK: PostgresDatabase conformance
 
 extension PostgresConnection: PostgresDatabase {
--- a/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
+++ b/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
@@ -88,7 +88,23 @@ struct ConnectionStateMachine {
         case sendParseDescribeBindExecuteSync(PostgresQuery)
         case sendBindExecuteSync(PSQLExecuteStatement)
         case failQuery(EventLoopPromise<PSQLRowStream>, with: PSQLError, cleanupContext: CleanUpContext?)
+        /// Fail a query's execution by throwing an error on the given continuation.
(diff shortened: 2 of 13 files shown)


Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -694,6 +694,100 @@ extension PostgresConnection {
...
+            return task()
+        }
+        return execute(task)
+    }
+}
+
+/// A handle to send 
+public struct PostgresCopyFromWriter: Sendable {
+    private let channelHandler: NIOLoopBound<PostgresChannelHandler>
+    private let context: NIOLoopBound<ChannelHandlerContext>
```
fabianfett, line 712: use the `handlerContext: ChannelHandlerContext?` with `!` in `PostgresChannelHandler` instead.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -694,6 +694,100 @@ extension PostgresConnection {
...
+    private let eventLoop: any EventLoop
+
+    struct NotWritableError: Error, CustomStringConvertible {
+        var description = "No data must be written to `PostgresCopyFromWriter` after it has sent a CopyDone or CopyFail message, ie. after the closure producing the copy data has finished"
+    }
```
fabianfett, line 717: this isn't used anywhere.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -694,6 +694,100 @@ extension PostgresConnection {
...
+        }
+    }
+}
+
+extension PostgresConnection {
+    // TODO: Instead of an arbitrary query, make this a structured data structure.
+    // TODO: Write doc comment
+    public func copyFrom(
+        _ query: PostgresQuery,
+        writeData: @escaping @Sendable (PostgresCopyFromWriter) async throws -> Void,
```
fabianfett, line 760: closure should be last argument.

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -480,6 +597,16 @@ struct ExtendedQueryStateMachine {
...
         }
     }
+
+    mutating func channelWritabilityChanged(isWritable: Bool) {
+        guard case .copyingData(.pendingBackpressureRelieve(let continuation)) = state else {
+            return
+        }
+        self.avoidingStateMachineCoW { state in
+            state = .copyingData(.readyToSend)
+            continuation.resume()
```
fabianfett, line 607: side effect triggered in state machine. return the continuation out and succeed it in the PostgresChannelHandler.

Sources/PostgresNIO/Connection/PostgresConnection.swift
fabianfett: You don't need to pass the context to `sendCopyDone` or `copyData`. Instead use the context that already exists in the `PostgresChannelHandler.handlerContext`.

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -88,7 +88,23 @@ struct ConnectionStateMachine {
         case sendParseDescribeBindExecuteSync(PostgresQuery)
         case sendBindExecuteSync(PSQLExecuteStatement)
         case failQuery(EventLoopPromise<PSQLRowStream>, with: PSQLError, cleanupContext: CleanUpContext?)
+        /// Fail a query's execution by throwing an error on the given continuation.
+        case failQueryContinuation(any AnyErrorContinuation, with: PSQLError, cleanupContext: CleanUpContext?)
```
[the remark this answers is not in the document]
ahoppen: The alternative is to have two cases, one for `CheckedContinuation<PostgresCopyFromWriter, any Error>` and one for `CheckedContinuation<Void, any Error>`. Since this is the failure case, it’s not performance sensitive and I’d argue that the improved readability is worth the extra allocation in the failure case. Happy to adjust if you disagree though.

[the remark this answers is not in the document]
fabianfett: Can we make this an enum instead? Best of both worlds?

Sources/PostgresNIO/Connection/PostgresConnection.swift
ahoppen: Oops, not sure how I missed that.

Sources/PostgresNIO/New/PostgresFrontendMessageEncoder.swift
```diff
@@ -167,6 +167,25 @@ struct PostgresFrontendMessageEncoder {
         self.buffer.writeMultipleIntegers(UInt32(8), Self.sslRequestCode)
     }
 
+    mutating func copyData(data: ByteBuffer) {
+        self.clearIfNeeded()
+        self.buffer.psqlWriteMultipleIntegers(id: .copyData, length: UInt32(data.readableBytes))
+        self.buffer.writeImmutableBuffer(data)
```
fabianfett, line 173: I wonder if in this case, we should not copy the byteBuffer here. Instead we could forward the OG buffer, after we have sent the head. Would be two NIO write calls followed by a single flush. Would save at least one unnecessary copy call.

Sources/PostgresNIO/New/PostgresChannelHandler.swift
```diff
@@ -169,10 +171,61 @@ final class PostgresChannelHandler: ChannelDuplexHandler {
...
+    func sendCopyFailed(message: String, continuation: CheckedContinuation<Void, any Error>) {
+        let action = self.state.sendCopyFail(message: message, continuation: continuation)
+        self.run(action, with: self.handlerContext!)
+    }
+
+    /// Send a `Sync` message to the backend.
+    func sendSync() {
+        self.encoder.sync()
+        self.handlerContext!.writeAndFlush(self.wrapOutboundOut(self.encoder.flushBuffer()), promise: nil)
+    }
```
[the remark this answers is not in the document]
ahoppen: The only place we call this is if we have `failContinuation` with `sync: true`, which is the case when we throw a `CopyCancellationError` because the server terminated the copy. In this case no `CopyDone` or `CopyFail` will be sent from the client, so no `CopyDone` or `CopyFail` message will be sent.

Sources/PostgresNIO/New/PostgresFrontendMessageEncoder.swift
fabianfett: Love that!

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -0,0 +1,234 @@
+/// Handle to send data for a `COPY ... FROM STDIN` query to the backend.
+public struct PostgresCopyFromWriter: Sendable {
+    /// The backend failed the copy data transfer, which means that no more data sent by the frontend would be processed.
+    ///
+    /// The `PostgresCopyFromWriter` should cancel the data transfer.
+    public struct CopyCancellationError: Error {
```
fabianfett, line 6: Where is the benefit of wrapping this here? Shouldn't we use the PSQLError directly?

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -0,0 +1,234 @@
...
+        ///
+        /// See the `DELIMITER` option in Postgres's `COPY` command.
+        ///
+        /// Uses the default delimiter of the format
+        public var delimiter: UnicodeScalar? = nil
+
+        public init() {}
+    }
+
+    case text(TextOptions)
```
fabianfett, line 129: should we mention what the default delimiter is?

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -315,17 +397,82 @@ struct ExtendedQueryStateMachine {
...
+            preconditionFailure("Not ready to send data")
+        }
+        if channelIsWritable {
+            promise.succeed()
+            return
+        }
+        return avoidingStateMachineCoW { state in
+            state = .copyingData(.pendingBackpressureRelieve(promise))
+        }
+    }
```
fabianfett, line 443: Instead of fulfilling the promise here we should communicate with an action.

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -0,0 +1,230 @@
...
+            continuation.resume(with: result)
+        }
+    }
+
+    #if compiler(>=6.0)
+    /// Send data for a `COPY ... FROM STDIN` operation to the backend.
+    ///
+    /// - Throws: If an error occurs during the write of if the backend sent an `ErrorResponse` during the copy 
+    ///   operation, eg. to indicate that a **previous** `write` call had an invalid format.
+    public func write(_ byteBuffer: ByteBuffer, isolation: isolated (any Actor)? = #isolation) async throws {
```
fabianfett: lets remove the `isolation` calls here and instead adopt `nonisolated nonsending` for the whole package.
```suggestion
    public func write(_ byteBuffer: ByteBuffer) async throws {
```

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -88,7 +88,33 @@ struct ConnectionStateMachine {
         case sendParseDescribeBindExecuteSync(PostgresQuery)
         case sendBindExecuteSync(PSQLExecuteStatement)
         case failQuery(EventLoopPromise<PSQLRowStream>, with: PSQLError, cleanupContext: CleanUpContext?)
+        /// Fail a query's execution by resuming the continuation with the given error. When `sync` is `true`, send a
+        /// `Sync` message to the backend.
+        case failQueryContinuation(AnyErrorContinuation, with: PSQLError, sync: Bool, cleanupContext: CleanUpContext?)
+        /// Fail a query's execution by resuming the continuation with the given error and send a `Sync` message to the
+        /// backend.
```
fabianfett: That comment doesnt make sense for the below case.

PR #568, opened 2025-07-07
Fix: Correctly decode jsonb to String by stripping version byte

### LINKED ISSUE #562: When loading data type as jsonb, an extra unicode will be added: \u0001
### Describe the issue

loading data type as `jsonb`, an extra unicode will be added: \u0001 (show in json result) or \u{01} (show in Xcode)

### Vapor version

1.26.2

macOS 26 DP Beta 1 and Ubuntu Linux 22.04

### Swift version

macOS: 6.1.2, Ubuntu: swift-6.1.0-release

### Steps to reproduce

1. Upgrade all packages to latest version;
2. Connect PostgreSQL and query a type as `jsonb` field;
3. the result ByteBuffer decode to String.self, an additional \u{01} will be displayed (in Xcode Console); return it as json to response, it will show \u0001 (additional)

- in Xcode Console
Image

- in Postman response raw
Image

- in DB (it is correct)
Image

When I downgraded back to `1.24.0`, everything worked fine!!

### Outcome

_No response_

### Additional notes

_No response_

nikodittmar:
resolves #562 

## Description
This pull request resolves an issue where decoding a `jsonb` column into a `String` would incorrectly add a `\u{01}` version byte to the resulting string.

## Changes
- Added a failing test case to reproduce the bug, which now passes.
- Added a specific case for `(.binary, .jsonb)` to the `PostgresDecodable` conformance for `String` within `String+PostgresCodable.swift`. The new case reads and discards the version byte before decoding the remainder of the buffer.

### DIFF
--- a/Tests/PostgresNIOTests/New/Data/String+PSQLCodableTests.swift
+++ b/Tests/PostgresNIOTests/New/Data/String+PSQLCodableTests.swift
@@ -52,4 +52,15 @@ class String_PSQLCodableTests: XCTestCase {
             XCTAssertEqual($0 as? PostgresDecodingError.Code, .failure)
         }
     }
+
+    func testDecodeFromJSONB() {
+        let json = #"{"hello": "world"}"#
+        var buffer = ByteBuffer()
+        buffer.writeInteger(UInt8(1))
+        buffer.writeString(json)
+
+        var decoded: String?
+        XCTAssertNoThrow(decoded = try String(from: &buffer, type: .jsonb, format: .binary, context: .default))
+        XCTAssertEqual(decoded, json)
+    }
 }


0xTim:
This seems to break some tests I have on a client project, including assertion failures like:

```
XCTAssertEqual failed: ("["bool": false, "array": [true, 4.56, "bar"], "number": 1.23, "object": ["string2": "baz"], "string": "foo", "empty": null]") is not equal to (""{\"bool\": false, \"array\": [true, 4.56, \"bar\"], \"empty\": null, \"number\": 1.23, \"object\": {\"string2\": \"baz\"}, \"string\": \"foo\"}"")
```

i.e. previously it was an object and now it's a JSON string

PR #573, opened 2025-07-24
Support binary data transfer in `COPY FROM`

ahoppen:
My benchmark of transferring the integers from 0 to 1,000,000 both as an integer and as a string was about the same speed as the old text-based transfer. I believe that the binary transfer will start to show significant benefits when transferring binary data, other fields that don't need to be represented as fields and also means that the user doesn't need to worry about escapping their data.

A call to binary transfer data looks as follows

```swift
let records: [(id: Int, name: String)] = [(1, "Alice"), (42, "Bob")]
try await conn.copyFromBinary(table: "copy_table", columns: ["id", "name"], logger: .psqlTest) { writer in
    for record in records {
        try await writer.writeRow { columnWriter in
            try columnWriter.writeColumn(Int32(record.id))
            try columnWriter.writeColumn(record.name)
        }
    }
}
```

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
@@ -98,17 +98,15 @@ public struct PostgresCopyFromWriter: Sendable {
     }
 }
 
-
+// PostgresBinaryCopyFromWriter relies on non-Escapable types, which were only introduced in Swift 6.2
+#if compiler(>=6.2)
 /// Handle to send binary data for a `COPY ... FROM STDIN` query to the backend.
 ///
 /// It takes care of serializing `PostgresEncodable` column types into the binary format that Postgres expects.
 public struct PostgresBinaryCopyFromWriter: ~Copyable {
     /// Handle to serialize columns into a row that is being written by `PostgresBinaryCopyFromWriter`.
-    public struct ColumnWriter: ~Copyable {
-        /// The `PostgresBinaryCopyFromWriter` that is gathering the serialized data.
-        ///
-        /// We need to model this as `UnsafeMutablePointer` because we can't express in the Swift type system that
-        /// `ColumnWriter` never exceeds the lifetime of `PostgresBinaryCopyFromWriter`.
+    public struct ColumnWriter: ~Escapable, ~Copyable {
+        /// Pointer to the `PostgresBinaryCopyFromWriter` that is gathering the serialized data.
         @usableFromInline
         let underlying: UnsafeMutablePointer<PostgresBinaryCopyFromWriter>
 
@@ -116,9 +114,26 @@ public struct PostgresBinaryCopyFromWriter: ~Copyable {
         @usableFromInline
         var columns: UInt16 = 0
 
+        /// - Warning: Do not call directly, call `withColumnWriter` instead
         @usableFromInline
-        init(underlying: UnsafeMutablePointer<PostgresBinaryCopyFromWriter>) {
-            self.underlying = underlying
+        init(_underlying: UnsafeMutablePointer<PostgresBinaryCopyFromWriter>) {
+            self.underlying = _underlying
+        }
+
+        @usableFromInline
+        static func withColumnWriter<T, E: Error>(
+            writingTo underlying: inout PostgresBinaryCopyFromWriter,
+            body: (inout ColumnWriter) throws(E) -> T
+        ) throws(E) -> T {
+            return try withUnsafeMutablePointer(to: &underlying) { pointerToUnderlying throws(E) in
+                // We can guarantee that `ColumWriter` never outlives `underlying` because `ColumnWriter` is
+                // `~Escapable` and thus cannot escape the context of the closure to `withUnsafeMutablePointer`.
+                // To model this without resorting to unsafe pointers, we would need to be able to declare an `inout`
+                // reference to `PostgresBinaryCopyFromWriter` as a member of `ColumnWriter`, which isn't possible at
+                // the moment (https://github.com/swiftlang/swift/issues/85832).
+                var columnWriter = ColumnWriter(_underlying: pointerToUnderlying)
+                return try body(&columnWriter)
+            }
         }
 
         /// Serialize a single column to a row.
@@ -128,16 +143,19 @@ public struct PostgresBinaryCopyFromWriter: ~Copyable {
         ///   be called with an `Int32`. Serializing an integer of a different width will cause a deserialization
         ///   failure in the backend.
         @inlinable
+        #if compiler(<6.3)
+        @_lifetime(&self)
+        #endif
         public mutating func writeColumn(_ column: (some PostgresEncodable)?) throws {
             columns += 1
             try invokeWriteColumn(on: underlying, column)
         }
 
-        // Needed to work around https://github.com/swiftlang/swift/issues/83309, copying the implementation into 
+        // Needed to work around https://github.com/swiftlang/swift/issues/83309, copying the implementation into
         // `writeColumn` causes an assertion failure when thread sanitizer is enabled.
-        @inlinable 
+        @inlinable
         func invokeWriteColumn(
-            on writer: UnsafeMutablePointer<PostgresBinaryCopyFromWriter>, 
+            on writer: UnsafeMutablePointer<PostgresBinaryCopyFromWriter>,
             _ column: (some PostgresEncodable)?
         ) throws {
             try writer.pointee.writeColumn(column)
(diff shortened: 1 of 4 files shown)


Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -98,6 +98,131 @@ public struct PostgresCopyFromWriter: Sendable {
...
+        self.underlying = underlying
+        // Allocate 10% more than the buffer size because we only flush the buffer once it has exceeded `bufferSize`
+        buffer.reserveCapacity(bufferSize + bufferSize / 10)
+        self.bufferSize = bufferSize
+    }
+
+    /// Serialize a single row to the backend. Call `writeColumn` on `columnWriter` for every column that should be
+    /// included in the row.
+    @inlinable
+    public mutating func writeRow(_ body: (_ columnWriter: inout ColumnWriter) throws -> Void) async throws {
```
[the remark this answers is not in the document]
FranzBusch: > We should probably make this nonsiolated(nonsending)

Ignore this. The package has adopted the upcoming language feature

FranzBusch (commented):
LGTM from my side. @gwynne @fabianfett mind reviewing and approving?

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -162,6 +302,52 @@ private func buildCopyFromQuery(
...
+        table: String,
+        columns: [String] = [],
+        options: PostgresCopyFromFormat.BinaryOptions = .init(),
+        bufferSize: Int = 100_000,
+        logger: Logger,
+        file: String = #fileID,
+        line: Int = #line,
+        writeData: (inout PostgresBinaryCopyFromWriter) async throws -> Void
+    )  async throws {
+        try await copyFrom(table: table, columns: columns, format: .binary(PostgresCopyFromFormat.BinaryOptions()), logger: logger) { writer in
```
ptoffy: shouldn't this be
```suggestion
        try await copyFrom(table: table, columns: columns, format: .binary(options), logger: logger) { writer in
```

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -98,6 +98,132 @@ public struct PostgresCopyFromWriter: Sendable {
...
+                // the moment (https://github.com/swiftlang/swift/issues/85832).
+                var columnWriter = ColumnWriter(_underlying: pointerToUnderlying)
+                return try body(&columnWriter)
+            }
+        }
+
+        /// Serialize a single column to a row.
+        ///
+        /// - Important: It is critical that that data type encoded here exactly matches the data type in the
+        ///   database. For example, if the database stores an a 4-bit integer the corresponding `writeColumn` must
```
ptoffy:
```suggestion
        ///   database. For example, if the database stores an a 4-byte integer the corresponding `writeColumn` must
```

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -162,6 +302,52 @@ private func buildCopyFromQuery(
...
+            // Flag fields
+            header.writeInteger(UInt32(0))
+
+            // Header extension area length
+            header.writeInteger(UInt32(0))
+            try await writer.write(header)
+
+            var binaryWriter = PostgresBinaryCopyFromWriter(underlying: writer, bufferSize: bufferSize)
+            try await writeData(&binaryWriter)
+            try await binaryWriter.flush()
```
ptoffy, line 348: This is missing the final file trailer as explained in https://www.postgresql.org/docs/current/sql-copy.html

```suggestion
            try await writeData(&binaryWriter)
            binaryWriter.buffer.writeInteger(Int16(-1))
            try await binaryWriter.flush()
```

ptoffy (changes requested):

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -98,6 +98,132 @@ public struct PostgresCopyFromWriter: Sendable {
     }
 }
 
+// PostgresBinaryCopyFromWriter relies on non-Escapable types, which were only introduced in Swift 6.2
+#if compiler(>=6.2)
+/// Handle to send binary data for a `COPY ... FROM STDIN` query to the backend.
+///
+/// It takes care of serializing `PostgresEncodable` column types into the binary format that Postgres expects.
```
gwynne:
```suggestion
/// It takes care of serializing ``PostgresEncodable`` column types into the binary format that Postgres expects.
```

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -98,6 +98,132 @@ public struct PostgresCopyFromWriter: Sendable {
...
+public struct PostgresBinaryCopyFromWriter: ~Copyable {
+    /// Handle to serialize columns into a row that is being written by `PostgresBinaryCopyFromWriter`.
```
gwynne:
```suggestion
    /// Handle to serialize columns into a row that is being written by ``PostgresBinaryCopyFromWriter``.
```

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -98,6 +98,132 @@ public struct PostgresCopyFromWriter: Sendable {
...
+    public struct ColumnWriter: ~Escapable, ~Copyable {
+        /// Pointer to the `PostgresBinaryCopyFromWriter` that is gathering the serialized data.
```
gwynne:
```suggestion
        /// Pointer to the ``PostgresBinaryCopyFromWriter`` that is gathering the serialized data.
```
And so on, I'm not gonna comment for every single instance of this. 😆

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -98,6 +98,132 @@ public struct PostgresCopyFromWriter: Sendable {
...
+        @usableFromInline
+        let underlying: UnsafeMutablePointer<PostgresBinaryCopyFromWriter>
+
+        /// The number of columns that have been written by this `ColumnWriter`.
+        @usableFromInline
+        var columns: UInt16 = 0
+
+        /// - Warning: Do not call directly, call `withColumnWriter` instead
```
gwynne:
```suggestion
        /// > Warning: Do not call this directly; call ``ColumnWriter/withColumnWriter(writingTo:body:)`` instead.
```

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -98,6 +98,132 @@ public struct PostgresCopyFromWriter: Sendable {
...
+                return try body(&columnWriter)
+            }
+        }
+
+        /// Serialize a single column to a row.
+        ///
+        /// - Important: It is critical that that data type encoded here exactly matches the data type in the
+        ///   database. For example, if the database stores an a 4-byte integer the corresponding `writeColumn` must
+        ///   be called with an `Int32`. Serializing an integer of a different width will cause a deserialization
+        ///   failure in the backend.
```
gwynne:
```suggestion
        /// > Important: It is critical that that data type encoded here exactly matches the data type in the
        /// > database. For example, if the database stores an a 4-byte integer the corresponding `writeColumn` must
        /// > be called with an `Int32`. Serializing an integer of a different width will cause a deserialization
        /// > failure in the backend.
```

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -98,6 +98,132 @@ public struct PostgresCopyFromWriter: Sendable {
...
+        // Fill in the number of columns
+        buffer.setInteger(columns, at: columnIndex)
+
+        if buffer.readableBytes > bufferSize {
+            try await flush()
+        }
+        return bodyResult
+    }
+
+    /// Serialize a single column to the buffer. Should only be called by `ColumnWriter`.
```
gwynne: Given this comment, should this method be `fileprivate`? Or would that mess with the inlinability?

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
```diff
@@ -162,6 +302,53 @@ private func buildCopyFromQuery(
 }
 
 extension PostgresConnection {
+    #if compiler(>=6.2)
```
gwynne, line 305: I'd be open to just bumping the package's minimum Swift requirement to 6.2 a bit early. 6.4 isn't far away anyway.

gwynne (changes requested):

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
ptoffy: It would mess with inlinability

Sources/PostgresNIO/Connection/PostgresConnection+CopyFrom.swift
ptoffy: I'd leave it as is for now, but you've got the last word :)

gwynne: It can stay like this. Doesn't make a ton of difference either way. 🙂

gwynne (approved):

ptoffy (approved):

ahoppen:
Thanks for reviewing and merging this.

Issue #576, opened 2025-08-13
Expose Column Metadata on `PostgresRowSequence`


SteveLauC:
Hey, thanks for providing and maintaining this lib!

I would like to know how to get the column names of a query result. I do take a look at the doc, but it looks like this is not covered.

Reading the code, both `PostgresRowSequence` and `PreparedQuery` store column metadata:

https://github.com/vapor/postgres-nio/blob/ca70d8cd4d50509afefb8a791144ff126600df3e/Sources/PostgresNIO/New/PostgresRowSequence.swift#L14
https://github.com/vapor/postgres-nio/blob/ca70d8cd4d50509afefb8a791144ff126600df3e/Sources/PostgresNIO/Connection/PostgresDatabase%2BPreparedQuery.swift#L32

 but the fields are private, and there are no public interfaces that can be used to get the information.

Thanks!

fabianfett:
Hi 👋 ,

you can get the information in the following way:

```swift
for try await row in rowSequence {
    for cell in row {
        cell.columnName // this is what you are looking for
    }
}
```

Hope that helps.

fabianfett:
That is true, I'm happy to accept a pr that exposes the column metadata on the row sequence. However we will need a new type for that. `RowDescription.Column` is required to stay internal.

PR #577, opened 2025-08-14
feat: add PostgresRowSequence.getColumns() to get column metadata

SteveLauC:
This commit implements `PostgresRowSequence.getColumns()` to enable users to retrieve column metadata of their query results.

Discussion: https://github.com/vapor/postgres-nio/issues/576

### DIFF
--- a/Sources/PostgresNIO/New/PostgresColumn.swift
+++ b/Sources/PostgresNIO/New/PostgresColumn.swift
@@ -1,0 +1,46 @@
+/// Information of a column.
+//
+// This type has the same definition as `RowDescription.column`, we need to keep 
+// that type private so we defines this type.
+public struct PostgresColumn: Hashable, Sendable {
+    /// The column name.
+    public let name: String
+    
+    /// If the field can be identified as a column of a specific table, the object ID of the table; otherwise zero.
+    public let tableOID: Int32
+    
+    /// If the field can be identified as a column of a specific table, the attribute number of the column; otherwise zero.
+    public let columnAttributeNumber: Int16
+    
+    /// The object ID of the field's data type.
+    public let dataType: PostgresDataType
+    
+    /// The data type size (see pg_type.typlen). Note that negative values denote variable-width types.
+    public let dataTypeSize: Int16
+    
+    /// The type modifier (see pg_attribute.atttypmod). The meaning of the modifier is type-specific.
+    public let dataTypeModifier: Int32
+    
+    /// The format being used for the field. Currently will be text or binary. In a RowDescription returned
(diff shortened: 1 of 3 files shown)


Sources/PostgresNIO/New/PostgresColumn.swift
```diff
@@ -0,0 +1,46 @@
...
+    
+    /// The data type size (see pg_type.typlen). Note that negative values denote variable-width types.
+    public let dataTypeSize: Int16
+    
+    /// The type modifier (see pg_attribute.atttypmod). The meaning of the modifier is type-specific.
+    public let dataTypeModifier: Int32
+    
+    /// The format being used for the field. Currently will be text or binary. In a RowDescription returned
+    /// from the statement variant of Describe, the format code is not yet known and will always be text.
+    public let format: PostgresFormat
```
fabianfett, line 26: can we back this with a `RowDescription.Column` please. Instead of let properties please lets use computed properties.

Example:

```swift
public var name: String {
    self.underlying.name
}
```

Sources/PostgresNIO/New/PostgresRowSequence.swift
```diff
@@ -26,6 +26,21 @@ public struct PostgresRowSequence: AsyncSequence, Sendable {
...
+                name: column.name,
+                tableOID: column.tableOID,
+                columnAttributeNumber: column.columnAttributeNumber,
+                dataType: column.dataType,
+                dataTypeSize: column.dataTypeSize,
+                dataTypeModifier: column.dataTypeModifier,
+                format: column.format
+            )
+        }
+    }
```
fabianfett, line 43: This is an O(n) call. Where `n` equals the number of columns. Do we need a returned Array? Can we return a custom PostgresColumns struct that wraps the existing `[RowDescription.Column]` structure?

Example:
```
struct PostgresColumns: Sequence {
    typealias Element == PostgresColumn
   
    var underlying: [RowDescription.Column]

    func makeIterator() -> Iterator {
        Iterator(underlying: self.underlying.makeIterator)
    }

    struct Iterator: IteratorProtocol {
        var underlying: [RowDescription.Column].Iterator

        func next() -> PostgresColumn {
           self.underlying.next()
        }
    }
}
```

fabianfett: If we do the above we can just add a get only property:

```swift
var columns: PostgresColumns {
    PostgresColumns(underlying: self.columns)
}
```

fabianfett:
Also thanks for pushing on this @SteveLauC. My apologies for not coming back to you sooner.

fabianfett:
@SteveLauC Do you want to take this over the finish line?

PR #579, opened 2025-08-26
Fix several CI issues

gwynne:
- Update the list of versions for Swift images
- Update actions/checkout to v5
- Disable TSan when building with Swift 5.10 and 6.0 due to swift-crypto hitting a compiler crasher
- Add macOS Sequoia tests
- Remove Homebrew bug workaround (no longer needed)


fabianfett (approved):
Love it!

PR #582, opened 2025-09-06
Fix json platform version warning

zunda-pixel:
Build Warning Log

```
Conformance of 'JSONEncoder' to 'Sendable' is only available in macOS 13.0 or newer
```

Screenshot 2025-09-06 at 14 35 09


gwynne:
We can't do this. Changing platform versions in `Package.swift` is a semver-major break.

PR #584, opened 2025-09-06
fix ExistentialAny(any/ some) warning

zunda-pixel:
fix ExistentialAny(any/ some) warning

Screenshot 2025-09-06 at 14 38 16


gwynne:
There's no need to enable this setting in PostgresNIO, and using `some` in the interface would be a public API break.

fabianfett:
Hi @zunda-pixel, I'd be happy to take this PR if it used `any` everywhere. Using some is, as @gwynne mentioned, an API break that we would not want right now.

PR #600, opened 2025-11-06
Bugfix: Closing PostgreSQL connection deadlocks listen(_:)

billyean:
Received the following diagnostics while designing an application with `postgres-nio` 1.27.0...1.29.0:

```
00:56:35.221 [NotificationsListener.swift:517] - Notifications: Failed to LISTEN {error: "PSQLError(code: server, serverInfo: [sqlState: 57P01, file: be-secure.c, line: 209, message: terminating connection due to unexpected postmaster exit, routine: secure_read, localizedSeverity: FATAL, severity: FATAL])", topic: hawq_39}
00:56:35.222 [PostgresChannelHandler.swift:104] - Channel error caught. {psql_error: "I/O on closed channel", psql_connection_id: "1"}
00:56:35.222 [NotificationsListener.swift:286] - Notifications: Restarting subscription {topic: hawq_205}
00:56:35.222 [NotificationsListener.swift:148] - Notifications: Starting listening {psql.conn.id: "0", psql.conn.isClosed: true, topic: hawq_157}

SWIFT TASK CONTINUATION MISUSE: listen(_:) leaked its continuation without resuming it. This may cause tasks waiting on it to remain suspended forever.
```

And then the application deadlocks because `listen()` doesn't ever return.

Reproduction:

1. The code looks like that:

                try await self.postgresClient.withConnection { connection in
                    148: log.debug("Notifications: Starting listening", metadata: [       
                        "topic": "\(topic)",                                         
                        "psql.conn.id": "\(connection.id)",                          
                        "psql.conn.isClosed": "\(connection.isClosed)"               
                    ])                                                               
                    sequence = try await connection.listen(topic.stringValue)        
                    154: log.debug("Notifications: Now listening", metadata: [            
                        "topic": "\(topic)",                                          
                        "psql.conn.id": "\(connection.id)",                          
                        "psql.conn.isClosed": "\(connection.isClosed)"               
                    ])

2. brew services restart postgresql
3. The error log similar to the log above gets generated, and the system goes into a deadlock.

Notice the log from line 148 is present, the connection appears to be CLOSED, and then the line 154 is unreachable. This might mean that `listen()` state machine is not correct on closed channels and result in deadlock.

### DIFF
--- a/Sources/ConnectionPoolModule/PoolStateMachine.swift
+++ b/Sources/ConnectionPoolModule/PoolStateMachine.swift
@@ -684,7 +684,9 @@ struct PoolStateMachine<
         if !requests.isEmpty {
             let leaseResult = self.connections.leaseConnection(at: index, streams: UInt16(requests.count))
             let connectionsRequired: Int
-            if self.requestQueue.count <= self.connections.stats.availableStreams + self.connections.stats.leasedStreams {
+            // if request count is less than available streams and leased streams plus incoming connections then only 
+            // ensure we have minimum connections otherwise grow the number of connections
+            if (self.requestQueue.count + 1) <= self.connections.stats.availableStreams + self.connections.stats.leasedStreams + self.connections.stats.connecting {
                 connectionsRequired = self.configuration.minimumConnectionCount - Int(self.connections.stats.active)
             } else {
                 connectionsRequired = 1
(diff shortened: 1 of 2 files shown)


fabianfett (approved):
Thanks @billyean, that's a great first PR.

PR #605, opened 2025-12-04
Throw error when `PostgresCopyFromWriter` is escaped

ahoppen:
When `PostgresCopyFromWriter` is escaped from the `copyFrom` closure (which users really shouldn’t do) and the user tries to write data to it after `copyFrom` has finished, throw an error instead of hitting a `preconditionFailure`.


gwynne (changes requested):
I would actually rather this throw a `preconditionFailure()`—albeit a more appropriate one—as escaping a value from a closure it's not supposed to escape from is unambiguously programmer error.

FranzBusch (commented):
I tend to agree. While I try to avoid preconditions as much as possible this seems like a genuine programmer error. @ahoppen what is your motivation here?

ahoppen:
I remember discussing this with @fabianfett at the time but don’t remember the rationale (it’s been a while). I’m equally happy with a precondition failure.

ptoffy:
Closing this then!

PR #606, opened 2025-12-04
Rethrow error thrown from `withAsyncTestingChannel` closure

ahoppen:
Currently, when the closure of `withAsyncTestingChannel` throws, the error is just swallowed silently and the test passes. Actually rethrow it so the test fails

### DIFF
--- a/Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
+++ b/Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
@@ -964,7 +964,8 @@ import Synchronization
         do {
             try await body(connection, channel)
         } catch {
-
+            try await connection.close()
+            throw error
         }
 
         try await connection.close()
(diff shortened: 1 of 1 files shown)


ptoffy (approved):

Tests/PostgresNIOTests/New/PostgresConnectionTests.swift
```diff
@@ -1098,7 +1098,8 @@ import Synchronization
         do {
             try await body(connection, channel)
         } catch {
-
+            try await connection.close()
```
gwynne:
```suggestion
            try? await connection.close()
```
In case closing the connection errors, this makes sure the original error gets rethrown instead of the connection close error.

gwynne (changes requested):

gwynne (approved):

PR #634, opened 2026-03-16
Structural updates

gwynne:
Enables Dependabot, adds more CI, and cleans up a couple things.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,18 +4,14 @@ concurrency:
   cancel-in-progress: true
 on: 
   push:
-    branches: 
-      - "main"
+    branches: [ "main" ]
   pull_request:
-    branches: 
-      - "*"
-env:
-  LOG_LEVEL: info
-permissions:
-  contents: read
+    branches: [ "*" ]
 
 jobs:
   linux-unit:
+    permissions:
+      contents: read
     strategy:
       fail-fast: false
       matrix:
@@ -27,15 +23,8 @@ jobs:
     container: ${{ matrix.swift-image }}
     runs-on: ubuntu-latest
     steps:
-      - name: Display OS and Swift versions
-        shell: bash
-        run: |
-          [[ -z "${SWIFT_PLATFORM}" ]] && SWIFT_PLATFORM="$(. /etc/os-release && echo "${ID}${VERSION_ID}")"
-          [[ -z "${SWIFT_VERSION}" ]] && SWIFT_VERSION="$(cat /.swift_tag 2>/dev/null || true)"
-          printf 'OS:  %s\nTag: %s\nVersion:\n' "${SWIFT_PLATFORM}-${RUNNER_ARCH}" "${SWIFT_VERSION}"
-          swift --version
       - name: Install curl for Codecov
-        run: apt-get update -y -q && apt-get install -y curl
+        run: apt-get update -yq && apt-get install -y curl
       - name: Check out package
         uses: actions/checkout@v6
       - name: Run unit tests with Thread Sanitizer
@@ -47,6 +36,8 @@ jobs:
           codecov_token: ${{ secrets.CODECOV_TOKEN }}
 
   linux-integration-and-dependencies:
+    permissions:
+      contents: read
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


Sources/PostgresNIO/New/PSQLRowStream.swift
```diff
@@ -287,7 +287,7 @@ final class PSQLRowStream: @unchecked Sendable {
         precondition(!newRows.isEmpty, "Expected to get rows!")
         self.eventLoop.preconditionInEventLoop()
         self.logger.trace("Row stream received rows", metadata: [
-            "row_count": "\(newRows.count)"
```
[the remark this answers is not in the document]
fabianfett: that is super, duper, wooper expensive though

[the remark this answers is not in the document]
fabianfett: please revert this part of the change because of this please.

.github/workflows/test.yml
```diff
@@ -159,16 +143,80 @@ jobs:
...
+      - name: Install SDK
+        run: |
+          tag="swift-$(swiftc --version | grep -oE '[0-9]+\.[0-9]+(\.[0-9]+)?' | head -1)-RELEASE"
+          checksum="$(curl -s https://www.swift.org/api/v1/install/releases.json | \
+            jq -r --arg tag "$tag" '.[] | select(.tag == $tag).platforms[] | select(.platform == "wasm-sdk").checksum // empty')"
+          swift sdk install https://download.swift.org/$(echo $tag | tr [A-Z] [a-z])/wasm-sdk/$tag/"$tag"_wasm.artifactbundle.tar.gz --checksum $checksum
+      - name: Build
+        run: |
+          tag="$(swiftc --version | grep -Po '(?<=\().+(?=\))')"
+          swift build --swift-sdk "$tag"_wasm
```
fabianfett: lets remove this for now, if we dont use it.

Sources/PostgresNIO/New/PSQLRowStream.swift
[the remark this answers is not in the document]
gwynne: Done

.github/workflows/test.yml
gwynne: Yanked

.github/workflows/test.yml
```diff
@@ -159,16 +143,56 @@ jobs:
...
+          checksum="$(echo "$sdk_info" | cut -f2)"
+          swift sdk install "https://download.swift.org/$(echo $tag | tr [A-Z] [a-z])/static-sdk/$tag/${tag}_static-linux-${version}.artifactbundle.tar.gz" --checksum "$checksum"
+      - name: Build
+        run: swift build --swift-sdk x86_64-swift-linux-musl
+
+  submit-dependencies:
+    permissions:
+      contents: write
+    if: github.event_name == 'push'
+    uses: vapor/ci/.github/workflows/submit-deps.yml@main
```
fabianfett, line 197: what does this do?

gwynne: It implements https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/using-the-dependency-submission-api

fabianfett: cool. but that's only important for leaf packages, isn't it? I assume this works on the Package.resolved, and this can look very different based on the checkout.

gwynne: No, it matters for libraries too, it helps Dependabot to determine whether you're affected by security advisories in dependencies. Yes, it's dependent on the most recent versions released at the time the submission job runs, since there's no lockfile in the repo (and yeah there can be platform-specific dependency issues too), but it's still better than nothing at all for the notices about security issues so you can update the minimum requirements in `Package.swift`. (Also, I believe the tool I'm currently using for dependency detection yields all deps, not just "currently active" ones—it's an SBOM generator, after all. I'll be switching to using the one that's now built into SwiftPM when it's available, which will actually be kind've annoying because that doesn't have a native "Github Dependency Submission API JSON" output mode like the third-party tooling does. 😆)

fabianfett: > Yes, it's dependent on the most recent versions released at the time the submission job runs, since there's no lockfile in the repo (and yeah there can be platform-specific dependency issues too), but it's still better than nothing at all for the notices about security issues so you can update the minimum requirements in `Package.swift`

But this means, that for tests we will always pick the latest version, and if there is a security advisory, by definition we will pick the version that has the issue already resolved.

gwynne: For tests, yes. But if you don't update the minimums in `Package.swift`, downstream users can end up with transitive dependencies that don't get updated because of some other requirement in the graph or just because SwiftPM hiccups (I've seen both scenarios in the real world multiple times).

fabianfett: but that would never be caught here, because `Package.resolved` is always using the latest and greatest.

PR #638, opened 2026-03-30
ConnectionPool should support draining connections

fabianfett:


Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -383,6 +383,47 @@ extension PoolStateMachine {
...
 
+        @usableFromInline
+        enum ConnectionWillCloseAction {
+            case closeConnection(CloseAction)
+            case none
+        }
+
+        @inlinable
+        mutating func connectionWillClose(_ connectionID: Connection.ID) -> ConnectionWillCloseAction {
+            guard let index = self.connections.firstIndex(where: { $0.id == connectionID }) else {
```
gwynne, line 394: General question: Why is `self.connections` an Array rather than a Dictionary or OrderedDictionary? I would assume it has something to do with speed on hotter paths than this particular line of code; I just cringe when I see "oh look, `O(n)` iteration of an ordered collection looking for an easily-hashed identifier".

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionState.swift
```diff
@@ -419,30 +433,54 @@ extension PoolStateMachine {
...
-            case .closing:
-                return nil
+            case .draining(let connection, let usedStreams):
+                precondition(usedStreams >= returnedStreams)
+                let newUsedStreams = usedStreams - returnedStreams
+                if newUsedStreams == 0 {
+                    self.state = .closing(connection)
+                    return .drainingComplete(connection)
+                } else {
+                    self.state = .draining(connection, usedStreams: newUsedStreams)
```
gwynne, line 477:
```suggestion
                if usedStreams == returnedStreams {
                    self.state = .closing(connection)
                    return .drainingComplete(connection)
                } else {
                    self.state = .draining(connection, usedStreams: usedStreams - returnedStreams)
```

gwynne (approved):
Couple of minor remarks, nothing blocking.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
fabianfett: yeah, actually O(n) is smaller than calculating O(1) hashes :)

gwynne: I'll take your word for it. 😆

PR #641, opened 2026-04-21
ConnectionPool: fix continuation leak when idle-timer reschedules

ltetzlaff:
### Motivation

Running a client built on `_ConnectionPoolModule` locally produces:

```
SWIFT TASK CONTINUATION MISUSE: runTimer(_:in:) leaked its continuation without resuming it.
```

### Cause

When the idle-timeout timer fires while a lease request is queued (because a keep-alive is currently consuming the connection's only stream), `connectionIdleTimerTriggered` takes the `rescheduleIdleTimer` path. That path replaced the stale idle timer with a fresh one but dropped the old timer's `cancellationContinuation` on the floor.

The dropped token is a `CheckedContinuation<Void, Never>` that `ConnectionPool.runTimer`'s child task registered via `withCheckedContinuation`. Because the state machine holds it strongly through `State.Timer`, it is only deallocated when the state transition discards the timer — at which point it has never been resumed, and the runtime logs the warning above.

### Modifications

- `ConnectionState.rescheduleIdleTimer()` and `ConnectionGroup.rescheduleIdleTimer(_:)` now also return the old idle timer's `TimerCancellationToken`.
- `PoolStateMachine.connectionIdleTimerTriggered(_:)` forwards that token via `.makeConnectionsCancelAndScheduleTimers` so `ConnectionPool.runConnectionAction` resumes the continuation before scheduling the new timer.

### Result

New regression test `testIdleTimerReschedulePreservesOldCancellationToken` in `PoolStateMachineTests` drives the state machine into the exact race (lease → establish → release → keep-alive fires → queue second request → idle timer fires while keep-alive is running) and asserts the returned action surfaces the old cancellation token. The test fails on `main` and passes with this fix.

Full existing `ConnectionPoolModuleTests` suite (106 tests) continues to pass.

---

Context: I originally opened this against `valkey-io/valkey-swift` (valkey-io/valkey-swift#377), and @adam-fowler asked me to port it here first since that repo sources its pool code from this one.

### DIFF
--- a/Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
+++ b/Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
@@ -698,7 +698,7 @@ extension PoolStateMachine {
         }
 
         @inlinable
-        mutating func rescheduleIdleTimer(_ connectionID: Connection.ID) -> ConnectionTimer? {
+        mutating func rescheduleIdleTimer(_ connectionID: Connection.ID) -> (ConnectionTimer, TimerCancellationToken?)? {
             guard let index = self.connections.firstIndex(where: { $0.id == connectionID }) else {
                 // because of a race this connection (connection close runs against trigger of timeout)
                 // was already removed from the state machine.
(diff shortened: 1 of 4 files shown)


adam-fowler (commented):
I can't approve here, but this looks good. @fabianfett what do you think?

fabianfett (approved):

fabianfett (approved):
This is a great patch! Thanks!

PR #644, opened 2026-04-28
Add missing @available() attributes

adam-fowler:
Add additional `@available` attributes where needed. This doesn't affect Postgres-nio directly as it has platform requirements already, but when the connection pool code is copied to other projects (eg valkey-swift) these have to be added back in, and sometimes this gets forgotten.


fabianfett (approved):
Awesome! Thank you!

Issue #649, opened 2026-07-07
`precondition()` should be an error thrown


herzi:
### Describe the issue

`PostgresRow.decode(_:context:file:line)` triggers a precondition failure when the number of column types does not match the number of returned column

### Vapor version

vapor=n/a postgres-nio=1.33.0

### Operating system and version

macOS 26.5.1

### Swift version

swift-driver version: 1.148.6 Apple Swift version 6.3.3 (swiftlang-6.3.3.1.3 clang-2100.1.1.101)

### Steps to reproduce

Use a code snippet like this to iterate a mismatch query/decode:

```swift
let client: PostgresClient // populate however you prefer
try await client.withConnection { connection in
  // note how the query just returns one value
  for try await row in try await connection.query("SELECT now() as time", logger: logger) {
    // note how we try to decode more than one column
    try row.decode((Date, String).self) 
  }
}
```

Expected outcomes:
1. Code compiles
2. Code runs and throws an error

### Outcome

Observed outcomes:
1. Code compiles
2. Code crashes with `PostgresNIO/VariadicGenerics.swift:56: Precondition failed` (which also isn't a very good error message for this thing happening)

### Additional notes

While this is clearly a programming error (on my side), I think it would be more consistently handled with the rest of the framework's behavior.  A programming error is also an invalid SQL query, however that will throw `PSQLError`, so that other tasks in a Swift-based server can continue.

More importantly, in a multi-service environment, throwing an error will at least provide a way to shut down other services cleanly compared to a hard crash that might lose data in-flight.

*NOTE:* If maintainers agree that this is a useful improvement, I will happily work on a PR to implement this.

gwynne:
I _definitely_ concur that this is a useful improvement! This absolutely should not cause a hard crash. It is technically a programming error, but such a mismatch can also be caused by changes on the database side over which the server has no control, such as a stored procedure's output having a different format after a schema update.

herzi:
Thanks a lot for the feedback, @gwynne.  I raised PR #650 to address this and would love to get some feedback.

ptoffy:
Closed in #650

PR #650, opened 2026-07-08
Prefer Throwing Error over Crashing API Consumers

herzi:
This change updates the twin-implementations of `PostgresRow.decode(…)` to throw an error instead of crashing client apps when there are not enough columns for decoding (see #649 for the initial bug report).


herzi:
Options disregarded for this implementation:

* Extend `PostgresError`:  That's a public `enum`, adding a new case can break API consumers' `switch` statements – as such, I would argue this is not an API compatible change
* Reuse `PostgresDecodingError`:  That type is strongly designed to inform about failing to decode a specific column of a row.

As such, I assumed it best to replicate the structure of `PostgresDecodingError` into a new `PostgresRowDecodingError`.

Sources/PostgresNIO/New/PostgresRowDecodingError.swift
```diff
@@ -0,0 +1,71 @@
+public struct PostgresRowDecodingError: Error {
```
fabianfett, line 1: please extend PSQLError instead of adding a new error type.

Sources/PostgresNIO/New/VariadicGenerics.swift
```diff
@@ -17,7 +17,15 @@ extension PostgresRow {
         file: String = #fileID,
         line: Int = #line
     ) throws -> (Column) {
-        precondition(self.columns.count >= 1)
+        guard self.columns.count >= 1 else {
+            throw PostgresRowDecodingError(
+                code: .columnCountMismatch,
```
fabianfett, line 22: lets not call this mismatch. mismatch implies that this would throw if we get more columns then we need. `notEnoughColumns` is probably better.

Tests/PostgresNIOTests/New/PostgresRowTests.swift
```diff
@@ -3,6 +3,49 @@ import XCTest
 import NIOCore
 
 final class PostgresRowTests: XCTestCase {
+    func testColumnCountMismatchSingleType() {
```
fabianfett, line 6: lets use Swift testing as soon as this pr has landed: https://github.com/vapor/postgres-nio/pull/655

Sources/PostgresNIO/New/PostgresRowDecodingError.swift
herzi: Thanks for the pointer 🙏  I'll try to find some time to adjust the PR this weekend.

fabianfett:
@herzi is this something that you want to bring over the finish line? Otherwise @ptoffy could maybe pick this up?

herzi:
At the moment, I don't have a lot of time at my hands for looking at this.  Happy to have someone else pick this up

fabianfett (approved):
Awesome. LGTM

PR #651, opened 2026-07-24
Add graceful shutdown

### LINKED ISSUE #595: [Connection Pool]: Implement triggerGracefulShutdown
### Please summarize your feature request

Implement ConnectionPool.triggerGracefulShutdown which call graceful shutdown on all open connections and moves to shutdown state once all connections are closed.

### Describe the functionality you're seeking in more detail

Another request from valkey-swift. 

The cluster client in valkey-swift manages multiple connection pools and will bring up and shutdown pools as when nodes appear and disappear from the cluster. Currently the client has to run a force shutdown on the connection pool which closes all the connections regardless of whether requests are still in progress. Ideally when shutting down a connection pool we should be able to trigger a graceful shutdown which waits until current requests are served.

The connection manager state machine has a stub for triggerGracefulShutdown which fatalErrors. It would be good to implement it.

### Have you considered any alternatives?

_No response_

ptoffy:
Closes #595 

This implements graceful shutdown behaviour in the connection state machine. When it's triggered, in flight requests are completed and open connections are kept open until all requests in the request queue are completed. The PostgresClient now makes use of this new behaviour

### DIFF
--- a/Sources/ConnectionPoolModule/PoolStateMachine.swift
+++ b/Sources/ConnectionPoolModule/PoolStateMachine.swift
@@ -748,30 +748,36 @@ struct PoolStateMachine<
     @usableFromInline
     mutating func triggerForceShutdown() -> Action {
         switch self.poolState {
-        case .running, .connectionCreationFailing, .circuitBreakOpen:
-            self.poolState = .shuttingDown
-            var shutdown = ConnectionAction.Shutdown()
-            self.connections.triggerForceShutdown(&shutdown)
-
-            if self.connections.isEmpty, shutdown.connections.isEmpty {
-                self.poolState = .shutDown
-                return .init(
-                    request: .failRequests(self.requestQueue.removeAll(), ConnectionPoolError.poolShutdown),
-                    connection: .cancelEventStreamAndFinalCleanup(shutdown.timersToCancel)
-                )
-            }
-
-            return .init(
-                request: .failRequests(self.requestQueue.removeAll(), ConnectionPoolError.poolShutdown),
-                connection: .initiateShutdown(shutdown)
-            )
+        case .shutDown:
+            return .init(request: .none, connection: .none)
+        
+        case .shuttingDown where self.gracefulShutdownTriggered:
+            // Graceful shutdown is in progress, escalate to force shutdown.
+            self.gracefulShutdownTriggered = false
 
         case .shuttingDown:
             return .none()
 
-        case .shutDown:
-            return .init(request: .none, connection: .none)
+        case .running, .connectionCreationFailing, .circuitBreakOpen:
+            break
         }
+
+        self.poolState = .shuttingDown
+        var shutdown = ConnectionAction.Shutdown()
+        self.connections.triggerForceShutdown(&shutdown)
+
+        if self.connections.isEmpty, shutdown.connections.isEmpty {
+            self.poolState = .shutDown
+            return .init(
+                request: .failRequests(self.requestQueue.removeAll(), ConnectionPoolError.poolShutdown),
+                connection: .cancelEventStreamAndFinalCleanup(shutdown.timersToCancel)
+            )
+        }
+
+        return .init(
+            request: .failRequests(self.requestQueue.removeAll(), ConnectionPoolError.poolShutdown),
+            connection: .initiateShutdown(shutdown)
+        )
     }
 
     @inlinable
--- a/Sources/ConnectionPoolModule/PoolStateMachine.swift
+++ b/Sources/ConnectionPoolModule/PoolStateMachine.swift
@@ -698,8 +698,30 @@ struct PoolStateMachine<
         var requests: [Request]
     }
 
+    @usableFromInline
     mutating func triggerGracefulShutdown() -> Action {
-        fatalError("Unimplemented")
+        switch self.poolState {
+        case .running, .connectionCreationFailing, .circuitBreakOpen:
+            self.poolState = .shuttingDown
+            var shutdown = ConnectionAction.Shutdown()
+            self.connections.triggerGracefulShutdown(&shutdown)
+
+            if self.connections.isEmpty, shutdown.connections.isEmpty {
+                self.poolState = .shutDown
+                return .init(
+                    request: .failRequests(self.requestQueue.removeAll(), ConnectionPoolError.poolShutdown),
+                    connection: .cancelEventStreamAndFinalCleanup(shutdown.timersToCancel)
+                )
+            }
+
+            return .init(
+                request: .failRequests(self.requestQueue.removeAll(), ConnectionPoolError.poolShutdown),
+                connection: .initiateShutdown(shutdown)
+            )
+
+        case .shuttingDown, .shutDown:
+            return .none()
+        }
     }
 
     @usableFromInline
--- a/Sources/ConnectionPoolModule/PoolStateMachine.swift
+++ b/Sources/ConnectionPoolModule/PoolStateMachine.swift
@@ -495,6 +495,15 @@ struct PoolStateMachine<
             if self.connections.isEmpty {
                 self.poolState = .shutDown
                 connectionAction = .cancelEventStreamAndFinalCleanup(timerToCancel.map {[$0]} ?? [])
+
+                if !requestQueue.isEmpty {
+                    // If there's no connections left but there are requests left in the queue, fail them.
+                    // We don't want to open other connections while draining
+                    return .init(
+                        request: .failRequests(self.requestQueue.removeAll(), .poolShutdown),
+                        connection: connectionAction
+                    )
+                }
             } else {
                 connectionAction = .cancelTimers(timerToCancel.map {[$0]} ?? [])
             }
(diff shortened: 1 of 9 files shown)


Sources/PostgresNIO/Pool/PostgresClient.swift
```diff
@@ -468,6 +468,10 @@ public final class PostgresClient: Sendable, ServiceLifecycle.Service {
     /// Cancelling the task that executes the ``run()`` method is equivalent to closing the client. Once the task
     /// has been cancelled the client is not able to process any new queries or prepared statements.
     ///
+    /// Graceful shutdown is different to cancellation. If the client is gracefully shut down, in-flight requests
+    /// are completed and the requests queue is drained, but no new connection is opened. If the last connection
```
gwynne:
```suggestion
    /// are completed and the requests queue is drained, but no new connections are opened. If the last connection
```

Sources/PostgresNIO/Pool/PostgresClient.swift
```diff
@@ -468,6 +468,10 @@ public final class PostgresClient: Sendable, ServiceLifecycle.Service {
...
+    /// fails while there's still queued requests, those requests fail.
```
gwynne:
```suggestion
    /// fails while there are still queued requests, those requests fail.
```

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -495,6 +495,15 @@ struct PoolStateMachine<
             if self.connections.isEmpty {
                 self.poolState = .shutDown
                 connectionAction = .cancelEventStreamAndFinalCleanup(timerToCancel.map {[$0]} ?? [])
+
+                if !requestQueue.isEmpty {
+                    // If there's no connections left but there are requests left in the queue, fail them.
+                    // We don't want to open other connections while draining
```
gwynne:
```suggestion
                    // We don't want to open new connections while draining.
```

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -495,6 +495,15 @@ struct PoolStateMachine<
             if self.connections.isEmpty {
                 self.poolState = .shutDown
                 connectionAction = .cancelEventStreamAndFinalCleanup(timerToCancel.map {[$0]} ?? [])
+
+                if !requestQueue.isEmpty {
+                    // If there's no connections left but there are requests left in the queue, fail them.
```
gwynne:
```suggestion
                    // If there are no connections left but requests remain in the queue, fail them.
```

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -677,6 +686,14 @@ struct PoolStateMachine<
             if self.connections.isEmpty {
                 self.poolState = .shutDown
                 connectionAction = .cancelEventStreamAndFinalCleanup(.init(closedConnectionAction.timersToCancel))
+
+                if !requestQueue.isEmpty {
+                    // If all connections were closed (errored) but there are requests left in the queue, fail them
```
gwynne:
```suggestion
                    // If all connections were closed (errored) but there are requests left in the queue, fail them.
```

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -698,8 +715,34 @@ struct PoolStateMachine<
...
+
+            if self.connections.isEmpty, shutdown.connections.isEmpty {
+                self.poolState = .shutDown
+                return .init(
+                    request: .failRequests(self.requestQueue.removeAll(), ConnectionPoolError.poolShutdown),
+                    connection: .cancelEventStreamAndFinalCleanup(shutdown.timersToCancel)
+                )
+            }
+
+            // Don't shut down connections if queue still has requests. We have to drain it
```
gwynne:
```suggestion
            // Don't shut down connections if the queue still has requests. We have to drain it.
```

gwynne (commented):
Just some grammatical nitpickery.

gwynne (changes requested):
Can you add a test that triggers graceful shutdown externally (e.g. by causing graceful shutdown of a ServiceGroup containing the PostgresClient)?

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -698,8 +715,34 @@ struct PoolStateMachine<
...
+            if self.connections.isEmpty, shutdown.connections.isEmpty {
+                self.poolState = .shutDown
+                return .init(
+                    request: .failRequests(self.requestQueue.removeAll(), ConnectionPoolError.poolShutdown),
+                    connection: .cancelEventStreamAndFinalCleanup(shutdown.timersToCancel)
+                )
+            }
+
+            // Don't shut down connections if the queue still has requests. We have to drain it.
+            if shutdown.connections.isEmpty && shutdown.timersToCancel.isEmpty {
```
adam-fowler: Where's the check for the request queue not being empty?

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -741,6 +784,13 @@ struct PoolStateMachine<
         if !requests.isEmpty {
             let leaseResult = self.connections.leaseConnection(at: index, streams: UInt16(requests.count))
             let connectionsRequired: Int
+
+            if gracefulShutdownTriggered {
```
adam-fowler, line 802: Should `gracefulShutdownTriggered` be an associated value of `State.shuttingDown`

adam-fowler (commented):
A couple of other things outside of the inline comments
- What happens if someone calls `triggerForceShutdown` after `triggerGracefulShutdown`. It should probably do a force shutdown. 
- I'm not certain about draining the request queue. If you are in the situation where you have just come out of the `connectionCreationFailing` and have a load of requests waiting to be served. You will probably have only one or two connections which have to complete all the requests which could take some time.

Sources/ConnectionPoolModule/PoolStateMachine.swift
ptoffy: The comment is outdated and should be a few lines up. The way it works now is if the connections are empty we return `.failRequests(self.requestQueue.removeAll(), ConnectionPoolError.poolShutdown)`, which just returns `.failRequests([])` if there's no more requests (and no-ops). It's the same thing that `triggerForceShutdown` is doing. Maybe it would be cleaner to have a special case returning `.none` for it though

Sources/ConnectionPoolModule/PoolStateMachine.swift
ptoffy: Hmm I just used it as it was written by Fabian

ptoffy:
> What happens if someone calls triggerForceShutdown after triggerGracefulShutdown. It should probably do a force shutdown.

Good shout, we now explicitly check for that and I added a couple of tests

gwynne (approved):

FranzBusch (commented):
This looks good to me on the surface and the tests make sense as well. Thanks!

0xTim (approved):
LGTM

Sources/ConnectionPoolModule/PoolStateMachine.swift
adam-fowler: You can remove this check completely and follow through and return the `initiateShutdown` with empty arrays. I think with it in, it just adds confusion (especially with the current comment)

adam-fowler:
I'm happy with this

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -698,37 +715,77 @@ struct PoolStateMachine<
...
-        fatalError("Unimplemented")
-    }
-
     @usableFromInline
-    mutating func triggerForceShutdown() -> Action {
+    mutating func triggerGracefulShutdown() -> Action {
         switch self.poolState {
         case .running, .connectionCreationFailing, .circuitBreakOpen:
             self.poolState = .shuttingDown
+            self.gracefulShutdownTriggered = true
```
fabianfett, line 723: this is wrong. even when we are in graceful shutdown, we want to allow opening new connections. `gracefulShutdownTriggered` shall only disallow accepting new requests. all old requests must be served. we need a follow up here.

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
```diff
@@ -1030,6 +1030,104 @@ import Testing
...
+            }
+            await factory.nextConnectAttempt { connectionID in
+                return 1
+            }
+            let lease = try await pool.leaseConnection()
+            let leasedConnection = lease.connection
+
+            pool.triggerGracefulShutdown()
+            #expect(leasedConnection.isRunning)
+
```
fabianfett, line 1070: we need to test that additional requests are rejected. but we should also test that requests that are already in, can still get through. even if the connection fails in between.

Tests/PostgresNIOTests/New/PostgresClientTests.swift
```diff
@@ -38,4 +39,44 @@ import Testing
             // If we reach here the test passed (the .timeLimit above enforces the deadline).
         }
     }
+
+    @Test(.timeLimit(.minutes(1)))
+    func gracefulShutdownWaitsForInFlightRequest() async throws {
```
fabianfett, line 44: this test should never finish. we only fail requests, if the user explicitly force closes.

PR #652, opened 2026-07-27
Make `PostgresDataType` PostgresCodable

ptoffy:
This allows usage of `PostgresDataType` inside queries so dynamic OIDs can be queried at runtime and decoded into this

### DIFF
--- a/Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
+++ b/Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
@@ -9,28 +9,28 @@ extension PostgresDataType: PostgresDecodable {
         context: PostgresDecodingContext<JSONDecoder>
     ) throws where JSONDecoder: PostgresJSONDecoder {
         switch (format, type) {
-        case (.binary, .oid), (.binary, .regproc):
-            guard 
+        case (.binary, .oid), (.binary, .regproc), (.binary, .regclass), (.binary, .regtype):
+            guard
                 byteBuffer.readableBytes == 4,
                 let value = byteBuffer.readInteger(as: UInt32.self)
-            else { 
-                throw PostgresDecodingError.Code.failure 
+            else {
+                throw PostgresDecodingError.Code.failure
             }
             self.rawValue = value
         case (.binary, .int4):
-            guard 
+            guard
                 byteBuffer.readableBytes == 4,
                 let value = byteBuffer.readInteger(as: Int32.self).flatMap(UInt32.init(exactly:))
-            else { 
-                throw PostgresDecodingError.Code.failure 
+            else {
+                throw PostgresDecodingError.Code.failure
             }
             self.rawValue = value
         case (.text, .oid), (.text, .int4):
-            guard 
+            guard
                 let string = byteBuffer.readString(length: byteBuffer.readableBytes),
                 let value = UInt32(string)
             else {
-                throw PostgresDecodingError.Code.failure 
+                throw PostgresDecodingError.Code.failure
             }
             self.rawValue = value
         default:
@@ -40,12 +40,12 @@ extension PostgresDataType: PostgresDecodable {
 }
 
 extension PostgresDataType: PostgresNonThrowingEncodable {
-    public static var psqlType: PostgresDataType { 
-        .oid 
+    public static var psqlType: PostgresDataType {
+        .oid
     }
 
-    public static var psqlFormat: PostgresFormat { 
-        .binary 
+    public static var psqlFormat: PostgresFormat {
+        .binary
     }
 
     @inlinable
@@ -53,6 +53,6 @@ extension PostgresDataType: PostgresNonThrowingEncodable {
         into byteBuffer: inout ByteBuffer,
         context: PostgresEncodingContext<JSONEncoder>
     ) {
-        byteBuffer.writeInteger(self.rawValue)
+        byteBuffer.writeInteger(self.rawValue, as: UInt32.self)
     }
 }
--- a/Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
+++ b/Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
@@ -25,6 +25,8 @@ extension PostgresDataType: PostgresDecodable {
                 throw PostgresDecodingError.Code.failure
             }
             self.rawValue = value
+        // The `oid` aliases (`regproc`, `regclass` etc) only carry an id in binary form.
+        // In text form they're a string containing a name and so cannot be decoded into a PostgresDataType.
         case (.text, .oid), (.text, .int4):
             guard
                 let string = byteBuffer.readString(length: byteBuffer.readableBytes),
--- a/Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
+++ b/Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
@@ -1,0 +1,58 @@
+import NIOCore
+
+extension PostgresDataType: PostgresDecodable {
+    @inlinable
+    public init<JSONDecoder>(
+        from byteBuffer: inout ByteBuffer,
+        type: PostgresDataType,
+        format: PostgresFormat,
+        context: PostgresDecodingContext<JSONDecoder>
+    ) throws where JSONDecoder: PostgresJSONDecoder {
+        switch (format, type) {
+        case (.binary, .oid), (.binary, .regproc):
+            guard 
+                byteBuffer.readableBytes == 4,
+                let value = byteBuffer.readInteger(as: UInt32.self)
+            else { 
+                throw PostgresDecodingError.Code.failure 
+            }
+            self.rawValue = value
+        case (.binary, .int4):
+            guard 
+                byteBuffer.readableBytes == 4,
+                let value = byteBuffer.readInteger(as: Int32.self).flatMap(UInt32.init(exactly:))
+            else { 
+                throw PostgresDecodingError.Code.failure 
+            }
+            self.rawValue = value
+        case (.text, .oid), (.text, .int4):
+            guard 
+                let string = byteBuffer.readString(length: byteBuffer.readableBytes),
+                let value = UInt32(string)
+            else {
+                throw PostgresDecodingError.Code.failure 
+            }
+            self.rawValue = value
+        default:
+            throw PostgresDecodingError.Code.typeMismatch
+        }
+    }
+}
+
+extension PostgresDataType: PostgresNonThrowingEncodable {
+    public static var psqlType: PostgresDataType { 
+        .oid 
+    }
+
+    public static var psqlFormat: PostgresFormat { 
+        .binary 
+    }
+
+    @inlinable
+    public func encode<JSONEncoder>(
+        into byteBuffer: inout ByteBuffer,
+        context: PostgresEncodingContext<JSONEncoder>
+    ) {
+        byteBuffer.writeInteger(self.rawValue)
+    }
+}
--- a/Tests/IntegrationTests/PostgresDataTypeTests.swift
+++ b/Tests/IntegrationTests/PostgresDataTypeTests.swift
@@ -41,7 +41,26 @@ final class PostgresDataTypeIntegrationTests: XCTestCase {
         }
     }
 
-    func testOIDArray() async throws {
+    func testOIDAboveInt32MaxRoundTrips() async throws {
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer { XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully()) }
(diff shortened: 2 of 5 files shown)


Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
```diff
@@ -0,0 +1,60 @@
...
+        // The `oid` aliases (`regproc`, `regclass` etc) only carry an id in binary form.
+        // In text form they're a string containing a name and so cannot be decoded into a PostgresDataType.
+        case (.text, .oid), (.text, .int4):
+            guard
+                let string = byteBuffer.readString(length: byteBuffer.readableBytes),
+                let value = UInt32(string)
+            else {
+                throw PostgresDecodingError.Code.failure
+            }
+            self.rawValue = value
```
gwynne, line 37:
```suggestion
        // The `oid` aliases (`regproc`, `regclass` etc) only carry an id in binary form.
        // In text form they're a string containing a name and so cannot be decoded into a PostgresDataType.
        case (.text, .oid), (.text, .int4):
            guard
                let string = byteBuffer.readString(length: byteBuffer.readableBytes),
            else {
                throw PostgresDecodingError.Code.failure
            }
            if let value = UInt32(string) {
                self.rawValue = value
            } else if let value = Self.dataTypeForKnownSQLName(string.uppercased()) {
                self = value
            } else {
                throw PostgresDecodingError.Code.failure
            }
```
Sure they can, at least some of the time. You'll also of course need to add:
```swift
extension PostgresDataType {
    static func dataTypeForKnownSQLName(_ name: String) -> Self? {
        switch name {
        case "NULL": .null
        case "BOOLEAN": .bool
        case "BYTEA": .bytea
        // ... and so on
        // Use the list from `knownSQLName`.
        // It can be easily swapped around with a find/replace pattern such as
        // `case \.(.+?): return "(.+?)"` -> `case "$2": .$1`
        default: nil
        }
    }
}
```
I'm willing to admit this could be called a bridge too far, though.

Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
```diff
@@ -0,0 +1,60 @@
+import NIOCore
+
+extension PostgresDataType: PostgresDecodable {
+    @inlinable
+    public init<JSONDecoder>(
+        from byteBuffer: inout ByteBuffer,
+        type: PostgresDataType,
+        format: PostgresFormat,
+        context: PostgresDecodingContext<JSONDecoder>
+    ) throws where JSONDecoder: PostgresJSONDecoder {
```
gwynne, line 10:
```suggestion
    public init(
        from byteBuffer: inout ByteBuffer,
        type: PostgresDataType,
        format: PostgresFormat,
        context: PostgresDecodingContext<some PostgresJSONDecoder>
    ) throws {
```

Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
```diff
@@ -0,0 +1,60 @@
...
+        switch (format, type) {
+        case (.binary, .oid), (.binary, .regproc), (.binary, .regclass), (.binary, .regtype):
+            guard
+                byteBuffer.readableBytes == 4,
+                let value = byteBuffer.readInteger(as: UInt32.self)
+            else {
+                throw PostgresDecodingError.Code.failure
```
gwynne, line 17:
```suggestion
                throw PostgresDecodingError.Code.missingData
```

Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
```diff
@@ -0,0 +1,60 @@
...
+            }
+            self.rawValue = value
+        case (.binary, .int4):
+            guard
+                byteBuffer.readableBytes == 4,
+                let value = byteBuffer.readInteger(as: Int32.self).flatMap(UInt32.init(exactly:))
+            else {
+                throw PostgresDecodingError.Code.failure
```
gwynne, line 25:
```suggestion
                throw PostgresDecodingError.Code.missingData
```

Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
```diff
@@ -0,0 +1,60 @@
...
+            }
+            self.rawValue = value
+        // The `oid` aliases (`regproc`, `regclass` etc) only carry an id in binary form.
+        // In text form they're a string containing a name and so cannot be decoded into a PostgresDataType.
+        case (.text, .oid), (.text, .int4):
+            guard
+                let string = byteBuffer.readString(length: byteBuffer.readableBytes),
+                let value = UInt32(string)
+            else {
+                throw PostgresDecodingError.Code.failure
```
gwynne, line 35:
```suggestion
                throw PostgresDecodingError.Code.missingData
```

Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    public static var psqlFormat: PostgresFormat {
+        .binary
+    }
+
+    @inlinable
+    public func encode<JSONEncoder>(
+        into byteBuffer: inout ByteBuffer,
+        context: PostgresEncodingContext<JSONEncoder>
+    ) {
```
gwynne, line 57:
```suggestion
    public func encode(
        into byteBuffer: inout ByteBuffer,
        context: PostgresEncodingContext<some PostgresJSONEncoder>
    ) {
```

Tests/PostgresNIOTests/New/Data/PostgresDataType+PSQLCodableTests.swift
```diff
@@ -0,0 +1,122 @@
...
+            #expect(decoded == oid)
+        }
+    }
+
+    @Test func testDecodeFromBinaryOID() {
+        let values: [UInt32] = [0, 1, 26, UInt32(UInt16.max), UInt32(Int32.max), UInt32.max]
+        for value in values {
+            for type in [PostgresDataType.oid, .regproc] {
+                var buffer = ByteBuffer()
+                buffer.writeInteger(value)
```
gwynne:
```suggestion
                var buffer = ByteBuffer(integer: value)
```

Tests/PostgresNIOTests/New/Data/PostgresDataType+PSQLCodableTests.swift
```diff
@@ -0,0 +1,122 @@
...
+                }
+                #expect(decoded?.rawValue == value)
+            }
+        }
+    }
+
+    @Test func testDecodeFromBinaryInt4() {
+        for value in [Int32(0), 1, 26, Int32(UInt16.max), Int32.max] {
+            var buffer = ByteBuffer()
+            buffer.writeInteger(value)
```
gwynne:
```suggestion
            var buffer = ByteBuffer(integer: value)
```

Tests/PostgresNIOTests/New/Data/PostgresDataType+PSQLCodableTests.swift
```diff
@@ -0,0 +1,122 @@
...
+            var decoded: PostgresDataType?
+            #expect(throws: Never.self) {
+                decoded = try PostgresDataType(from: &buffer, type: .int4, format: .binary, context: .default)
+            }
+            #expect(decoded?.rawValue == UInt32(value))
+        }
+
+        for value in [Int32(-1), -26, Int32.min] {
+            var buffer = ByteBuffer()
+            buffer.writeInteger(value)
```
gwynne:
```suggestion
            var buffer = ByteBuffer()
            buffer.writeInteger(integer: value)
```

Tests/PostgresNIOTests/New/Data/PostgresDataType+PSQLCodableTests.swift
```diff
@@ -0,0 +1,122 @@
...
+                try PostgresDataType(from: &buffer, type: .int4, format: .binary, context: .default)
+            }
+        }
+    }
+
+    @Test func testDecodeFromText() {
+        for value in [0, 1, 26, UInt32(UInt16.max), UInt32.max] {
+            for type in [PostgresDataType.oid, .int4] {
+                var buffer = ByteBuffer()
+                buffer.writeString(String(value))
```
gwynne:
```suggestion
                var buffer = ByteBuffer(string: String(value))
```

Tests/PostgresNIOTests/New/Data/PostgresDataType+PSQLCodableTests.swift
```diff
@@ -0,0 +1,122 @@
...
+                    decoded = try PostgresDataType(from: &buffer, type: type, format: .text, context: .default)
+                }
+                #expect(decoded?.rawValue == value)
+            }
+        }
+    }
+
+    @Test func testDecodeFailureFromInvalidByteCount() {
+        var buffer = ByteBuffer()
+        buffer.writeInteger(UInt32(26))
```
gwynne:
```suggestion
        var buffer = ByteBuffer(integer: UInt32(26))
```

Tests/PostgresNIOTests/New/Data/PostgresDataType+PSQLCodableTests.swift
```diff
@@ -0,0 +1,122 @@
...
+
+        #expect(throws: PostgresDecodingError.Code.failure) {
+            try PostgresDataType(from: &buffer, type: .oid, format: .binary, context: .default)
+        }
+    }
+
+    @Test func testDecodeFailureFromInvalidText() {
+        for string in ["", "notanumber", "-1", "4294967296"] {
+            var buffer = ByteBuffer()
+            buffer.writeString(string)
```
gwynne:
```suggestion
            var buffer = ByteBuffer(string: string)
```

Tests/PostgresNIOTests/New/Data/PostgresDataType+PSQLCodableTests.swift
```diff
@@ -0,0 +1,122 @@
...
+
+            #expect(throws: PostgresDecodingError.Code.failure) {
+                try PostgresDataType(from: &buffer, type: .oid, format: .text, context: .default)
+            }
+        }
+    }
+
+    @Test func testDecodeFailureFromInvalidPostgresType() {
+        var buffer = ByteBuffer()
+        buffer.writeInteger(UInt32(26))
```
gwynne:
```suggestion
        let buffer = ByteBuffer(integer: UInt32(26))
```

gwynne (changes requested):

Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
ptoffy: That does seem a bit overkill

Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
ptoffy: Your suggestion breaks the convention that all the other `PostgresCodable` extensions use

Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
ptoffy: This also is not the convention. Other `PostgresCodable` conformances use `.failure`. Why would `.missingData` be more appropriate?

Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
gwynne: Just because it's the old convention that dates back to before opaque parameter types existed doesn't make it any less pointlessly verbose. 😛 But it's not a hill I care enough about to die on.

Sources/PostgresNIO/New/Data/PostgresDataType+PostgresCodable.swift
gwynne: I would argue that other conformances are wrong, because the only other thing `missingData` could mean is that there's _no_ data, and that has a well-defined non-error meaning of being `NULL`. But again, not a hill I care about dying on; I very much doubt anyone is relying on exact `PostgresDecodingError`s.

gwynne (approved):

Tests/IntegrationTests/PostgresDataTypeTests.swift
```diff
@@ -0,0 +1,104 @@
+import Logging
+import NIOCore
+import NIOPosix
+import PostgresNIO
+import XCTest
+
+final class PostgresDataTypeIntegrationTests: XCTestCase {
```
fabianfett, line 7: all new tests should be in swift-testing from the very beginning. @ptoffy

Issue #522, opened 2024-11-24
Support replication


pnewell:
### Please summarize your feature request

Using with Postgres logical replication to get a stream of changes to the database

### Describe the functionality you're seeking in more detail

I'd like to use postgres-nio to work with logical replication but it looks like for now messages related to it are marked precondition failures. It looks like a lot of the functionality in the New folder is moving towards making messages, etc internal, which may mean trying to extend things to intercept and process these messages myself won't be possible. Is there any recommended way I could do this while still using postgres-nio as a base? Or is the only option to fork/wait for it to be implemented given the direction the library is headed?

### Have you considered any alternatives?

I've started to work through forking but would rather avoid that if possible

fabianfett:
As replication likely requires different state machines, I'm pretty confident that forking is the best option right now. We might want to consider putting the messages into an extra target, that you could depend upon. Not sure if there is much more reusable besides the message definitions.

PR #654, opened 2026-07-28
feat(PostgresConnection): Add 'prepareChannel' for freeform channel usage

CrownedPhoenix:
This allows users to piggyback on PostgresConnection's channel setup/initialization so that they don't have to reimplement TCP/TLS/startup/authentication/etc in order to make a custom channel handler over a Postgres connection. 

At the time of this writing, the specific use case this is intended to enable is streaming replication.

This is in-line with the suggestion made [here](https://github.com/vapor/postgres-nio/issues/522) (specifically, that replication doesn't belong within postgres-nio and should instead be implemented as a separate library or fork).

These changes are meant to support my implementation of a separate library for streaming replication while making as much reuse as possible from prior art.

### DIFF
--- a/Sources/PostgresNIO/Connection/PostgresConnection.swift
+++ b/Sources/PostgresNIO/Connection/PostgresConnection.swift
@@ -58,11 +58,61 @@ public final class PostgresConnection: @unchecked Sendable {
         assert(self.isClosed, "PostgresConnection deinitialized before being closed.")
     }
 
+    /// Connect to the Postgres server described by `configuration` — negotiate
+    /// TLS (if configured), send the startup message, and authenticate — then
+    /// remove PostgresNIO's handlers, returning the authenticated channel for the
+    /// caller to drive with its own pipeline. Any TLS handler and the socket are
+    /// left in place; only PostgresNIO's own protocol handlers are removed. This
+    /// lets a caller reuse PostgresNIO's TCP/TLS/startup/authentication (including
+    /// SCRAM-SHA-256) and then speak a protocol PostgresNIO does not implement —
+    /// for example, streaming replication.
+    public static func establishAndAuthChannelWithoutPostgresHandler(
+        using configuration: Configuration,
+        on eventLoop: any EventLoop,
+        logger: Logger
+    ) -> EventLoopFuture<any Channel> {
+        let configuration = InternalConfiguration(configuration)
+        return eventLoop.flatSubmit {
+            Self.makeChannel(on: eventLoop, configuration: configuration).flatMap { channel in
+                Self.installHandlersAndStartup(
+                    on: channel,
+                    configuration: configuration,
+                    logger: logger
+                ).flatMap { () -> EventLoopFuture<any Channel> in
+                    do {
+                        let pipeline = channel.pipeline.syncOperations
+                        let channelHandler = try pipeline.context(handlerType: PostgresChannelHandler.self)
+                        let eventsHandler = try pipeline.context(handlerType: PSQLEventsHandler.self)
+                        return pipeline.removeHandler(context: channelHandler)
+                            .and(pipeline.removeHandler(context: eventsHandler))
+                            .map { _ in channel }
+                    } catch {
+                        return channel.eventLoop.makeFailedFuture(error)
+                    }
+                }
+            }
+        }
+    }
+
     func start(configuration: InternalConfiguration) -> EventLoopFuture<Void> {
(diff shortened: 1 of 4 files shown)


CrownedPhoenix:
As context, I'm actually using this on my branch to implement a channel handler over the streaming replication protocol - so I have that confidence that it serves at least my purposes well.

CrownedPhoenix:
Also, I'm happy to spell these changes differently if there's a more consistent pattern/spelling.

gwynne:
Other than TLS support, why does the existing `establishedChannel` option for using an existing channel for a Postgres connection not suffice? (per [`PostgresConnection.Configuration.init(establishedChannel:username:password:database:)`](https://api.vapor.codes/postgresnio/postgresconnection/configuration/init(establishedchannel:username:password:database:)/)).

As for TLS with respect to that option, adding the ability to request that PostgresNIO manage TLS when using an existing channel is certainly possible; we chose not to before on account of it raising issues with the usual process wherein an `SSLRequest` is sent to the Postgres server before TLS is initiated (see [PostgreSQL Documentation Chapter 18 § 54.2.10. SSL Session Encrpytion](https://www.postgresql.org/docs/18/protocol-flow.html#PROTOCOL-FLOW-SSL)).

CrownedPhoenix:
> Other than TLS support, why does the existing `establishedChannel` option for using an existing channel for a Postgres connection not suffice? (per [`PostgresConnection.Configuration.init(establishedChannel:username:password:database:)`](https://api.vapor.codes/postgresnio/postgresconnection/configuration/init(establishedchannel:username:password:database:)/)).

Unless I'm missing something, just creating a `PostgresConnection.Configuration` still requires me to initialize an instance of `PostgresConnection` using that configuration. That's not very suitable for my situation because I have my own entity that I expect to retain ownership of the established Channel and by making an instance of `PostgresConnection` I am then wrestling with sharing the channel between the two entities.

In particular, I can't just let the `PostgresConnection` go out of scope because it's deinit assert will trigger if the channel isn't closed.

So this strategy that I'm using here specifically uses static methods to uncouple me from having to initialize a `PostgresConnection` I'm not going to use.

Is there perhaps a strategy I could use that wouldn't have these issues?

gwynne:
Ah, I see, I've misunderstood your use case—you want to reuse the connection establishment logic without the rest of the wire protocol. Now it makes sense. 😆 I'll give this a proper review shortly.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -58,11 +58,61 @@ public final class PostgresConnection: @unchecked Sendable {
...
+                        return pipeline.removeHandler(context: channelHandler)
+                            .and(pipeline.removeHandler(context: eventsHandler))
+                            .map { _ in channel }
+                    } catch {
+                        return channel.eventLoop.makeFailedFuture(error)
+                    }
+                }
+            }
+        }
+    }
```
fabianfett, line 95: we will need at least a test case for this, that verifies that the PostgresChannelHandler and the PSQLEventsHandler where removed.

Side note: We should 100% get rid of the PSQLEventsHandler.

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -58,11 +58,61 @@ public final class PostgresConnection: @unchecked Sendable {
...
+    /// caller to drive with its own pipeline. Any TLS handler and the socket are
+    /// left in place; only PostgresNIO's own protocol handlers are removed. This
+    /// lets a caller reuse PostgresNIO's TCP/TLS/startup/authentication (including
+    /// SCRAM-SHA-256) and then speak a protocol PostgresNIO does not implement —
+    /// for example, streaming replication.
+    public static func prepareChannel(
+        using configuration: Configuration,
+        on eventLoop: any EventLoop,
+        logger: Logger
+    ) -> EventLoopFuture<any Channel> {
```
fabianfett: I think this should get another (more scary) name. Mayve `establishAndAuthChannelWithoutPostgresHandler()`. `prepare` can mean anything

CrownedPhoenix: Renamed. Also had something like `establishBareSSLChannel` but I don't really have much of an opinion.

Sources/PostgresNIO/Connection/PostgresConnection.swift
CrownedPhoenix: Added.

fabianfett (approved):
LGTM. Thank you!

PR #656, opened 2026-08-05
Allow new connection creation to drain queue during graceful shutdown

ptoffy:
This is a follow up to #651, in particular this attempts to fix https://github.com/vapor/postgres-nio/pull/651#discussion_r3702117041.
This allows us to open up new connections during graceful shutdown to allow draining the rest of the queue even if open connections fail

### DIFF
--- a/Sources/ConnectionPoolModule/PoolStateMachine.swift
+++ b/Sources/ConnectionPoolModule/PoolStateMachine.swift
@@ -266,6 +266,14 @@ struct PoolStateMachine<
 
     @inlinable
     mutating func leaseConnection(_ request: Request) -> Action {
+        if self.gracefulShutdownTriggered {
+            // we're in the process of shutting down, reject new requests
+            return .init(
+                request: .failRequest(request, ConnectionPoolError.poolShutdown),
+                connection: .none
+            )
+        }
+
         switch self.poolState {
         case .running:
             // if requestQueue is non-empty and we cannot create more connections add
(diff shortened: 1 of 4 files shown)


Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -441,6 +449,20 @@ struct PoolStateMachine<
...
+
+            if self.connections.isEmpty {
+                self.poolState = .shutDown
+                connectionAction = .cancelEventStreamAndFinalCleanup(timerToCancel.flatMap { [$0] } ?? [])
+            } else {
+                connectionAction = .cancelTimers(timerToCancel.map {[$0]} ?? [])
+            }
+
+            return .init(request: .none, connection: connectionAction)
+        }
```
fabianfett, line 464: I'm pretty sure this should be inside the running state.

ptoffy: We might be in `connectionCreationFailing` or `circuitBreakOpen` and still gracefully shutting down though, so we should exit early in any case not just running

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -717,22 +768,27 @@ struct PoolStateMachine<
 
     @usableFromInline
     mutating func triggerGracefulShutdown() -> Action {
+        if gracefulShutdownTriggered {
+            return .none()
+        }
+
         switch self.poolState {
         case .running, .connectionCreationFailing, .circuitBreakOpen:
-            self.poolState = .shuttingDown
```
0xTim, line 722: Why are we removing the state change here? Shutdown has been called and we're in the process of shutting down so we should be in the shutting down state

PR #658, opened 2026-08-06
Revert "Add graceful shutdown"

ptoffy:
Reverts vapor/postgres-nio#651 in favour of another PR that will be opened shortly after this


gwynne (approved):
Any particular reason?

fabianfett (approved):
The og PR may cancel requests that are in flight even though this is a graceful shutdown. We think we can do this way less invasive and with clearer state machine transitions.

PR #659, opened 2026-08-06
Add graceful shutdown

ptoffy:


Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -827,6 +827,26 @@ extension PoolStateMachine {
...
+                    switch closeConnection(at: index, deleteConnection: false) {
+                    case .close(let closeAction):
+                        cleanup.connections.append(closeAction.connection)
+                        cleanup.timersToCancel.append(contentsOf: closeAction.timersToCancel)
+                    case .cancelTimers(let timers):
+                        cleanup.timersToCancel.append(contentsOf: timers)
+                    case .doNothing:
+                        break
+                    }
+                default:
```
fabianfett: lets make the states here explicit. Please don't use `default`.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -827,6 +827,26 @@ extension PoolStateMachine {
 
         // MARK: Shutdown
 
+        mutating func triggerGracefulShutdown(_ cleanup: inout ConnectionAction.Shutdown) {
```
fabianfett: we should add a comment that this method will cleanup any non leased connection. We should probably call it like: `closeAnyNonLeasedConnection`.

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -738,6 +786,12 @@ struct PoolStateMachine<
     ) -> Action {
         // this connection was busy before
         let requests = self.requestQueue.pop(max: availableContext.info.availableStreams)
+
+        if self.gracefulShutdownTriggered && self.requestQueue.isEmpty {
+            // no more work to take up, start shutting down
+            self.poolState = .shuttingDown
```
fabianfett: i suspect that we will need to force cleanup connections here. do we have a test for that?

Sources/PostgresNIO/Pool/PostgresClient.swift
```diff
@@ -467,6 +467,9 @@ public final class PostgresClient: Sendable, ServiceLifecycle.Service {
     ///
     /// Cancelling the task that executes the ``run()`` method is equivalent to closing the client. Once the task
     /// has been cancelled the client is not able to process any new queries or prepared statements.
+    /// 
+    /// Graceful shutdown is different to cancellation. If the client is gracefully shut down, work continues as
+    /// before until the request queue is drained and work is completed.
```
fabianfett: callout that new work is rejected.

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
```diff
@@ -1030,6 +1030,104 @@ import Testing
...
+            let lease = try await pool.leaseConnection()
+            let leasedConnection = lease.connection
+
+            pool.triggerGracefulShutdown()
+            #expect(leasedConnection.isRunning)
+
+            pool.triggerForceShutdown()
+            #expect(leasedConnection.isClosing)
+
+            pool.releaseConnection(leasedConnection)
```
fabianfett: do we need this for the pool to finish?

Tests/PostgresNIOTests/New/PostgresClientTests.swift
```diff
@@ -38,4 +39,44 @@ import Testing
...
+                backgroundLogger: Logger(label: "test")
+            )
+
+            let elapsed = await ServiceLifecycleTestKit.testGracefulShutdown { gracefulShutdownTrigger in
+                let start = ContinuousClock.now
+                await withTaskGroup(of: Void.self) { group in
+                    // We don't have a server to connect to, so  we'll reach `connectTimeout` (2s).
+                    group.addTask { await client.run() }
+
+                    try? await Task.sleep(for: .milliseconds(200))
```
fabianfett: this is a test where no work is scheduled. i think we should also schedule work, and cancel that after 500ms. does the pool still close correctly?

Tests/PostgresNIOTests/New/PostgresClientTests.swift
```diff
@@ -38,4 +39,44 @@ import Testing
...
+
+                    gracefulShutdownTrigger.triggerGracefulShutdown()
+                }
+                return ContinuousClock.now - start
+            }
+
+            // If this took ~200ms (our sleep time) it would mean the connect timeout wasn't reached,
+            // and the client was canceled instead of gracefully shut down.
```
fabianfett: we never actually cancel connects. so even when we cancel we will wait for more than 200ms.

fabianfett (commented):
That is 100% the correct direction! Thanks so much for pushing through!

Tests/ConnectionPoolModuleTests/ConnectionPoolTests.swift
ptoffy: no actually that's a no-op, force shutdown will deal with it. will remove

fabianfett: please add a comment that we have omitted this deliberately to test the forceful nature.

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -698,37 +706,77 @@ struct PoolStateMachine<
...
     mutating func triggerGracefulShutdown() -> Action {
-        fatalError("Unimplemented")
-    }
+        if gracefulShutdownTriggered { return .none() }
 
-    @usableFromInline
-    mutating func triggerForceShutdown() -> Action {
         switch self.poolState {
         case .running, .connectionCreationFailing, .circuitBreakOpen:
+            self.gracefulShutdownTriggered = true
```
adam-fowler: Is the idea we want gracefulShutdown to wait until connections are available if we are in the state connectionCreationFailing? The circuitBreakOpen state will go straight to shutdown as we have already failed all the requests.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
0xTim: We should trigger a state change to shuttingDown here. Leaving it as running feels like false advertising

fabianfett: @0xTim I very much disagree. Graceful Shutdown means basically continuing to run normally with two caveats:

- do not accept any new work
- once all work is done shut everything down

This means if connections fail, those have to be retried. The pool would wait until a connection is available again (except if the circuit breaker has tripped).

`shuttingDown` means, we are in a state in which we actually close connections. Connection closures can happen in two ways:

- forceful (we don't wait until a connection has been released)
- graceful (we wait until the connection has been returned to the pool)

Therefore graceful shutdown is a substate of the three states:

- running
- connectionCreationFailing
- shuttingDown

I agree that we should likely move the property into the three substates:

- ConnectionCreationFailingContext
- RunningContext (new type)
- ShuttingDownContext (new type)

Sources/ConnectionPoolModule/PoolStateMachine.swift
fabianfett: @adam-fowler great callout on going straight to closed, if we are in circuitBreakerOpen, or if we transition into circuitBreakerOpen.

@ptoffy we need a test case for gracefulShutdown is triggered when the circuit breaker is open. And another one: We have work in the queue, but fail to create connections. Graceful shutdown is triggered (must not close the pool), timer fires for circuit breaker (pool goes straight to closed).

Sources/ConnectionPoolModule/PoolStateMachine.swift
ptoffy: hmm why? connections are closed lazily as their lifecycle proceeds. we can add a test that connections in all states get closed during graceful shutdown but I think most of them are covered across the other tests

Sources/ConnectionPoolModule/PoolStateMachine.swift
ptoffy: https://github.com/vapor/postgres-nio/pull/659/changes/2e6ce234c37740d115731bc0ee350d1dd07eeaa4

Sources/ConnectionPoolTestUtils/MockConnection.swift
```diff
@@ -19,6 +19,20 @@ public final class MockConnection: PooledConnection, Sendable {
         self.id = id
     }
 
+    public var isRunning: Bool {
+        self.lock.withLockedValue { state in
+            if case .running = state { return true }
+            return false
```
fabianfett: please use a full switch statement here. This will remind us to check this place if we ever add a new state

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -698,37 +769,82 @@ struct PoolStateMachine<
         var requests: [Request]
     }
 
+    @usableFromInline
     mutating func triggerGracefulShutdown() -> Action {
-        fatalError("Unimplemented")
+        if gracefulShutdownTriggered { return .none() }
```
fabianfett, line 774: please follow the code pattern in this repo and add an explicit `self` everywhere.

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -231,9 +257,16 @@ struct PoolStateMachine<
     @usableFromInline
     /*private*/ var requestQueue: RequestQueue
     @usableFromInline
-    /*private*/ var poolState: PoolState = .running
+    /*private*/ var poolState: PoolState = .running(.init(gracefulShutdownTriggered: false))
     @usableFromInline
-    /*private*/ var gracefulShutdownTriggered: Bool = false
+    /*private*/ var gracefulShutdownTriggered: Bool {
```
fabianfett, line 262: if this is a computed property, make the above line `@inlinable`

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -534,11 +601,15 @@ struct PoolStateMachine<
...
+        case .shuttingDown(let context) where context.gracefulShutdownTriggered:
+            // the connection needs to be closed here
+            break
+
         case .shuttingDown, .shutDown:
             return .none()
         }
 
-        switch self.connections.backoffDone(connectionID, retry: true) {
+        switch self.connections.backoffDone(connectionID, retry: !(isShuttingDown && gracefulShutdownTriggered)) {
```
fabianfett:
```suggestion
        switch self.connections.backoffDone(connectionID, retry: !(self.isShuttingDown && self.gracefulShutdownTriggered)) {
```

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -698,37 +769,83 @@ struct PoolStateMachine<
         var requests: [Request]
     }
 
+    @usableFromInline
     mutating func triggerGracefulShutdown() -> Action {
-        fatalError("Unimplemented")
+        if gracefulShutdownTriggered { return .none() }
```
fabianfett:
```suggestion
        if self.gracefulShutdownTriggered { return .none() }
```

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -698,37 +769,83 @@ struct PoolStateMachine<
...
+            self.poolState = .connectionCreationFailing(context)
+
+        case .circuitBreakOpen:
+            self.poolState = .shuttingDown(.init(gracefulShutdownTriggered: true))
+
+        case .shuttingDown, .shutDown:
+            return .none()
+        }
+
+        guard requestQueue.isEmpty else {
```
fabianfett:
```suggestion
        guard self.requestQueue.isEmpty else {
```

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -827,6 +827,27 @@ extension PoolStateMachine {
 
         // MARK: Shutdown
 
+        @usableFromInline
+        mutating func closeAnyNonLeasedConnection(_ cleanup: inout ConnectionAction.Shutdown) {
+            for index in self.connections.indices {
+                switch self.connections[index].state {
+                case .idle, .backingOff:
+                    switch closeConnection(at: index, deleteConnection: false) {
```
fabianfett:
```suggestion
                    switch self.closeConnection(at: index, deleteConnection: false) {
```

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -827,6 +827,27 @@ extension PoolStateMachine {
 
         // MARK: Shutdown
 
+        @usableFromInline
+        mutating func closeAnyNonLeasedConnection(_ cleanup: inout ConnectionAction.Shutdown) {
```
fabianfett: can we collapse this with `triggerForceShutdown` into one method. New name `closeConnections(onlyNonLeased: Bool)`. It's also fine if triggerForceShutdown calls this method then.

Sources/ConnectionPoolTestUtils/MockConnection.swift
```diff
@@ -19,6 +19,22 @@ public final class MockConnection: PooledConnection, Sendable {
...
+            switch state {
+            case .running: return true
+            case .closing, .closed: return false
+            }
+        }
+    }
+
+    public var isClosing: Bool {
+        self.lock.withLockedValue { state in
+            if case .closing = state { return true }
```
fabianfett: please also use a full switch here.

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -119,19 +119,43 @@ struct PoolStateMachine<
...
+        @usableFromInline
+        struct ShuttingDownContext: Sendable {
+            @usableFromInline
+            init(gracefulShutdownTriggered: Bool) {
+                self.gracefulShutdownTriggered = gracefulShutdownTriggered
+            }
+
+            @usableFromInline
+            var gracefulShutdownTriggered: Bool
+        }
```
fabianfett: right now this seems quite excessive for something that we could solve with just an attached Bool. Let's use:

```
case running(gracefulShutdownTriggered: Bool)
case shuttingDown(gracefulShutdownTriggered: Bool)
```

for now

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -171,7 +197,7 @@ struct PoolStateMachine<
         /// Can transition to:
         ///   - `shuttingDown` if the pool is being shut down (graceful shutdown behavior is managed by an external flag),
```
fabianfett: stale doc comment:

```
(graceful shutdown behavior is managed by an external flag),
```

Sources/ConnectionPoolModule/PoolStateMachine.swift
fabianfett: "is managed by an external `gracefulShutdownTriggered` flag," stale doc comment

Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -144,6 +168,8 @@ struct PoolStateMachine<
             var numberOfFailedAttempts: Int
             @usableFromInline
             var connectionIDToRetry: ConnectionID
+            @usableFromInline
+            var gracefulShutdownTriggered: Bool
```
fabianfett, line 150: this is good, as we already have quite some associated state

Sources/ConnectionPoolModule/PoolStateMachine.swift
fabianfett: Not a change in this PR, but related:
```suggestion
            let closedConnectionAction = self.connections.connectionClosed(connection.id, shuttingDown: false)
```
We do not want to shutdown until we are actually shutting down. Our state doesn't indicate that.

Please add this test case to the connection pool state machine tests:
```swift
    @available(macOS 13.0, iOS 16.0, tvOS 16.0, watchOS 9.0, *)
    @Test func connectionDiesDuringGracefulDrainMinOne() {
        var configuration = PoolConfiguration()
        configuration.minimumConnectionCount = 1
        configuration.maximumConnectionSoftLimit = 1
        configuration.maximumConnectionHardLimit = 1
        configuration.keepAliveDuration = nil

        var sm = TestPoolStateMachine(
            configuration: configuration, generator: .init(),
            timerCancellationTokenType: MockTimerCancellationToken.self, clock: MockClock()
        )
        _ = sm.refillConnections()
        let conn = MockConnection(id: 0)
        _ = sm.connectionEstablished(conn, maxStreams: 1)
        let r1 = MockRequest(connectionType: MockConnection.self)
        _ = sm.leaseConnection(r1)
        let r2 = MockRequest(connectionType: MockConnection.self)
        _ = sm.leaseConnection(r2)
        #expect(sm.requestQueue.count == 1)

        _ = sm.triggerGracefulShutdown()

        _ = sm.connectionWillClose(conn.id)
        _ = sm.releaseConnection(conn, streams: 1)
        let closed = sm.connectionClosed(conn)
        guard case .makeConnection = closed.connection else {
            Issue.record("No replacement connection even though minimumConnectionCount == 1 -> r2 stranded, pool never shuts down")
            return
        }
    }
```

fabianfett (commented):
This looks really good now. We are really close!

fabianfett (approved):
This is great! Thank you!

PR #661, opened 2026-08-11
Add group interpolation for allowing `IN` usage

ptoffy:


Tests/PostgresNIOTests/New/PostgresQueryTests.swift
```diff
@@ -97,6 +97,91 @@ final class PostgresQueryTests: XCTestCase {
         XCTAssertEqual(query.binds.bytes, expected)
     }
 
+    func testStringInterpolationWithGroup() throws {
+        let ids = [1, 2, 3]
+
+        let query: PostgresQuery = "SELECT * FROM foo WHERE id IN (\(group: ids));"
```
MahdiBM, line 103: I'd agree it's not the most-discoverable thing in Postgres to find out and know, but I'd say the better syntax here would be:

```sql
WHERE id = any(\(ids))
```

It also won't care if the array is empty. IIRC the `IN` syntax would throw an error or something.

Sources/PostgresNIO/New/PostgresQuery.swift
```diff
@@ -88,6 +88,26 @@ extension PostgresQuery {
...
 
+        @inlinable
+        public mutating func appendInterpolation<Value: PostgresThrowingDynamicTypeEncodable>(group: some Collection<Value>) throws {
+            var first = true
+            for value in group {
+                if first { first = false } else { self.sql.append(", ") }
+                try self.binds.append(value, context: .default)
+                self.sql.append(contentsOf: "$\(self.binds.count)")
+            }
+        }
```
MahdiBM, line 99: This is fine but if we could instead find a way to instruct users to simply use the `= any` syntax, that'd be the optimal solution IMO. It will be only 1 bind for PostgresNIO to append, and a smaller `String` to create.

I don't know what PostgresNIO likes to do in these situations, but I'd personally pre-reserve a minimum amount of required `String` space as well. e.g. assume each bind is of minimum length (`$1` -> 2) and also count the `, `s. But of course Fabian's opinion is the one that actually matters.

gwynne: It has been my experience that `= ANY(<array bind>)` is massively more efficient in Postgres, both for bindings overhead and query execution.

FranzBusch (commented):
I agree with the points from @MahdiBM and @gwynne. I don't think we should add this as a helper for `IN` but it still has value (more below).

`= ANY(\(ids))` already works today. `Array` conforms to `PostgresEncodable` when its element conforms to `PostgresArrayEncodable`, so an array binds as a single native Postgres array. This PR adds new public API for something the library can already do, and the new way is less efficient.

There is one more problem that hasn't been mentioned yet. `IN ($1, $2, $3)` produces a different SQL string for every number of elements. Every cache that is keyed on the query text sees a new statement. That affects the server side plan cache and our own prepared statement API. A user that pages through ID lists of varying length will fill up those caches. `= ANY($1)` is always a single statement.

The empty case is also broken. The test asserts this SQL:

```swift
XCTAssertEqual(query.sql, "SELECT * FROM foo WHERE id IN ();")
```

AFAIK `IN ()` is a syntax error in Postgres. We should enforce at runtime that this isn't possible.

That said, group interpolation is useful for cases that `= ANY` can't express: multi row inserts, row and composite comparisons, and element types that have no array OID. I would reframe it around those cases. I would also rename the label, since `group` says very little at the call site; `list` or `values` reads better. And please document that users should prefer `= ANY` for `IN`, and mention the statement cache behaviour.

Two smaller notes on the code itself. The new API has no doc comments, and it is missing the `context:` overload that the single value interpolation has, so users can't pass a custom JSON encoder. 

@fabianfett WDYT?

PR #662, opened 2026-08-12
ConnectionPool: fix stream accounting leaks and mislabeled idle timer

robertoscipionecom:
The pool's per-group stream counters (`availableStreams`, `leasedStreams`, both `UInt16`) can drift upwards without ever coming back, until an arithmetic operation on them traps. We hit this in production as an `EXC_BREAKPOINT` / *arithmetic overflow* inside `PoolStateMachine.ConnectionGroup.releaseConnection(_:streams:)`, on an app that had been talking to unreachable PostgreSQL servers for hours after a machine wake.

Four independent defects, each fixed in a separate hunk:

**1. `ConnectionGroup.keepAliveFailed`, `.idle` branch** — the stream used by the running keep-alive is counted by `keepAliveIfIdle` but never subtracted. The connection moves to `.closing`, where `closed()` reports `usedStreams == 0`, so the stream stays counted forever. (The `.leased` branch already subtracts it.)

**2. `ConnectionGroup.closeConnection(at:deleteConnection:)`** — same stream lost when the pool closes a connection whose keep-alive is running (force-shutdown path). Leased streams are deliberately left counted: they are subtracted when the outstanding lease is released.

**3. `ConnectionState.closing` did not carry the streams still leased on the connection.** If the connection died before the lease came back, `closed()` reported 0 streams, the connection was removed from the group, and the late `releaseConnection` found no connection to subtract from. `.closing` now carries `usedStreams`, decremented on release and reported by `closed()`. `closed()` for `.closing` and `.draining` now reports `maxStreams == usedStreams`, so the group's `availableStreams -= maxStreams - usedStreams` is zero instead of a `UInt16` underflow — the free streams were already removed at close time. (The `.draining` case could already underflow today with `usedStreams > 0`.)

**4. `ConnectionState.parkConnection`** — in two branches the *idle timeout* timer is built with `usecase: .keepAlive`. It therefore runs for the keep-alive duration instead of the idle timeout, the idle timeout never fires for those connections, and when the timer does fire it is dispatched to `keepAliveIfIdle` on a connection that already has a keep-alive running → `preconditionFailure("Invalid state")`.

### How these were found

Defects 1–3 were found with a state-machine auditor that, after every transition, compares the group counters against the sum of the individual connection states, and tracks owed streams per connection. It was driven by a fuzzer reproducing the production conditions: unreachable servers, failing keep-alives, connections dying mid-query, failing connection attempts, cancelled lease requests, pools shut down with outstanding leases, and request queues under pressure. Defect 4 was reproduced directly.

Defect 3's first attempt introduced the `maxStreams`/`usedStreams` underflow described above; it was caught by your own `testCircuitBreaker`, which is why that part of the fix is included.

### Verification

- `swift test --filter ConnectionPoolModuleTests` — 108 tests in 10 suites, all passing.
- Fuzz campaign (7 seeds × 4 failure profiles × 4 configurations, including shutdown storms and 40 concurrent leases over 5 connections) — no counter divergence.

Happy to add regression tests in the style of the existing suite if you'd like them in this PR.

### DIFF
--- a/Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
+++ b/Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
@@ -587,6 +587,12 @@ extension PoolStateMachine {
             case .idle:
                 self.stats.idle -= 1
                 self.stats.closing += 1
+                // The stream used by the running keep-alive was counted in
+                // `leasedStreams` by `keepAliveIfIdle`. It must be returned here: the
+                // connection moves to `.closing`, where nobody subtracts it anymore
+                // (`closed()` from `.closing` reports usedStreams = 0), so it would
+                // stay counted forever and eventually overflow `leasedStreams`.
+                self.stats.leasedStreams -= closeAction.usedStreams
             case .leased:
                 self.stats.leased -= 1
                 self.stats.leasedStreams -= closeAction.usedStreams
@@ -654,6 +660,12 @@ extension PoolStateMachine {
 
             self.stats.runningKeepAlive -= closeAction.runningKeepAlive ? 1 : 0
             self.stats.availableStreams -= closeAction.maxStreams - closeAction.usedStreams
+            // The stream used by a running keep-alive must be returned here: the
+            // connection moves to `.closing`, where `keepAliveSucceeded`/`keepAliveFailed`
+            // are no-ops. The LEASED streams instead stay counted on purpose: they are
+            // subtracted when the outstanding lease is released.
+            let keepAliveStream: UInt16 = (closeAction.runningKeepAlive && self.keepAliveReducesAvailableStreams) ? 1 : 0
+            self.stats.leasedStreams -= keepAliveStream
 
             switch closeAction.previousConnectionState {
             case .idle:
--- a/Sources/ConnectionPoolModule/PoolStateMachine+ConnectionState.swift
+++ b/Sources/ConnectionPoolModule/PoolStateMachine+ConnectionState.swift
@@ -123,7 +123,7 @@ extension PoolStateMachine {
             /// to: `.closing`, `.closed`.
             case draining(Connection, usedStreams: UInt16)
             /// The pool has given the order to the connection to close. Valid transitions to: `.closed`
-            case closing(Connection)
+            case closing(Connection, usedStreams: UInt16)
             /// The connection is closed. Final state.
             case closed
         }
@@ -315,7 +315,12 @@ extension PoolStateMachine {
 
                 if scheduleIdleTimeoutTimer {
                     idleTimerState = self._nextTimer()
-                    idleTimer = ConnectionTimer(timerID: idleTimerState!.timerID, connectionID: self.id, usecase: .keepAlive)
+                    // This is the IDLE TIMEOUT timer: with `usecase: .keepAlive` it ran for
+                    // the keep-alive duration instead of the idle timeout, the idle timeout
+                    // never fired for these connections, and if the timer did fire it was
+                    // dispatched to `keepAliveIfIdle` on a connection that already had a
+                    // keep-alive running → preconditionFailure.
+                    idleTimer = ConnectionTimer(timerID: idleTimerState!.timerID, connectionID: self.id, usecase: .idleTimeout)
                 }
                 self.state = .idle(connection, maxStreams: maxStreams, keepAlive: .scheduled(keepAliveTimer), idleTimer: idleTimerState)
                 return Max2Sequence(idleTimer, nil)
@@ -323,7 +328,12 @@ extension PoolStateMachine {
             case .idle(let connection, let maxStreams, keepAlive: .running(let usingStream), idleTimer: .none):
                 if scheduleIdleTimeoutTimer {
                     idleTimerState = self._nextTimer()
-                    idleTimer = ConnectionTimer(timerID: idleTimerState!.timerID, connectionID: self.id, usecase: .keepAlive)
+                    // This is the IDLE TIMEOUT timer: with `usecase: .keepAlive` it ran for
+                    // the keep-alive duration instead of the idle timeout, the idle timeout
+                    // never fired for these connections, and if the timer did fire it was
+                    // dispatched to `keepAliveIfIdle` on a connection that already had a
+                    // keep-alive running → preconditionFailure.
+                    idleTimer = ConnectionTimer(timerID: idleTimerState!.timerID, connectionID: self.id, usecase: .idleTimeout)
                 }
                 self.state = .idle(connection, maxStreams: maxStreams, keepAlive: .running(usingStream), idleTimer: idleTimerState)
                 return Max2Sequence(keepAliveTimer, idleTimer)
@@ -471,14 +481,20 @@ extension PoolStateMachine {
                 precondition(usedStreams >= returnedStreams)
                 let newUsedStreams = usedStreams - returnedStreams
                 if newUsedStreams == 0 {
-                    self.state = .closing(connection)
+                    self.state = .closing(connection, usedStreams: 0)
                     return .drainingComplete(connection)
                 } else {
                     self.state = .draining(connection, usedStreams: newUsedStreams)
                     return .none
                 }
 
-            case .closing, .closed:
+            case .closing(let connection, let usedStreams):
+                // A lease comes back on a connection the pool already closed: the
+                // streams must be subtracted here too, otherwise they stay counted.
+                self.state = .closing(connection, usedStreams: usedStreams >= returnedStreams ? usedStreams - returnedStreams : 0)
+                return .none
+
+            case .closed:
                 return .none
 
             case .backingOff, .starting, .idle:
@@ -672,7 +688,7 @@ extension PoolStateMachine {
         mutating func closeIfIdle() -> CloseAction? {
             switch self.state {
             case .idle(let connection, let maxStreams, var keepAlive, let idleTimerState):
-                self.state = .closing(connection)
+                self.state = .closing(connection, usedStreams: 0)
                 return CloseAction(
                     connection: connection,
                     previousConnectionState: .idle,
@@ -717,7 +733,7 @@ extension PoolStateMachine {
                 return nil
 
             case .idle(let connection, let maxStreams, var keepAlive, let idleTimerState):
-                self.state = .closing(connection)
+                self.state = .closing(connection, usedStreams: 0)
                 return CloseAction(
                     connection: connection,
                     previousConnectionState: .idle,
@@ -731,7 +747,10 @@ extension PoolStateMachine {
                 )
 
             case .leased(let connection, usedStreams: let usedStreams, maxStreams: let maxStreams, var keepAlive):
-                self.state = .closing(connection)
+                // Streams still leased stay charged to the connection while closing:
+                // they are subtracted on release, or by `closed()` if the connection
+                // dies first. Otherwise they would be lost from the accounting.
+                self.state = .closing(connection, usedStreams: usedStreams)
                 return CloseAction(
                     connection: connection,
                     previousConnectionState: .leased,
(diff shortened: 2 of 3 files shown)


fabianfett (commented):
Thanks for reporting and proposing this initial fix. can we move some of the comments into this PR review pane.

One thing that would be important to me, can we please add a test that showed the issue previously and shows that it is indeed now fixed. Extending an existing test is also fine – maybe even preferred.

robertoscipionecom:
Added the four regression tests to `PoolStateMachine_ConnectionGroupTests`, one per defect:

- `failingKeepAliveOnIdleConnectionReturnsItsStream`
- `closingConnectionWithRunningKeepAliveReturnsItsStream`
- `connectionDyingWhileClosingReturnsItsLeasedStreams`
- `parkingConnectionWithRunningKeepAliveSchedulesAnIdleTimeoutTimer`

Each one fails on 1.33.1 and passes with the fixes (verified by reverting the two source files to the tag and running them individually). Full suite: 112 tests, all green.

robertoscipionecom (commented):
@fabianfett moved the reasoning for each hunk into the review pane, as asked — every comment now sits on the line it explains, in diff order.

One thing worth your call: the same explanations currently also live as comment blocks in the two source files. If you would rather they not stay in the code permanently, say the word and I'll strip them down to one-liners (or remove them entirely) — the full reasoning stays here in the review either way.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -587,6 +587,12 @@ extension PoolStateMachine {
             case .idle:
                 self.stats.idle -= 1
                 self.stats.closing += 1
+                // The stream used by the running keep-alive was counted in
+                // `leasedStreams` by `keepAliveIfIdle`. It must be returned here: the
+                // connection moves to `.closing`, where nobody subtracts it anymore
+                // (`closed()` from `.closing` reports usedStreams = 0), so it would
+                // stay counted forever and eventually overflow `leasedStreams`.
+                self.stats.leasedStreams -= closeAction.usedStreams
```
robertoscipionecom, line 595: **Defect 1 — the keep-alive stream is lost when the keep-alive fails on an idle connection.**

`keepAliveIfIdle` charges one stream to `leasedStreams`. When that keep-alive fails on an *idle* connection, the connection moves to `.closing`, and `closed()` out of `.closing` used to report `usedStreams == 0` — so nobody ever gave the stream back. `leasedStreams` drifts up by one per failed keep-alive, which is exactly what an unreachable server produces for hours.

The `.leased` branch immediately below already performs this same subtraction.

Covered by `failingKeepAliveOnIdleConnectionReturnsItsStream`.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionGroup.swift
```diff
@@ -654,6 +660,12 @@ extension PoolStateMachine {
 
             self.stats.runningKeepAlive -= closeAction.runningKeepAlive ? 1 : 0
             self.stats.availableStreams -= closeAction.maxStreams - closeAction.usedStreams
+            // The stream used by a running keep-alive must be returned here: the
+            // connection moves to `.closing`, where `keepAliveSucceeded`/`keepAliveFailed`
+            // are no-ops. The LEASED streams instead stay counted on purpose: they are
+            // subtracted when the outstanding lease is released.
+            let keepAliveStream: UInt16 = (closeAction.runningKeepAlive && self.keepAliveReducesAvailableStreams) ? 1 : 0
+            self.stats.leasedStreams -= keepAliveStream
```
robertoscipionecom, line 668: **Defect 2 — the same stream is lost on the force-shutdown path.**

`closeConnection(at:deleteConnection:)` can close a connection with a keep-alive in flight; that stream was charged to `leasedStreams` and nothing returned it.

The asymmetry here is deliberate: only the keep-alive stream is returned. Streams held by an outstanding *lease* stay charged, because they are subtracted when the lease is released — or by `closed()` if the connection dies first (see the `.closing` handling in `ConnectionState`). Exactly one of the two runs.

Covered by `closingConnectionWithRunningKeepAliveReturnsItsStream`.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionState.swift
```diff
@@ -123,7 +123,7 @@ extension PoolStateMachine {
             /// to: `.closing`, `.closed`.
             case draining(Connection, usedStreams: UInt16)
             /// The pool has given the order to the connection to close. Valid transitions to: `.closed`
-            case closing(Connection)
+            case closing(Connection, usedStreams: UInt16)
```
robertoscipionecom, line 126: **This is the one structural change in the PR; most of the rest follows from it.**

`.closing` was the only state that forgot how many streams were still out on the connection. So a connection that died while a lease was outstanding took those streams with it: the group had already dropped the connection, and the late `releaseConnection` had nothing left to subtract from.

The three call sites that enter `.closing` now pass what is actually owed — `0` from idle, `usedStreams` from leased and draining.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionState.swift
```diff
@@ -315,15 +315,25 @@ extension PoolStateMachine {
 
                 if scheduleIdleTimeoutTimer {
                     idleTimerState = self._nextTimer()
-                    idleTimer = ConnectionTimer(timerID: idleTimerState!.timerID, connectionID: self.id, usecase: .keepAlive)
+                    // This is the IDLE TIMEOUT timer: with `usecase: .keepAlive` it ran for
+                    // the keep-alive duration instead of the idle timeout, the idle timeout
+                    // never fired for these connections, and if the timer did fire it was
+                    // dispatched to `keepAliveIfIdle` on a connection that already had a
+                    // keep-alive running → preconditionFailure.
+                    idleTimer = ConnectionTimer(timerID: idleTimerState!.timerID, connectionID: self.id, usecase: .idleTimeout)
```
robertoscipionecom, line 323: **Defect 4 — the idle-timeout timer was built with `usecase: .keepAlive`.**

Three consequences, in order of severity:

1. it ran for the keep-alive duration instead of the idle timeout;
2. the idle timeout therefore never fired for these connections;
3. when it did fire, the pool dispatched it to `keepAliveIfIdle` on a connection that already had a keep-alive running → `preconditionFailure("Invalid state")`.

The value is stored in the `idleTimer:` slot two lines below, so the intent of the timer is not ambiguous — it was simply the wrong `usecase`.

Covered by `parkingConnectionWithRunningKeepAliveSchedulesAnIdleTimeoutTimer`.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionState.swift
```diff
@@ -315,15 +315,25 @@ extension PoolStateMachine {
...
             case .idle(let connection, let maxStreams, keepAlive: .running(let usingStream), idleTimer: .none):
                 if scheduleIdleTimeoutTimer {
                     idleTimerState = self._nextTimer()
-                    idleTimer = ConnectionTimer(timerID: idleTimerState!.timerID, connectionID: self.id, usecase: .keepAlive)
+                    // This is the IDLE TIMEOUT timer: with `usecase: .keepAlive` it ran for
+                    // the keep-alive duration instead of the idle timeout, the idle timeout
+                    // never fired for these connections, and if the timer did fire it was
+                    // dispatched to `keepAliveIfIdle` on a connection that already had a
+                    // keep-alive running → preconditionFailure.
+                    idleTimer = ConnectionTimer(timerID: idleTimerState!.timerID, connectionID: self.id, usecase: .idleTimeout)
```
robertoscipionecom, line 336: Same fix as the branch above. This is the `keepAlive: .running` variant — the one where the mislabeled timer could actually reach `keepAliveIfIdle` on a connection with a keep-alive already in flight, and trip the precondition.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionState.swift
```diff
@@ -471,14 +481,20 @@ extension PoolStateMachine {
...
                 } else {
                     self.state = .draining(connection, usedStreams: newUsedStreams)
                     return .none
                 }
 
-            case .closing, .closed:
+            case .closing(let connection, let usedStreams):
+                // A lease comes back on a connection the pool already closed: the
+                // streams must be subtracted here too, otherwise they stay counted.
+                self.state = .closing(connection, usedStreams: usedStreams >= returnedStreams ? usedStreams - returnedStreams : 0)
```
robertoscipionecom, line 494: A lease coming back on a connection the pool has already ordered to close used to land in `case .closing, .closed: return .none` and silently disappear from the accounting. Now the returned streams are subtracted from what `.closing` carries.

On the ternary: it is a saturating subtraction rather than the `precondition(usedStreams >= returnedStreams)` used in the `.draining` branch above. I chose not to add a new trap on a path that is already handling a connection on its way out, but I have no attachment to that — happy to switch it to a `precondition` for symmetry if you prefer to fail loudly on an over-return.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionState.swift
```diff
@@ -731,7 +747,10 @@ extension PoolStateMachine {
                 )
 
             case .leased(let connection, usedStreams: let usedStreams, maxStreams: let maxStreams, var keepAlive):
-                self.state = .closing(connection)
+                // Streams still leased stay charged to the connection while closing:
+                // they are subtracted on release, or by `closed()` if the connection
+                // dies first. Otherwise they would be lost from the accounting.
+                self.state = .closing(connection, usedStreams: usedStreams)
```
robertoscipionecom, line 753: Entering `.closing` from `.leased`, the streams still out are carried over instead of dropped.

They are then returned by exactly one of two paths: `releaseConnection` (the `.closing` case above) if the lease comes back, or `closed()` if the connection dies first.

Covered by `connectionDyingWhileClosingReturnsItsLeasedStreams`.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionState.swift
```diff
@@ -835,18 +854,29 @@ extension PoolStateMachine {
                 return ClosedAction(
                     previousConnectionState: .leased,
                     cancelTimers: .init(),
-                    maxStreams: 0,
+                    // `maxStreams` must not be smaller than `usedStreams`: the group
+                    // computes `availableStreams -= maxStreams - usedStreams` in UInt16.
+                    // The free streams were already removed when the connection was
+                    // marked for close, so nothing is left to remove here: reporting
+                    // maxStreams == usedStreams makes that subtraction exactly zero.
+                    maxStreams: usedStreams,
```
robertoscipionecom, line 862: **This one is a pre-existing underflow in `.draining`, reachable on today's `main`.**

The group computes `availableStreams -= maxStreams - usedStreams` in `UInt16`. Reporting `maxStreams: 0` while `usedStreams > 0` makes that `0 - usedStreams` → trap.

The free streams were already removed from `availableStreams` when the connection was marked for close, so there is nothing left to remove here. `maxStreams == usedStreams` makes the subtraction exactly zero, which is also the truthful value.

This is the part your own `testCircuitBreaker` caught when my first attempt at defect 3 got it wrong, which is why it ended up in this PR.

Sources/ConnectionPoolModule/PoolStateMachine+ConnectionState.swift
```diff
@@ -835,18 +854,29 @@ extension PoolStateMachine {
...
                 self.state = .closed
                 return ClosedAction(
                     previousConnectionState: .closing,
                     cancelTimers: .init(),
-                    maxStreams: 0,
-                    usedStreams: 0,
+                    // Same reason as the `.draining` case above: the available streams
+                    // were already removed at close time, so `maxStreams - usedStreams`
+                    // must be zero, not a UInt16 underflow.
+                    maxStreams: usedStreams,
```
robertoscipionecom, line 875: Two separate changes in this hunk:

- `maxStreams: usedStreams` — same reason as `.draining` above, it keeps `maxStreams - usedStreams` from underflowing in the group.
- `usedStreams:` is no longer hard-coded to `0` (line below). This is the change that actually closes the leak: a connection dying with streams still leased returns them here, because no `releaseConnection` will ever be able to.

Covered by `connectionDyingWhileClosingReturnsItsLeasedStreams`.

PR #663, opened 2026-08-12
Remove unchecked Sendable from NotificationListener

fabianfett:
### Motivation

- `@unchecked Sendable` is dangerous

### Changes

- Use `NIOLoopBoundBox` in `NotificationListener` to show the compiler what is safe

### Result

One instance of `@unchecked Sendable` less

### DIFF
--- a/Sources/PostgresNIO/New/NotificationListener.swift
+++ b/Sources/PostgresNIO/New/NotificationListener.swift
@@ -1,24 +1,29 @@
 import NIOCore
 
-// This object is @unchecked Sendable, since we synchronize state on the EL
-final class NotificationListener: @unchecked Sendable {
+// Thread safety is guaranteed in the NotificationListener by dispatching all access to the shared
+// state onto the underlying EventLoop.
+final class NotificationListener: Sendable {
     let eventLoop: any EventLoop
 
     let channel: String
     let id: Int
 
-    private var state: State
+    private let stateBox: NIOLoopBoundBox<State>
 
     enum State {
         case streamInitialized(CheckedContinuation<PostgresNotificationSequence, any Error>)
         case streamListening(AsyncThrowingStream<PostgresNotification, any Error>.Continuation)
 
-        case closure(PostgresListenContext, (PostgresListenContext, PostgresMessage.NotificationResponse) -> Void)
+        case closure(PostgresListenContext, @Sendable (PostgresListenContext, PostgresMessage.NotificationResponse) -> Void)
         case done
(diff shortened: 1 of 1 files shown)


Sources/PostgresNIO/New/NotificationListener.swift
```diff
@@ -1,24 +1,29 @@
...
-        case closure(PostgresListenContext, (PostgresListenContext, PostgresMessage.NotificationResponse) -> Void)
+        case closure(PostgresListenContext, @Sendable (PostgresListenContext, PostgresMessage.NotificationResponse) -> Void)
         case done
     }
 
     deinit {
-        switch self.state {
+        // The state may only be inspected on the EventLoop. If we are deinitialized somewhere else,
+        // we can't validate that the listener has been used correctly.
+        guard self.eventLoop.inEventLoop else { return }
```
gwynne, line 24: I take it this is something that can happen too easily for an assertion to be safe?

Sources/PostgresNIO/New/NotificationListener.swift
```diff
@@ -50,104 +58,176 @@ final class NotificationListener: @unchecked Sendable {
...
+                    case .cancelled:
+                        eventLoop.execute {
+                            handlerLoopBound.value.cancelNotificationListener(channel: channel, id: listenerID)
+                        }
+
+                    case .finished:
+                        break
+
+                    @unknown default:
+                        break
```
gwynne, line 103: Should we cancel for unknown termination reasons? Seems like the safer choice IMO.

Sources/PostgresNIO/New/NotificationListener.swift
```diff
@@ -1,24 +1,29 @@
...
+// Thread safety is guaranteed in the NotificationListener by dispatching all access to the shared
+// state onto the underlying EventLoop.
+final class NotificationListener: Sendable {
     let eventLoop: any EventLoop
 
     let channel: String
     let id: Int
 
-    private var state: State
+    private let stateBox: NIOLoopBoundBox<State>
```
gwynne, line 11: Not a nit, just curiosity—how much overhead does `NIOLoopBoundBox` impose? My guess is not very much.

gwynne (approved):

Sources/PostgresNIO/New/NotificationListener.swift
fabianfett: Mostly one additional alloc.

PR #665, opened 2026-08-13
Allow subquery interpolation

ptoffy:
This implements `PostgresQuery` subquery interpolation by analysing the query and renumbering all binds depending on the existing ones in the base query. Implements a lexer based on [postgres' `scan.l`](https://github.com/postgres/postgres/blob/master/src/backend/parser/scan.l)

### DIFF
--- a/Sources/PostgresNIO/New/PostgresQuery+PlaceholderLexer.swift
+++ b/Sources/PostgresNIO/New/PostgresQuery+PlaceholderLexer.swift
@@ -5,6 +5,10 @@ extension PostgresQuery {
     }
 
     enum PlaceholderLexer {
+        /// Finds every top-level bind placeholder (`$1`, `$2`, ...) in the query,
+        /// returning the UTF-8 offset of each placeholder's `$` and its number
+        /// in order of finding. Placeholder-like text inside string literals, quoted
+        /// identifiers, comments, and dollar-quoted strings is not matched.
         static func placeholders(in query: String) -> [Placeholder] {
             let utf8 = query.utf8
             var placeholders: [Placeholder] = []
--- a/Sources/PostgresNIO/New/PostgresQuery+PlaceholderLexer.swift
+++ b/Sources/PostgresNIO/New/PostgresQuery+PlaceholderLexer.swift
@@ -1,0 +1,271 @@
+extension PostgresQuery {
+    struct Placeholder {
+        let position: Int
+        let number: Int
+    }
+
+    enum PlaceholderLexer {
+        static func placeholders(in query: String) -> [Placeholder] {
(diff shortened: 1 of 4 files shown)


Sources/PostgresNIO/New/PostgresQuery.swift
```diff
@@ -33,10 +39,13 @@ extension PostgresQuery {
         var sql: String
         @usableFromInline
         var binds: PostgresBindings
+        @usableFromInline
+        var placeHolderPositions: [Int]
```
fabianfett: puh. allocating an extra array for this feature on every query seems wasteful.

Sources/PostgresNIO/New/PostgresQuery.swift
```diff
@@ -6,22 +6,28 @@ public struct PostgresQuery: Sendable, Hashable {
     public var sql: String
     /// The query binds.
     public var binds: PostgresBindings
+    // UTF-8 offsets of the `$` of each interpolation-written placeholder in `sql`.
```
fabianfett: I think instead of using the number `$` approach we should go back to just `?`. With this we probaly don't need to do the extra bookkeeping.

ptoffy: PSQL uses `$x` on the wire so we'd be simply pushing the `?` -> `$x` down the pipeline. This also means we'd be doing it for every query rather than only for the subquery interpolation ones

ptoffy:
@fabianfett the other approach that would allow this feature would be looking for the placeholders in the query manually instead of keeping track of them while interpolating. So when and only when we need it we'd be feeding the query through a small lexer to find the various `$x` (only the ones that are actual bind placeholders) and renumber them. this would also solve the issue of not being able to track placeholders that come from `unsafeSQL`, but it would require keeping a piece of code to correctly scan the query

fabianfett:
@ptoffy I like the proposed solution.

Sources/PostgresNIO/New/PostgresQuery+PlaceholderLexer.swift
```diff
@@ -0,0 +1,271 @@
+extension PostgresQuery {
+    struct Placeholder {
+        let position: Int
+        let number: Int
+    }
+
+    enum PlaceholderLexer {
+        static func placeholders(in query: String) -> [Placeholder] {
```
fabianfett, line 12: please add a comment here what we are looking for an why.

ptoffy: 75a1e8101393a3ffc45582c115ae7d51dc81e96c

PR #667, opened 2026-08-17
Fix replacement after connection closes with queued request

### LINKED ISSUE #657: Request is stranded if last connection fails with `minimumConnectionCount == 0` and no other request is enqueued after
### Describe the issue

If `minimumConnectionCount` is 0 and there's a request being processed, with no other requests enqueued afterwards and the connection that the work is being done on fails, no subsequent connection is opened to continue the work and the request is left stranded

### Vapor version

Not using Vapor

MacOS 26.4.1 (25E253)

### Swift version

Swift Package Manager - Swift 6.3.2

### Steps to reproduce

```swift
@available(macOS 13.0, iOS 16.0, tvOS 16.0, watchOS 9.0, *)
@Test func testLastConnectionDyingWithQueuedRequestCreatesReplacement() {
    var configuration = PoolConfiguration()
    configuration.minimumConnectionCount = 0
    configuration.maximumConnectionSoftLimit = 1
    configuration.maximumConnectionHardLimit = 1
    configuration.keepAliveDuration = nil

    var stateMachine = TestPoolStateMachine(
        configuration: configuration,
        generator: .init(),
        timerCancellationTokenType: MockTimerCancellationToken.self,
        clock: MockClock()
    )

    // first connection
    let request1 = MockRequest(connectionType: MockConnection.self)
    guard case .makeConnection = stateMachine.leaseConnection(request1).connection else {
        Issue.record()
        return
    }
    let connection1 = MockConnection(id: 0)
    #expect(stateMachine.connectionEstablished(connection1, maxStreams: 1).request == .leaseConnection(.init(element: request1), connection1))

    let request2 = MockRequest(connectionType: MockConnection.self)
    #expect(stateMachine.leaseConnection(request2) == .none())

    // connection dies while request2 is queued
    let closedAction = stateMachine.connectionClosed(connection1)
    #expect(closedAction.request == .none)
    guard case .makeConnection(let replacementRequest, _) = closedAction.connection else {
        Issue.record("Expected a replacement connection for the queued request, got \(closedAction.connection)") // fails here
        return
    }

    // the replacement serves the queued request
    let connection2 = MockConnection(id: replacementRequest.connectionID)
    #expect(stateMachine.connectionEstablished(connection2, maxStreams: 1).request == .leaseConnection(.init(element: request2), connection2))
}
```

### Outcome

The work is completed correctly by creating a new request

### Additional notes

_No response_

EhsanAzish80:
Fixes #657.

When the last connection closed with `minimumConnectionCount` set to zero, the pool only considered its configured minimum and ignored queued lease requests. This left the request stranded when no later request arrived to trigger another connection attempt.

Create a replacement connection when the pool is running, work remains queued, and the existing connection-attempt and hard limits permit it. Backoff and shutdown behavior remain unchanged.

Added a regression test covering the reported sequence.

Tests:

- `testLastConnectionDyingWithQueuedRequestCreatesReplacement`
- `PoolStateMachineTests`: 30 passed
- `ConnectionPoolModuleTests`: 109 passed
- `PostgresNIOTests`: 172 passed

### DIFF
--- a/Sources/ConnectionPoolModule/PoolStateMachine.swift
+++ b/Sources/ConnectionPoolModule/PoolStateMachine.swift
@@ -659,7 +659,19 @@ struct PoolStateMachine<
         case .running, .connectionCreationFailing, .circuitBreakOpen:
             self.cacheNoMoreConnectionsAllowed = false
 
-            let closedConnectionAction = self.connections.connectionClosed(connection.id, shuttingDown: self.gracefulShutdownTriggered)
+            let createConnectionForQueuedRequest: Bool
+            if case .running = self.poolState {
+                createConnectionForQueuedRequest = !self.requestQueue.isEmpty
+                    && self.connections.stats.connecting < self.configuration.maximumConcurrentConnectionRequests
+            } else {
+                createConnectionForQueuedRequest = false
+            }
+
+            let closedConnectionAction = self.connections.connectionClosed(
+                connection.id,
+                shuttingDown: self.gracefulShutdownTriggered,
+                createConnectionForQueuedRequest: createConnectionForQueuedRequest
+            )
 
             let connectionAction: ConnectionAction
             if let newRequest = closedConnectionAction.newConnectionRequest {
@@ -671,7 +683,11 @@ struct PoolStateMachine<
             return .init(request: .none, connection: connectionAction)
 
         case .shuttingDown:
-            let closedConnectionAction = self.connections.connectionClosed(connection.id, shuttingDown: true)
+            let closedConnectionAction = self.connections.connectionClosed(
+                connection.id,
+                shuttingDown: true,
+                createConnectionForQueuedRequest: false
+            )
 
             let connectionAction: ConnectionAction
             if self.connections.isEmpty {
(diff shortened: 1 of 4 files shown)


Sources/ConnectionPoolModule/PoolStateMachine.swift
```diff
@@ -664,6 +664,17 @@ struct PoolStateMachine<
...
+                !self.requestQueue.isEmpty,
+                self.connections.stats.connecting < self.configuration.maximumConcurrentConnectionRequests
+            {
+                let newRequest = self.connections.createNewDemandConnectionIfPossible()
+                    ?? self.connections.createNewOverflowConnectionIfPossible()
+                if let newRequest {
+                    connectionAction = .makeConnection(newRequest, closedConnectionAction.timersToCancel)
+                } else {
+                    connectionAction = .cancelTimers(closedConnectionAction.timersToCancel)
+                }
```
fabianfett: this looks weird. my feeling is that we should forward the info here into:

```
            let closedConnectionAction = self.connections.connectionClosed(connection.id, shuttingDown: self.gracefulShutdownTriggered)
```

and change the behavior there.

fabianfett (commented):
Thanks for the issue and the proposed fix. I think though we should fix this inside the `connections`.

Sources/ConnectionPoolModule/PoolStateMachine.swift
EhsanAzish80: Good point — I moved the replacement decision into `ConnectionGroup.connectionClosed` and kept the queued-demand and connection-request limits in the caller. The focused state-machine suites pass.

PR #669, opened 2026-08-27
Do not close connections when receiving a server error with a 28xxx SQL state

gwynne:
In the current implementation, 28xxx SQL state errors from the server, which are listed as "Class 28 - Invalid Authorization Specification" are treated as connection-fatal. This PR removes that behavior, as there are conditions under which such errors can be generated without meaning that the connection is no longer valid. The most obvious example is the case where a stored procedure `RAISE`s an exception with a 28xxx SQL state to indicate specific error conditions. An actual authentication failure still results in connection closure without this check, and by not closing the connection in case of error, such stored procedures—and other cases where 28xxx SQL states are use—can operate as designed.

### DIFF
--- a/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
+++ b/Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
@@ -1088,12 +1088,6 @@ extension ConnectionStateMachine {
                 // any error message that doesn't have a sql state field, is unexpected by default.
                 return true
             }
-            
-            if sqlState.starts(with: "28") {
-                // these are authentication errors
-                return true
-            }
-            
             return false
         case .clientClosedConnection, .poolClosed:
             preconditionFailure("A pure client error was thrown directly in PostgresConnection, this shouldn't happen")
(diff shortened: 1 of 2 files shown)


0xTim (approved):

Sources/PostgresNIO/New/Connection State Machine/ConnectionStateMachine.swift
```diff
@@ -1088,12 +1088,6 @@ extension ConnectionStateMachine {
                 // any error message that doesn't have a sql state field, is unexpected by default.
                 return true
             }
-            
-            if sqlState.starts(with: "28") {
```
fabianfett, line 1092: Can we add a test that validates that this still correctly fails for auth issues?

gwynne: The existing auth test already does so.

fabianfett: okay, then we also need a test case for the reverse one... 28xxx codes should not close the connection after auth.

ptoffy: Test added

fabianfett (approved):
Thank you!

PR #670, opened 2026-09-03
Clamp Date instead of crashing

### LINKED ISSUE #632: Crash in `Date.encode` if the date is long distant future
### Describe the issue

`Date.encode` method crashes if the date is long distant future

### Vapor version

4.111.1

macOS 26.3.1

### Swift version

6.2.3

### Steps to reproduce

_No response_

### Outcome

_No response_

### Additional notes

https://github.com/vapor/postgres-nio/blob/cc0201cfdc166c0e85af606e617ab7afc3b938be/Sources/PostgresNIO/New/Data/Date%2BPostgresCodable.swift#L18-L19

`Date.encode` internally computes its `seconds` from `_psqlDateStart` and convert it to `Int64`.
If the `seconds` is greater than `Int64.max`, this method crashes.

> Fatal error: Double value cannot be converted to Int64 because the result would be greater than Int64.max

ptoffy:
Fixes #632 

If we tried encoding a date > Int64.max or < Int64.min we crashed. This avoids the crash by clamping to PSQL's `END_TIMESTAMP` and `MIN_TIMESTAMP`. We cannot throw since `Date` is `PostgresNonThrowingEncodable`, so we convert to one of `END_TIMESTAMP` or `MIN_TIMESTAMP - 1` so it's Postgres who returns an error, instead of silently using a different value. Also updates the `Date_PSQLCodableTests` to Swift Testing

### DIFF
--- a/Sources/PostgresNIO/Data/PostgresData+Date.swift
+++ b/Sources/PostgresNIO/Data/PostgresData+Date.swift
@@ -4,8 +4,7 @@ import NIOCore
 extension PostgresData {
     public init(date: Date) {
         var buffer = ByteBufferAllocator().buffer(capacity: 0)
-        let seconds = date.timeIntervalSince(_psqlDateStart) * Double(_microsecondsPerSecond)
-        buffer.writeInteger(Int64(seconds))
+        buffer.writeInteger(date._psqlMicroseconds)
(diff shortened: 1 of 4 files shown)


fabianfett (approved):
LGTM, thank you!

PR #671, opened 2026-09-04
Fix prepared statement result format

### LINKED ISSUE #303: Prepared queries encoding/decoding error
### Describe the bug

Substitution of parameters ($1, $2, ...) to prepared queries fails. E.g. for `Int` value 10 it fails with error message «server: invalid byte sequence for encoding "UTF8": 0x00 (report_invalid_encoding)». I guess the problem is that format (text or binary) is incorrect. The same for unprepared queries works fine.

Decoding of values fails too. Simple `SELECT 10` as prepared query can't decode the result as `Int`, but decodes it as `Data` of 4 bytes (0x00, 0x00, 0x00, 0x0A). In this case the problem is in «Int+PostgresCodable.swift:254» converting "\0\0\0\10" string to `Int`. But if format of the received cells is changed to `.binary` before decoding then values are successfully decoded.

### To Reproduce

That's Xcode tests:
```swift
func testSelectPlain() async throws {
        let postgres = try await Postgres()

        let result = try postgres.connection.query("SELECT 10, 20", logger: postgres.logger).wait()

        let rows = result.rows
        XCTAssertEqual(rows.count, 1)

        let values = try rows.first!.decode((Int, Int).self, context: .default)
        XCTAssertEqual(values.0, 10)
        XCTAssertEqual(values.1, 20)
    }


    func testSelectPlainPrepared() async throws {
        let postgres = try await Postgres()

        let query = try postgres.connection.prepare(query: "SELECT 10, 20").wait()

        let rows = try query.execute().wait()
        XCTAssertEqual(rows.count, 1)

        let values = try rows.first!.decode((Int, Int).self, context: .default)
        XCTAssertEqual(values.0, 10)
        XCTAssertEqual(values.1, 20)
    }


    func testSelectPlainPreparedBinary() async throws {
        let postgres = try await Postgres()

        let query = try postgres.connection.prepare(query: "SELECT 10, 20").wait()

        let rows = try query.execute().wait()
        XCTAssertEqual(rows.count, 1)

        var iterator = rows.first!.makeIterator()
        if var cell = iterator.next() {
            cell.format = .binary
            let value = try cell.decode(Int.self, context: .default)
            XCTAssertEqual(value, 10)
        }
        if var cell = iterator.next() {
            cell.format = .binary
            let value = try cell.decode(Int.self, context: .default)
            XCTAssertEqual(value, 20)
        }
    }


    func testSelectFormatted() async throws {
        let postgres = try await Postgres()

        let result = try postgres.connection.query("SELECT $1, $2", [ 10, 20 ].map { $0.postgresData! }).wait()

        let rows = result.rows
        XCTAssertEqual(rows.count, 1)

        let values = try rows.first!.decode((Int, Int).self, context: .default)
        XCTAssertEqual(values.0, 10)
        XCTAssertEqual(values.1, 20)
    }


    func testSelectFormattedPrepared() async throws {
        let postgres = try await Postgres()

        let query = try postgres.connection.prepare(query: "SELECT $1, $2").wait()

        let rows = try query.execute([ 10, 20 ].map { $0.postgresData! }).wait()
        XCTAssertEqual(rows.count, 1)

        let values = try rows.first!.decode((Int, Int).self, context: .default)
        XCTAssertEqual(values.0, 10)
        XCTAssertEqual(values.1, 20)
    }


    private class Postgres {

        let connection: PostgresConnection

        let logger = Logger(label: "postgres-logger")


        init() async throws {

            do {
                connection = try await PostgresConnection.connect(
                    on: eventLoopGroup.next(),
                    configuration: .init(
                        connection: .init(
                            host: "localhost",
                            port: 5432
                        ),
                        authentication: .init(
                            username: "postgres",
                            database: nil,
                            password: nil
                        ),
                        tls: .disable
                    ),
                    id: 1,
                    logger: logger
                )
                self.eventLoopGroup = eventLoopGroup
            }
            catch {
                try? eventLoopGroup.syncShutdownGracefully()
                throw error
            }
        }

        deinit {
            do {
                try connection.close().wait()
                try eventLoopGroup.syncShutdownGracefully()
            }
            catch { print("Failed to stop instance of KvPostgreSQL with error: \(error)") }
        }


        private let eventLoopGroup: MultiThreadedEventLoopGroup
    }
```

1. Add package with configuration '.package(url: "https://github.com/vapor/postgres-nio.git", from: "1.11.0")'
2. Run tests above.
3. See errors in `testSelectPlainPrepared()` and `testSelectFormattedPrepared()`.
4. See walk around in `testSelectPlainPreparedBinary()`.

### Expected behavior

Correct encoding and decoding of values for prepared queries.

### Environment

* Vapor Framework version: PostgresNIO 1.11.0
* Postgres: PostgreSQL 14.4 on aarch64-apple-darwin20.6.0, compiled by Apple clang version 12.0.5 (clang-1205.0.22.9), 64-bit
* OS version: MacOS 12.4
* Xcode: 13.4.1

ptoffy:
Fixes #303 

We're not correctly parsing metadata, parsing it as text instead of binary. This isn't hit by the new prepared statement APIs but since it's a one line fix we might as well add it

### DIFF
--- a/Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
+++ b/Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
@@ -267,7 +267,7 @@ struct ExtendedQueryStateMachine {
             return .wait
 
         case .prepareStatement(_, _, _, let eventLoopPromise):
-            return .succeedPreparedStatementCreation(eventLoopPromise, with: rowDescription)
+            return .succeedPreparedStatementCreation(eventLoopPromise, with: RowDescription(columns: columns))
         }
     }
     
--- a/Tests/IntegrationTests/PreparedQueryIntegrationTests.swift
+++ b/Tests/IntegrationTests/PreparedQueryIntegrationTests.swift
@@ -1,0 +1,58 @@
+import Logging
+import NIOCore
+import NIOPosix
+import PostgresNIO
+import Testing
+
+@Suite
+struct PreparedQueryIntegrationTests {
+    @Test(.bug("https://github.com/vapor/postgres-nio/issues/303"))
+    func selectPlain() async throws {
+        try await withTestConnection(on: MultiThreadedEventLoopGroup.singleton.any()) { connection in
+            let result = try await connection.query("SELECT 10, 20", logger: .psqlTest).collect()
+
+            #expect(result.count == 1)
+            let values = try #require(result.first).decode((Int, Int).self)
+            #expect(values == (10, 20))
+        }
+    }
+
+    @Test(.bug("https://github.com/vapor/postgres-nio/issues/303"))
+    func selectPlainPrepared() async throws {
+        try await withTestConnection(on: MultiThreadedEventLoopGroup.singleton.any()) { connection in
+            let prepared = try await connection.prepare(query: "SELECT 10, 20").get()
+            let rows = try await prepared.execute().get()
+
+            #expect(rows.count == 1)
+            let values = try #require(rows.first).decode((Int, Int).self)
+            #expect(values == (10, 20))
+
+            try await prepared.deallocate().get()
+        }
+    }
+
+    @Test(.bug("https://github.com/vapor/postgres-nio/issues/303"))
+    func selectBound() async throws {
+        try await withTestConnection(on: MultiThreadedEventLoopGroup.singleton.any()) { connection in
+            let result = try await connection.query("SELECT \(10)::int8, \(20)::int8", logger: .psqlTest).collect()
+
+            #expect(result.count == 1)
+            let values = try #require(result.first).decode((Int, Int).self)
+            #expect(values == (10, 20))
+        }
+    }
+
+    @Test(.bug("https://github.com/vapor/postgres-nio/issues/303"))
+    func selectBoundPrepared() async throws {
+        try await withTestConnection(on: MultiThreadedEventLoopGroup.singleton.any()) { connection in
+            let prepared = try await connection.prepare(query: "SELECT $1::int8, $2::int8").get()
+            let rows = try await prepared.execute([PostgresData(int: 10), PostgresData(int: 20)]).get()
+
+            #expect(rows.count == 1)
+            let values = try #require(rows.first).decode((Int, Int).self)
+            #expect(values == (10, 20))
+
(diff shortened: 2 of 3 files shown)


Tests/IntegrationTests/PreparedQueryIntegrationTests.swift
```diff
@@ -0,0 +1,58 @@
+import Logging
+import NIOCore
+import NIOPosix
+import PostgresNIO
+import Testing
+
+@Suite
+struct PreparedQueryIntegrationTests {
```
fabianfett, line 8: which of these tests didn't succeed before?

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
```diff
@@ -267,7 +267,7 @@ struct ExtendedQueryStateMachine {
             return .wait
 
         case .prepareStatement(_, _, _, let eventLoopPromise):
-            return .succeedPreparedStatementCreation(eventLoopPromise, with: rowDescription)
+            return .succeedPreparedStatementCreation(eventLoopPromise, with: RowDescription(columns: columns))
```
fabianfett, line 270: Generally I agree with the fix. I'm just wondering why this worked before.

fabianfett (changes requested):
Love it, lets add a unit test as well please though.

Sources/PostgresNIO/New/Connection State Machine/ExtendedQueryStateMachine.swift
ptoffy: it did sometimes when the text format format matched the binary one. when it didn't it would throw. this is specific to the legacy `PreparedQuery` API because with the new API we're setting to binary later on in https://github.com/vapor/postgres-nio/blob/main/Sources/PostgresNIO/New/Connection%20State%20Machine/PreparedStatementStateMachine.swift#L51

Tests/IntegrationTests/PreparedQueryIntegrationTests.swift
ptoffy: `selectPlainPrepared` and `selectBoundPrepared`. we can probably remove the other ones. I added them since they were in the issue

fabianfett (approved):
LGTM, thanks!

PR #674, opened 2026-09-14
Convert `PostgresClientTests` to Swift Testing

ptoffy:


Tests/IntegrationTests/PostgresClientTests.swift
```diff
@@ -36,130 +36,127 @@ final class PostgresClientTests: XCTestCase {
...
-                        XCTAssertEqual((error.closureError as? PSQLError)?.serverInfo?[.severity], "ERROR")
+                } catch {
+                    guard let error = error as? PostgresTransactionError else { 
+                        Issue.record("Unexpected error type: \(error)") 
+                        return
                     }
+
+                    #expect((error.closureError as? PSQLError)?.code == .server)
+                    #expect((error.closureError as? PSQLError)?.serverInfo?[.severity] == "ERROR")
                 }
```
gwynne: This can be much more nicely written using `let error = try #require(throws: PostgresTransactionError.self) { ... }` etc.

Tests/IntegrationTests/PostgresClientTests.swift
```diff
@@ -204,107 +200,102 @@ final class PostgresClientTests: XCTestCase {
                     do {
                         try await client.query("SELECT 1", logger: logger)
                     } catch {
-                        XCTFail("Unexpected error: \(error)")
+                        Issue.record("Unexpected error: \(error)")
```
gwynne: GitHub won't let me attach a suggestion to the entire correct span of code, but replace the entire `do{}catch{}` with:
```swift
await #expect(throws: Never.self) { try await client.query("SELECT 1", logger: logger) }
```

gwynne (changes requested):

gwynne (approved):

PR #675, opened 2026-09-14
Add structured query support

ptoffy:
Also adds some doc comments to some state machine states for future ease of use

Some tests were generated with Claude


Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -708,6 +708,49 @@ extension PostgresConnection {
...
+    ///   - file: The file the query was started in. Used for better error reporting.
+    ///   - line: The line the query was started in. Used for better error reporting.
+    ///   - body: The closure that's used to consume the query result.
+    /// - Returns: The result of the `body` closure.
+    public func query<Result>(
+        _ query: PostgresQuery,
+        logger: Logger,
+        file: String = #fileID,
+        line: Int = #line,
+        isolation: isolated (any Actor)? = #isolation,
```
fabianfett: lets remove those. `isolation: isolated (any Actor)? = #isolation` has more or less been superseded by `nonisolated(nonsending)`

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -708,6 +708,49 @@ extension PostgresConnection {
...
+        _ body: (PostgresRowSequence) async throws -> sending Result
+    ) async throws -> sending Result {
```
fabianfett: if we are `nonisolated(nonsending)` `sending` shouldn't be needed here.

fabianfett: we want to overloads:

1. that just returns the result
2. that returns a tuple of result and `PostgresQueryMetadata`, I think it is called.

Sources/PostgresNIO/New/PSQLError.swift
```diff
@@ -412,6 +418,8 @@ public struct PSQLError: Error, @unchecked Sendable {
 
     static let receivedUnencryptedDataAfterSSLRequest = PSQLError(code: .receivedUnencryptedDataAfterSSLRequest)
 
+    static let rowSequenceUsedOutsideScope = PSQLError(code: .rowSequenceUsedOutsideScope)
```
fabianfett, line 457: thing for a follow up: all of those should be `static vars`.

Tests/PostgresNIOTests/New/PostgresStructuredQueryTests.swift
```diff
@@ -0,0 +1,399 @@
...
+                        return count
+                    }
+                    #expect(count == 0)
+                }
+
+                let request = try await channel.waitForUnpreparedRequest()
+                #expect(request.parse.query == "DELETE FROM users")
+
+                try await channel.sendUnpreparedRequestWithNoParametersBindResponse()
+                try await channel.testingEventLoop.executeInContext { channel.read() }
```
fabianfett: are the channel reads needed here?

Sources/PostgresNIO/Connection/PostgresConnection.swift
```diff
@@ -708,6 +708,107 @@ extension PostgresConnection {
...
+    /// `PSQLError.rowSequenceNotFullyConsumed`.
+    ///
+    /// - Parameters:
+    ///   - query: The ``PostgresQuery`` to run
+    ///   - logger: The `Logger` to log into for the query
+    ///   - file: The file the query was started in. Used for better error reporting.
+    ///   - line: The line the query was started in. Used for better error reporting.
+    ///   - body: The closure that's used to consume the query result. Must consume the sequence to its end.
+    /// - Returns: The result of the `body` closure and the query metadata.
+    public func queryWithMetadata<Result>(
```
fabianfett, line 770: I think, I'd go for just query and mark this as @disfavoredOverload if needed. As soon as someone uses the tuple syntax, they should be able to use this API.

ptoffy: I do think writing `queryWithMetatadata` is more intuitive than `let (result, metadata) = query`. Although it's not a hill I'm willing to die on

PR #676, opened 2026-09-14
Convert `PSQLIntegrationTests` to Swift Testing

ptoffy:


Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -301,226 +214,186 @@ final class IntegrationTests: XCTestCase {
...
-        try await conn.copyFromBinary(table: "copy_table", columns: ["id", "name"], logger: .psqlTest) { writer in
-            let records: [(id: Int, name: String)] = [
-                (1, "Alice"),
-                (42, "Bob")
-            ]
-            for record in records {
-                try await writer.writeRow { columnWriter in
-                    try columnWriter.writeColumn(Int32(record.id))
-                    try columnWriter.writeColumn(record.name)
+    #if compiler(>=6.2)  // copyFromBinary is only available in Swift 6.2+
```
gwynne, line 350: IIRC Package.swift specifies 6.2 as the current minimum requirement, so this should not be needed. Double-check that I'm remembering correctly first, though. 🙂

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -1,28 +1,26 @@
 import Atomics
-import XCTest
+import Foundation
```
gwynne: Can we use `FoundationEssentials`?

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -32,264 +30,179 @@ final class IntegrationTests: XCTestCase {
...
+            let metadata = try await connection.query("SELECT generate_series(1, 10000);", logger: .psqlTest) { row in
                 let expected = received.wrappingIncrementThenLoad(ordering: .relaxed)
-                XCTAssertEqual(expected, try row.decode(Int64.self, context: .default))
-            }
-
-            workaround()
-        }.wait())
+                let value = try row.decode(Int64.self, context: .default)
+                #expect(expected == value)
+            }.get()
```
gwynne: There is no async version of this API? (This question should be asked of numerous other places where `.get()` is being used in these tests, not just this one instance.)

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -32,264 +30,179 @@ final class IntegrationTests: XCTestCase {
...
-        XCTAssertNoThrow(conn = try PostgresConnection.test(on: eventLoop).wait())
-        defer { XCTAssertNoThrow(try conn?.close().wait()) }
-
-        var metadata: PostgresQueryMetadata?
-        let received = ManagedAtomic<Int64>(0)
-        XCTAssertNoThrow(metadata = try conn?.query("SELECT generate_series(1, 10000);", logger: .psqlTest) { row in
-            func workaround() {
+    @Test func query10kItems() async throws {
+        try await withConnection { connection in
+            let received = ManagedAtomic<Int64>(0)
```
gwynne: Does this actually need to be atomic? Can the row callback be called concurrently or reentrantly?

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -1,28 +1,26 @@
 import Atomics
```
gwynne: Don't forget to remove this import if it turns out that the atomic in the query 10k items test isn't needed.

gwynne (changes requested):

Tests/IntegrationTests/PSQLIntegrationTests.swift
ptoffy: Nope currently it's 6.1

Tests/IntegrationTests/PSQLIntegrationTests.swift
```diff
@@ -32,264 +30,179 @@ final class IntegrationTests: XCTestCase {
...
-        defer { XCTAssertNoThrow(try conn?.close().wait()) }
-
-        var result: PostgresQueryResult?
-        XCTAssertNoThrow(result = try conn?.query("SELECT version()", logger: .psqlTest).wait())
-        let rows = result?.rows
-        var version: String?
-        XCTAssertNoThrow(version = try rows?.first?.decode(String.self, context: .default))
-        XCTAssertEqual(version?.contains("PostgreSQL"), true)
+    @Test func queryVersion() async throws {
+        try await withConnection { connection in
```
fabianfett, line 48: please use an explicit `self` everywhere.

ptoffy: I was about to move `withConnection` to the Utilities file outside of this scope since other files will probably need it as well, which would mean self would be wrong. Does that sound good?

ptoffy: Will add it in all other cases

Tests/IntegrationTests/PSQLIntegrationTests.swift
gwynne: Then bump it. 6.4 is out, so 6.2 is now the ecosystem baseline. 😁

gwynne (approved):

Tests/IntegrationTests/PSQLIntegrationTests.swift
ptoffy: https://github.com/vapor/postgres-nio/pull/677

